PR #6, opened 2022-07-27
Bitcoin interface poller

darosior:
This introduces the Bitcoin poller, an event loop in a new thread that continuously poll our Bitcoin backend to update our state. This is kept minimal; only introduce the necessary components (Bitcoin interface, DB interface) and as such it only takes care of updating the best tip at the moment.


darosior:
self-ACK dd37255d7b03051aa8924127112ff8fac6c38ff7 -- again, it was inspired and adapted from revaultd, where a similar logic was reviewed and well tested. This PR was tested with the following one, #9.

PR #7, opened 2022-07-27
Descriptors creation

darosior:
Eventually this module will contain the routines to create all the descriptors we support. For now it only contains the one for the inheritance descriptor.


edouardparis (approved):
ACK 9b98b0852d1d34496493f3b26219593f549f0ac4

PR #9, opened 2022-07-30
Introduce the `commands` module

darosior:
This module contains the implementation of our external API. Like the others, it's very inspired (or grossly copied from) `revaultd`.

This PR implements the `getinfo` and `getnewaddress` commands to showcase the usage of the new `DaemonControl` command.

PR #1, opened 2022-07-21
Add doc directory with API.md

edouardparis:

PR #11, opened 2022-08-05
JSONRPC2 server

darosior:
On top of #9, this introduces a simple JSONRPC server calling into the exposed commands. I started by reusing code from revaultd, but ended up rewriting it entirely. Compared to it, it t is simpler, smaller and dependency-less.

In order to sanity check the whole server startup and teardown in unit tests, this took a detour by introducing a `testutils` module with mocked Bitcoin and database interfaces. Those will be useful later.

This introduces a JSONRPC-specifc command: `stop`, which shuts down the daemon. We add a `doc/API.md` adapted from #1 with documentation for the existing commands.


src/jsonrpc/server.rs
```diff
@@ -0,0 +1,429 @@
...
+    let connections_counter = sync::Arc::from(atomic::AtomicU32::new(0));
+    let shutdown = sync::Arc::from(atomic::AtomicBool::new(false));
+
+    listener.set_nonblocking(true)?;
+    while !shutdown.load(atomic::Ordering::Relaxed) {
+        let (connection, _) = match listener.accept() {
+            Ok(c) => c,
+            Err(e) => {
+                log::error!("Accepting new connection: '{}'", e);
+                continue;
```
darosior, line 137: Since the socket is set to nonblocking: don't log as error here, and sleep a few milliseconds.

src/commands/mod.rs
```diff
@@ -26,33 +27,76 @@ impl DaemonControl {
...
+#[derive(Debug, Clone, Serialize)]
 pub struct GetInfoResult {
     pub version: String,
     pub network: bitcoin::Network,
     pub blockheight: i32,
     pub sync: f64,
     pub descriptors: GetInfoDescriptors,
 }
+
+#[derive(Debug, Clone, Serialize)]
```
edouardparis, line 63: Please add Deserializer also, It is useful for the gui to reuse the struct

edouardparis: Same for GetInfoResult

darosior: Hmm i'll do it here for what is touched, but for the other results i'll make a new PR.

darosior: In the end i just added a commit here.

darosior:
self-ACK 798bf67e21977ad18ee680a14e23e534f64a3ea5 -- tested using following PRs #13 and #17.

PR #13, opened 2022-08-16
Coin tracking

darosior:

PR #17, opened 2022-08-18
Typesafe descriptor management

darosior:
This introduces types for our inheritance descriptor. This is taken and adapted from revaultd, where it turned out to be very helpful.

This also adds a helper to query the timelock of the heir's spending path in the inheritance descriptor for downstream users to be able to compute the expiration block of each coin.

This is based on #13.


edouardparis (approved):
utACK ee86a1bd5ebbda39885f1e8869bf8ae03552d682

PR #8, opened 2022-07-27
gui: loader and installer

edouardparis:
Wait before merging

PR #19, opened 2022-08-22
Gui settings

edouardparis:
based on #8


edouardparis (commented):
ACK 29e134748df363963092e16dcd288ed6431d7434
Code from revault-gui with `context` module renamed for `cache`

PR #29, opened 2022-09-16
Check spend status in bitcoin poller

edouardparis:
Add two new columns:

- `blocktime`: timestamp of the block containing the transaction funding the coin.
- `spent_at`: timestamp of the block containing the transaction spending the coin.

Update the coin `spent_at` when the spend transaction is confirmed

### DIFF
--- a/src/bitcoin/mod.rs
+++ b/src/bitcoin/mod.rs
@@ -6,6 +6,7 @@ pub mod poller;
 
 use d::LSBlockEntry;
 
+use std::collections::HashMap;
 use std::sync;
 
 use miniscript::bitcoin::{self, hashes::Hash};
@@ -45,6 +46,12 @@ pub trait BitcoinInterface: Send {
         &self,
         outpoints: &[bitcoin::OutPoint],
     ) -> Vec<(bitcoin::OutPoint, bitcoin::Txid)>;
+
+    /// Get all coins that are spent with the final spend tx txid and blocktime.
+    fn spent_coins(
+        &self,
+        outpoints: &[(bitcoin::OutPoint, bitcoin::Txid)],
+    ) -> Vec<(bitcoin::OutPoint, bitcoin::Txid, u32)>;
 }
 
 impl BitcoinInterface for d::BitcoinD {
@@ -141,6 +148,39 @@ impl BitcoinInterface for d::BitcoinD {
 
         spent
     }
+
+    fn spent_coins(
+        &self,
+        outpoints: &[(bitcoin::OutPoint, bitcoin::Txid)],
+    ) -> Vec<(bitcoin::OutPoint, bitcoin::Txid, u32)> {
+        let mut spent = Vec::with_capacity(outpoints.len());
+
+        let mut cache: HashMap<bitcoin::Txid, Option<d::GetTxRes>> = HashMap::new();
+
+        for (op, txid) in outpoints {
+            let tx: Option<&d::GetTxRes> = match cache.get(txid) {
+                Some(tx) => tx.as_ref(),
+                None => {
+                    let tx = self.get_transaction(txid);
+                    cache.insert(*txid, tx);
+                    cache.get(txid).unwrap().as_ref()
+                }
+            };
+
+            if let Some(tx) = tx {
+                if let Some(block_height) = tx.block_height {
+                    if block_height > 1 {
+                        spent.push((*op, *txid, tx.block_time.expect("Spend is confirmed")))
+                    }
+                } else {
+                    // TODO: handle the case where new transaction which txid is not the
+                    // coin spending_txid, spent the coin.
+                }
+            }
+        }
+
+        spent
+    }
 }
 
 // FIXME: do we need to repeat the entire trait implemenation? Isn't there a nicer way?
@@ -177,6 +217,13 @@ impl BitcoinInterface for sync::Arc<sync::Mutex<dyn BitcoinInterface + 'static>>
         outpoints: &[bitcoin::OutPoint],
     ) -> Vec<(bitcoin::OutPoint, bitcoin::Txid)> {
         self.lock().unwrap().spending_coins(outpoints)
+    }
+
+    fn spent_coins(
+        &self,
+        outpoints: &[(bitcoin::OutPoint, bitcoin::Txid)],
+    ) -> Vec<(bitcoin::OutPoint, bitcoin::Txid, u32)> {
+        self.lock().unwrap().spent_coins(outpoints)
     }
 }
 
--- a/src/bitcoin/mod.rs
+++ b/src/bitcoin/mod.rs
@@ -156,7 +156,6 @@ impl BitcoinInterface for d::BitcoinD {
         let mut spent = Vec::with_capacity(outpoints.len());
 
         let mut cache: HashMap<bitcoin::Txid, Option<d::GetTxRes>> = HashMap::new();
-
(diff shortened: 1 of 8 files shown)


src/database/mod.rs
```diff
@@ -52,16 +52,22 @@ pub trait DatabaseConnection {
     ) -> Option<bip32::ChildNumber>;
 
     /// Get all UTxOs.
-    fn unspent_coins(&mut self) -> HashMap<bitcoin::OutPoint, Coin>;
+    fn list_unspent_coins(&mut self) -> HashMap<bitcoin::OutPoint, Coin>;
+
+    /// Get all UTxOs.
```
darosior, line 57: Bad copy pasta

src/database/mod.rs
```diff
@@ -52,16 +52,22 @@ pub trait DatabaseConnection {
...
+    fn list_spending_coins(&mut self) -> HashMap<bitcoin::OutPoint, Coin>;
 
     /// Store new UTxOs. Coins must not already be in database.
     fn new_unspent_coins<'a>(&mut self, coins: &[Coin]);
 
     /// Mark a set of coins as being confirmed at a specified height and block time.
     fn confirm_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, i32, u32)]);
 
-    /// Mark a set of coins as being spent by a specified txid.
+    /// Mark a set of coins as being spent by a specified txid of a pending transaction.
```
darosior, line 66: nit: Looks like it belongs in the previous commit

src/database/mod.rs
```diff
@@ -52,16 +52,22 @@ pub trait DatabaseConnection {
...
     fn spending_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, bitcoin::Txid)]);
+
+    /// Mark a set of coins as spent by a specified txid at a specified block time.
+    fn spent_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, bitcoin::Txid, u32)]);
```
darosior, line 70: nit: again the method name sounds like you are querying the spent coins. What about having (if you rename the other method i suggested in the previous commit):
- `spend_coins` -> Mark coins as being spent by a specifc txid.
- `confirm_spend` -> Mark the spending transaction of those coins as being confirmed at a given time.

src/database/sqlite/mod.rs
```diff
@@ -260,6 +260,17 @@ impl SqliteConn {
         .expect("Db must not fail")
     }
 
+    /// List coins that are being spent and that are waiting confirmation.
```
darosior, line 263: nit: "List coins that are being spent and whose spending transaction is still unconfirmed"

src/testutils.rs
```diff
@@ -137,6 +148,17 @@ impl DatabaseConnection for DummyDbConn {
         }
     }
 
+    fn spent_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, bitcoin::Txid, u32)]) {
+        for (op, spend_txid, time) in outpoints {
+            let mut db = self.db.write().unwrap();
+            let spent = &mut db.coins.get_mut(op).unwrap();
+            assert!(spent.spend_txid.is_none());
+            assert!(spent.spent_at.is_none());
```
darosior, line 156: `spend_txid` shouldn't be `None` at the time you are confirming the spend transaction, it looks like a copy paste of the above method.

src/database/sqlite/mod.rs
```diff
@@ -320,6 +331,29 @@ impl SqliteConn {
...
 
+    /// Mark a set of coins as spent.
+    pub fn spent_coins<'a>(
+        &mut self,
+        outpoints: impl IntoIterator<Item = &'a (bitcoin::OutPoint, bitcoin::Txid, u32)>,
+    ) {
+        db_exec(&mut self.conn, |db_tx| {
+            for (outpoint, spend_txid, time) in outpoints {
+                db_tx.execute(
+                    "UPDATE coins SET spend_txid = ?1, spent_at = ?2 WHERE txid = ?3 AND vout = ?4",
```
darosior, line 342: These two new methods need (even trivial) unit testing. :)

src/bitcoin/mod.rs
```diff
@@ -172,9 +171,28 @@ impl BitcoinInterface for d::BitcoinD {
                     if block_height > 6 {
                         spent.push((*op, *txid, tx.block_time.expect("Spend is confirmed")))
                     }
-                } else {
-                    // TODO: handle the case where new transaction which txid is not the
-                    // coin spending_txid, spent the coin.
+                } else if !tx.conflicting_txs.is_empty() {
+                    for txid in tx.conflicting_txs.clone() {
```
darosior, line 175: Do you need to `clone()`? Can't you just `as_ref()`?

src/bitcoin/mod.rs
```diff
@@ -172,9 +171,28 @@ impl BitcoinInterface for d::BitcoinD {
...
+                            Some(tx) => tx.as_ref(),
+                            None => {
+                                let tx = self.get_transaction(&txid);
+                                cache.insert(txid, tx);
+                                cache.get(&txid).unwrap().as_ref()
+                            }
+                        };
+                        if let Some(tx) = tx {
+                            if let Some(block_height) = tx.block_height {
+                                if block_height > 6 {
```
darosior, line 186: Same comment as in previous commit. I think we should remove the check for more than 6 confirmations.

darosior (commented):
Cool. I've found one "bug" (an edge case that isn't covered) in the poller, and the rest is mostly nits. Let me know if you prefer to ignore them. Though i do feel somewhat strongly about (in order of preference):
1. Removing the arbitrary 6-confs rule. I added it in Revault since we had presigned transactions but here i think assuming it's confirmed with a single conf is fine for now.
2. The interface for marking a spend as confirmed / the name of some of the trait methods introduced

src/database/mod.rs
```diff
@@ -123,12 +127,12 @@ impl DatabaseConnection for SqliteConn {
...
     }
 
-    fn confirm_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, i32)]) {
+    fn confirm_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, i32, u32)]) {
         self.confirm_coins(outpoints)
     }
 
-    fn spend_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, bitcoin::Txid)]) {
-        self.spend_coins(outpoints)
+    fn spending_coins<'a>(&mut self, outpoints: &[(bitcoin::OutPoint, bitcoin::Txid)]) {
```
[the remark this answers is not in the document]
edouardparis: ok then my spend_coins become spent_coins and I let this spend_coin as it is, ok ?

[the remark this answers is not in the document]
darosior: More context in a comment from the next commit https://github.com/revault/minisafe/pull/29#discussion_r984986968

src/bitcoin/mod.rs
edouardparis: There is an immutable borrow on the `cache.get`, but you are right that it is a certainly better  way to do it

darosior (commented):
Almost ACK. I'm just confused by the check that the `block_height` of the spending transaction be higher than `1`. Feel free to ignore the rest.

darosior:
ACK 172cda19a0a72b77e3832f2e29ed7ff6f8062c44

PR #60, opened 2022-10-12
Clippification

darosior:
This fixes *all* clippy warning from the whole repository. Some of them are... Well very opinionated to say the least. But to avoid bikeshedding i just applied all of them. Let's start fresh!

@edouardparis can you skim through the clippy fixes in the GUI folder? I checked the daemon one already.

Fixes #10.

PR #61, opened 2022-10-13
gui: fix clippy errors

edouardparis:


darosior:
ACK on the correctness. But this was already done as part of #60, which actually contains more clippy fixes for the GUI. So i think it is preferable to go with #60 instead.

PR #74, opened 2022-10-24
Decent change management (using multipath descriptors)

### LINKED ISSUE #18: Don't mixup receiving and change addresses

darosior:
This fixes #18 by implementing the de-facto standard of using a `/0/*` keychain for receiving addresses and a `/1/*` keychain for change addresses. Note that once we'll have multisig, reusing addresses will still be possible since wallet don't share the same "next derivation index".

In order to avoid forcing the user to configure and backup two almost identical descriptors, we make use of the recently proposed BIP389 (https://github.com/bitcoin/bips/pull/1354). In order to prevent as much as possible introducing a backward incompatibility in the configuration file after the first release, we restrict the usage of multipath descriptors to `<0;1>` here.
We now derive public keys from `xpub/0/*` and `xpub/1/*` while we were previously deriving them from `xpub/*`.

This triggered a pretty invasive refactoring, as most parts of the codebase had to be updated to support the new change/receive separation (even the functional test miniscript dependency had to be updated, see https://github.com/darosior/python-bip380/pull/21).
Broadly, this:
1. Update our Miniscript dependency to my upstream PR (https://github.com/rust-bitcoin/rust-miniscript/pull/470) rebased on top of the 8.0.0 release.
2. Updates the descriptors module to handle somewhat safely the multipath descriptors (to avoid mixing up the single, multi, and derived descriptors).
3. Makes a multipath descriptor mandatory in the configuration file.
4. Updates the Bitcoin backend poller aware of descriptors for which to track coins.
   - Necessarily this updates the bitcoind implementation to import two descriptors
5. Record in database whether a coin was for the change or receive descriptor, in addition to its derivation index

### DIFF
--- a/src/database/sqlite/schema.rs
+++ b/src/database/sqlite/schema.rs
@@ -44,6 +44,7 @@ CREATE TABLE coins (
     vout INTEGER NOT NULL,
     amount_sat INTEGER NOT NULL,
     derivation_index INTEGER NOT NULL,
+    is_change BOOLEAN NOT NULL CHECK (is_change IN (0,1)),
     spend_txid BLOB,
     spend_block_height INTEGER,
     spend_block_time INTEGER,
@@ -146,6 +147,7 @@ pub struct DbCoin {
     pub block_time: Option<u32>,
     pub amount: bitcoin::Amount,
     pub derivation_index: bip32::ChildNumber,
+    pub is_change: bool,
     pub spend_txid: Option<bitcoin::Txid>,
     pub spend_block: Option<DbSpendBlock>,
 }
@@ -168,12 +170,13 @@ impl TryFrom<&rusqlite::Row<'_>> for DbCoin {
         let amount = bitcoin::Amount::from_sat(amount);
         let der_idx: u32 = row.get(7)?;
         let derivation_index = bip32::ChildNumber::from(der_idx);
+        let is_change: bool = row.get(8)?;
 
-        let spend_txid: Option<Vec<u8>> = row.get(8)?;
+        let spend_txid: Option<Vec<u8>> = row.get(9)?;
         let spend_txid =
             spend_txid.map(|txid| encode::deserialize(&txid).expect("We only store valid txids"));
-        let spend_height: Option<i32> = row.get(9)?;
-        let spend_time: Option<u32> = row.get(10)?;
+        let spend_height: Option<i32> = row.get(10)?;
+        let spend_time: Option<u32> = row.get(11)?;
         assert_eq!(spend_height.is_none(), spend_time.is_none());
         let spend_block = spend_height.map(|height| DbSpendBlock {
             height,
@@ -188,6 +191,7 @@ impl TryFrom<&rusqlite::Row<'_>> for DbCoin {
             block_time,
             amount,
             derivation_index,
+            is_change,
             spend_txid,
             spend_block,
         })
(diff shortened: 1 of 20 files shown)


src/database/sqlite/schema.rs
```diff
@@ -57,7 +59,8 @@ CREATE TABLE coins (
  * we can get the derivation index from the parent descriptor from bitcoind.
  */
 CREATE TABLE addresses (
-    address TEXT NOT NULL UNIQUE,
+    receive_address TEXT NOT NULL UNIQUE,
+    change_address TEXT NOT NULL UNIQUE,
```
edouardparis, line 63: why having row in database with the tuple (receive and change) and not having a `is_change` bool ?

src/descriptors.rs
```diff
@@ -162,42 +171,62 @@ fn csv_check(csv_value: u32) -> Result<(), DescCreationError> {
...
+/// and a timelocked branch (the heir). All keys in this descriptor are singlepath.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct InheritanceDescriptor(descriptor::Descriptor<descriptor::DescriptorPublicKey>);
 
 /// Derived (containing only raw Bitcoin public keys) version of the inheritance descriptor.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct DerivedInheritanceDescriptor(descriptor::Descriptor<DerivedPublicKey>);
 
-impl fmt::Display for InheritanceDescriptor {
+impl fmt::Display for MultipathDescriptor {
```
edouardparis, line 212: Can I use this method in the installer to fill the config descriptor field ?

edouardparis: I guess yes: https://github.com/revault/minisafe/pull/74/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR93

src/database/sqlite/schema.rs
darosior: Maybe i misunderstand your question, but you need an address->derivation index mapping. You necessarily need a key per address in this mapping.

edouardparis: I meant why storing:
| receive_address | change_address | derivation_index |
|------------------|------------------------|--------------------------|
| addr1 | addr2 | 1|

and not
 | address | is_change | derivation_index |
|------------------|------------------------|--------------------------|
| addr1 | false | 1|
| addr2 | true | 1|

darosior: Because it's wasteful? That's 2 added integer per derivation index. What advantage would it have?

edouardparis: To only store address that we use, but maybe that is not our process:
If we do no spend at all and receive only deposit until derivation_index = 10, does it means that we want to store the 10 change addresses that are not used at all ?

edouardparis:
Is it possible to add in this PR, helpers that are retrieving the timelock, the user and the heir multiXkey from the multipathdescriptor ?

src/database/sqlite/schema.rs
darosior: This mapping is for detecting coins up to the gap limit, not to store addresses we used. Necessarily we need to have addresses up to the gap limit.

>  If we do no spend at all and receive only deposit until derivation_index = 10, does it means that we want to store the 10 change addresses that are not used at all ?

Are we sure we are *never* ever going to receive a payment on an address before derivation index 10? I don't think so at all. What if we gave away this address already? We trust others to not use it? What if we didn't but another wallet did? What if we recovered from backup? Etc..

darosior:
It'd be pretty orthogonal to this PR. But i can implement that for sure in another PR! Note you can already query the timelock though.

edouardparis (approved):
ACK 117171f24ff5bd6731d9e5e50e4515a03013a9eb

PR #71, opened 2022-10-21
gui: import xpubs from hws

edouardparis:

PR #75, opened 2022-10-27
Add spend panel

edouardparis:
based on #71


darosior:
Does this fix https://github.com/revault/minisafe/issues/34?

edouardparis:
No am comparing psbt output script_pubkey with entered user addresses to find the one added by minisafed

darosior:
Ok, nice. Does it mean we can close the issue then? Do you think we'll still need it eventually?

Keep in mind you can't rely on the ordering though. Unfortunately you *have* to walk through all the outputs of the transaction if you do that. It can still be `O(n)` if you make the addresses fed by the users in a HashMap though, so it should be acceptable for now.

edouardparis (commented):
Self-ACK 6f75515f9b572d6fa77facdba9ddd2cd1bf94a64
ci errors are related to the daemon.
I tested the PR the entire week

PR #79, opened 2022-11-04
commands: actually use the change descriptor when creating a Spend

### LINKED ISSUE #78: listspendtxs does not return change_index
https://github.com/revault/minisafe/blob/master/src/commands/utils.rs#L37
it is deprecated
listspendtxs does not return a change_index (always null)

darosior:
Fixes #78

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -348,7 +348,7 @@ impl DaemonControl {
             let change_desc = self
                 .config
                 .main_descriptor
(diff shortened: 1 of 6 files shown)


darosior:
Thanks to @edouardparis for pointing out i had forgotten to update the change index detection in `listspendtxs`, which uncovered this bug...

darosior:
ACK 08505b25452e61a92d5b7eb4a02c893f0d585a91

PR #84, opened 2022-11-09
Block chain rescanning for our descriptor(s)

### LINKED ISSUE #28: Rescan from a given timestamp

darosior:
This implements the ability for a user to rescan the block chain from an earlier timestamp to find transactions the wallet isn't currently aware of.

The main difficulty arises with the fact there are two rescans to be performed:
1. The block chain rescan on the Bitcoin backend. For the bitcoind implementation this means we need to rescan the bitcoind watchonly wallet down to the given timestamp (or height). We can imagine for an indexer backend such as an Electrum backend it could be almost instant.
2. The update of our state with the new state of the Bitcoin backend. In the poller we scan blocks linearly, we will never scan a block that we already scanned and that includes blocks anterior to the first block we scanned. In order to implement this rescan of past blocks after a Bitcoin backend rescan was performed we reused our current chain reorganization logic: we rewind our state down to the first block that was rescanned by the backend. This is overkill (we don't need to erase existing information), but sufficient for now.

The implementation is not as robust as it could be. Notably, in order to rescan the chain on the bitcoind implementation of the backend i use a trick to not make the connection block for the entire duration of the rescan when calling `importdescriptors`: i introduce a new client that has a timeout of 1 second and that ignores the response as well as timeout errors. Obviously this is problematic since it makes it more complicated to know whether the call actually succeeded.
~~A more robust solution would be to use `rescanblockchain` which starts the rescan without blocking but~~
1. We should be storing a height in the database, not a timestamp
2. The timestamp of the descriptors in the bitcoind watchonly wallet wouldn't be updated

EDIT: i'm dumb, `rescanblockchain` does block too!

I'm happy to have feedback on this. I think the current implementation is fine for now but eventually we should move to using `rescanblockchain` and get rid of the "noreply" client hack.

Also, some  functional tests for edge cases we could encounter are included but happy to have some more suggestions.

Fixes #28.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -43,13 +43,14 @@ This command does not take any parameter for now.
 
 #### Response
 
-| Field                | Type    | Description                                                                                  |
-| -------------------- | ------- | -------------------------------------------------------------------------------------------- |
-| `version`            | string  | Version following the [SimVer](http://www.simver.org/) format                                |
-| `network`            | string  | Answer can be `mainnet`, `testnet`, `regtest`                                                |
-| `blockheight`        | integer | The block height we are synced at.                                                           |
-| `sync`               | float   | The synchronization progress as percentage (`0 < sync < 1`)                                  |
-| `descriptors`        | object  | Object with the name of the descriptor as key and the descriptor string as value             |
+| Field                | Type    | Description                                                                                        |
+| -------------------- | ------- | -------------------------------------------------------------------------------------------------- |
+| `version`            | string        | Version following the [SimVer](http://www.simver.org/) format                                |
+| `network`            | string        | Answer can be `mainnet`, `testnet`, `regtest`                                                |
+| `blockheight`        | integer       | The block height we are synced at.                                                           |
+| `sync`               | float         | The synchronization progress as percentage (`0 < sync < 1`)                                  |
+| `descriptors`        | object        | Object with the name of the descriptor as key and the descriptor string as value             |
+| `rescan_progress`    | float or null | Progress of an ongoing rescan as a percentage (between 0 and 1) if there is any              |
 
 ### `getnewaddress`
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -13,6 +13,7 @@ Commands must be sent as valid JSONRPC 2.0 requests, ending with a `\n`.
 | [`listspendtxs`](#listspendtxs)                             | List all stored Spend transactions                   |
 | [`delspendtx`](#delspendtx)                                 | Delete a stored Spend transaction                    |
 | [`broadcastspend`](#broadcastspend)                         | Finalize a stored Spend PSBT, and broadcast it       |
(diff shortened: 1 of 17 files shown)


doc/API.md
```diff
@@ -42,13 +43,14 @@ This command does not take any parameter for now.
...
+| Field                | Type    | Description                                                                                        |
+| -------------------- | ------- | -------------------------------------------------------------------------------------------------- |
+| `version`            | string        | Version following the [SimVer](http://www.simver.org/) format                                |
+| `network`            | string        | Answer can be `mainnet`, `testnet`, `regtest`                                                |
+| `blockheight`        | integer       | The block height we are synced at.                                                           |
+| `sync`               | float         | The synchronization progress as percentage (`0 < sync < 1`)                                  |
+| `descriptors`        | object        | Object with the name of the descriptor as key and the descriptor string as value             |
+| `rescan_progress`    | float or null | Progress of an ongoing rescan as a percentage (between 0 and 1) if there is any              |
```
edouardparis, line 53: It feels like it is a duplicate of `sync`, but maybe you are right to add it explicitly

src/bitcoin/d/mod.rs
```diff
@@ -116,7 +132,11 @@ impl From<simple_http::Error> for BitcoindError {
 }
 
 pub struct BitcoinD {
+    /// Client for generalistic calls.
     node_client: Client,
+    /// A client that will disregard responses to the queries it makes.
+    sendonly_client: Client,
```
edouardparis, line 138: Ok for the hack, but I feel that we can change it later for a new param for the `make_request` method, instead of allocating a new whole client.

darosior: Yeah though i checked and allocating a new whole client is just 64 bytes plus this struct:
```rust
pub struct SimpleHttpTransport {
    addr: net::SocketAddr,
    path: String,
    timeout: Duration,
    /// The value of the `Authorization` HTTP header.
    basic_auth: Option<String>,
}
```
So that's definitely reasonable enough in terms of memory footprint.

darosior:
i'm dumb, rescanblockchain does block too!

doc/API.md
darosior: I think it makes sense to have both in the raw RPC interface. It could make sense to display it similarly to the sync progress in the GUI though?

darosior:
Alright, i've pushed a fix for some errors we've encountered while testing this:
- The retry logic in the bitcoind interface may try to start the rescan multiple times (would previously make us crash, h/t @edouardparis for the repro of this bug)
- We were missing a key parameter to `importdescriptors` when updating the descriptors: `range`

Since this logic is inherently racy, i've also removed the panics and assert to propagate a potential error to the command interface instead of crashing.

edouardparis:
ACK 55cf21fa2aa7a66d998061684630a32cfb972a35
Tested with gui

darosior:
ACK 32f3bdeb16b66878e23ba71a6f2c401c5d1f601a

PR #85, opened 2022-11-10
Delete install target datadir in case of failure

### LINKED ISSUE #20: At fresh daemon startup, setup first bitcoind before database
During a fresh install the bitcoind should be setup before installing the sqlite database,
because if done after and the bitcoind port in the config is wrong for example, the database and the datadir are already created which force user to remove it by hand before installing again with the good port number

edouardparis:
close #20


edouardparis (commented):
Self-ACK e9fa816be7c6ddbf64413dc4df6940d0f90c1f87

PR #89, opened 2022-11-11
installer: check xpub network

edouardparis:


gui/src/installer/step/descriptor.rs
```diff
@@ -176,6 +200,14 @@ impl Step for DefineDescriptor {
     }
 }
 
+fn check_key_network(key: &DescriptorPublicKey, network: Network) -> bool {
+    match key {
+        DescriptorPublicKey::XPub(key) => key.xkey.network == network,
+        DescriptorPublicKey::MultiXPub(key) => key.xkey.network == network,
+        _ => true,
+    }
```
darosior, line 208: It's not that simple, unfortunately. For regtest or signet the expected xpub network is testnet. See how it's done in minisafed:
https://github.com/revault/minisafe/blob/5595e91e5700e17a692890b691affa7ffd039fdc/src/config.rs#L199-L216

gui/src/installer/view.rs
```diff
@@ -82,7 +82,7 @@ pub fn define_descriptor<'a>(
                     form::Form::new("Xpub", user_xpub, |msg| {
                         Message::DefineDescriptor(message::DefineDescriptor::UserXpubEdited(msg))
                     })
-                    .warning("Please enter correct xpub")
+                    .warning("Please enter correct xpub for the chosen network")
```
darosior, line 85: That's not a very helpful error message. I think we should go further and say something like:
```suggestion
                    .warning("Please enter an extended key ('xpub') encoded for the configured network (that is, a 'tpub' for test networks)")
```

Also, what about when they import it from their signing device? Can they in all cases get a `tpub`? If so we should probably tell them to do so. I'm not sure they really can on the Ledger, can they?

edouardparis: The message becomes too long, I will see if I can change it according to the network choosen previously

darosior: Since we only ever support mainnet and test network we can be more concise. Something like:
> Please enter a valid `xpub` (or `tpub` for test networks)

gui/src/installer/view.rs
```diff
@@ -82,7 +83,11 @@ pub fn define_descriptor<'a>(
                     form::Form::new("Xpub", user_xpub, |msg| {
                         Message::DefineDescriptor(message::DefineDescriptor::UserXpubEdited(msg))
                     })
-                    .warning("Please enter correct xpub")
+                    .warning(if network == bitcoin::Network::Bitcoin {
+                        "Please enter correct xpub"
+                    } else {
+                        "Please enter correct tpub"
+                    })
```
darosior, line 90: Nice :+1:, disregard my above comment it's good that way.

Issue #81, opened 2022-11-06
Update derivation index from the chain


darosior:
This not only matters in a multisig situation but also for instance after a rescan.

PR #93, opened 2022-11-16
Derivation index update from the chain

darosior:
Currently, we wouldn't update our "next derivation index" to derive receive/change addresses if we noticed an address from a higher derivation index was used onchain. See #81. This fixes it.

Note that it can be tempting to set it with some leeway (for instance if we notice index `n` was used, we set our next derivation index to `n + 10`) since other wallets probably have generated addresses for more addresses than were actually used onchain.
However, i think it could have cascading effects in a multisig situation: A would use `n`, B would set its to `n + 10` and use `n + 11`, C would set its to `n + 21` and use `n + 22`. This is clear that we could end up very far indeed down the derivation tree without having used most indexes.

A functional test is included, demonstrating the behaviour after a rescan where we lost the database (and therefore knowledge of the previous "next derivation index"). It is in a separate commit to help reviewers attest it would fail on master and pass on this branch.

### DIFF
--- a/src/bitcoin/poller/looper.rs
+++ b/src/bitcoin/poller/looper.rs
@@ -9,7 +9,7 @@ use std::{
     thread, time,
 };
 
-use miniscript::bitcoin;
+use miniscript::bitcoin::{self, secp256k1};
 
 #[derive(Debug, Clone)]
 struct UpdatedCoins {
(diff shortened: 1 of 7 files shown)


darosior:
self-ACK 94d8ea6d600479d00e08992aaf153e136e5198fa

PR #95, opened 2022-11-16
Add check height to coins panel

### LINKED ISSUE #82: gui: add height / sequence check in coins panel

edouardparis:
close #82

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -24,6 +24,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+
+[[package]]
 name = "ab_glyph"
 version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


edouardparis (commented):
Self-ACK 44d146ab2d316e2db0dbb65f095c3b8bf757399c

darosior:
This looks nice. My suggestions:
- When a coin's timelock is expired the text should rather be "Recovery path available"
- Remove the sentence from the tile, move it to a new line in the details. In the detail either write:
  - "The recovery path will be available in X blocks"
  - "The recovery path is available"
- In the tile keep the hourglass. Either have:
  - "X <hourglass>" in black or grey when the timelocked isn't expired yet
  - "0 <hourglass>" in red when the timelocked is expired

darosior:
Also, i don't think we should list spent coins by default (as discused in our call this morning). But let's keep that for another PR.

darosior:
Nice! Why the minus "-" sign before the 0 though?

edouardparis:
its a ~

edouardparis:
2022-11-17T11:40:57,748940765+01:00

darosior:
Oh, ok. I still think we shouldn't have it. Your latest screenshot looks good to me!

edouardparis (commented):
Self-ACK 34b833f1484f1979087ad76847cc2942c06d0817

PR #21, opened 2022-08-25
Gethistory

edouardparis:
A new command `gethistory` that retrieves the `<limit>` events between `<start>` and `<end>` timestamps (See doc/API.md)
I requires two new columns in the `coins` table: 
* `blocktime`: timestamp of the block containing the transaction funding the coin.
* `spent_at`: timestamp of the block containing the transaction spending the coin.

PR #99, opened 2022-11-17
command: add listtransactions

edouardparis:


src/commands/mod.rs
```diff
@@ -622,25 +670,42 @@ pub struct ListSpendResult {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ListTransactionsResult {
+    pub transactions: Vec<TransactionInfo>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TransactionInfo {
+    pub tx: bitcoin::Transaction,
```
edouardparis, line 680: bug: it is not hex encoded in the response

doc/API.md
```diff
@@ -15,6 +15,7 @@ Commands must be sent as valid JSONRPC 2.0 requests, ending with a `\n`.
 | [`delspendtx`](#delspendtx)                                 | Delete a stored Spend transaction                    |
 | [`broadcastspend`](#broadcastspend)                         | Finalize a stored Spend PSBT, and broadcast it       |
 | [`startrescan`](#startrescan)                               | Start rescanning the block chain from a given date   |
+| [`listtransactions`](#listtransactions)                     | List of transactions of incoming and outgoing funds  |
```
darosior: The commit says " Add list_confirmed_transactions to Command " but this is listtransactions?

doc/API.md
```diff
@@ -218,3 +217,29 @@ This command does not return anything for now.
...
+
+`listtransactions` retrieves a paginated and ordered list of transactions.
+
+#### Request
+
+| Field         | Type         | Description                                                           |
+| ------------- | ------------ | --------------------------------------------------------------------  |
+| `start`       | int          | Timestamp of the beginning of the period to retrieve transactions for |
+| `end`         | int          | Timestamp of the end of the period to retrieve transactions for       |
+| `limit`       | int          | Maximum number of events to retrieve                                  |
```
darosior, line 232: It looks like it should be `listconfirmed`?

doc/API.md
```diff
@@ -5,17 +5,18 @@ interface over a Unix Domain socket.
...
+| [`getinfo`](#getinfo)                                       | Get general information about the daemon                      |
+| [`getnewaddress`](#getnewaddress)                           | Get a new receiving address                                   |
+| [`listcoins`](#listcoins)                                   | List all wallet transaction outputs.                          |
+| [`listspendtxs`](#listspendtxs)                             | List all stored Spend transactions                            |
+| [`delspendtx`](#delspendtx)                                 | Delete a stored Spend transaction                             |
+| [`broadcastspend`](#broadcastspend)                         | Finalize a stored Spend PSBT, and broadcast it                |
+| [`startrescan`](#startrescan)                               | Start rescanning the block chain from a given date            |
+| [`listconfirmedtxs`](#listconfirmedtxs)                     | List of confirmed transactions of incoming and outgoing funds |
+| [`listtxs`](#listtxs)                                       | List of transactions with the given tx ids                    |
```
darosior, line 19: Ooh ok you actually fix that in the commit that introduces the second command..

darosior (commented):
Your commit history is confused again. I'll review both commits in a single diff but please next time try to have a [clean history](https://github.com/revault/liana/blob/master/CONTRIBUTING.md#workflow). #99 (Do not do it now, we are tight on schedule and we're better prioritizing the rest)

darosior (commented):
Nice, i only did a first superficial review on the Github diff. I'll do a more in depth review later. This is definitely a better API than #21 i think.

I'll add a functional test for this myself, but please add a unit test for the database query (see the other unit tests in the sqlite module for examples).

src/bitcoin/d/mod.rs
```diff
@@ -1060,11 +1071,21 @@ impl From<Json> for GetTxRes {
...
+            (Some(hash), Some(height), Some(time)) => Some(Block { hash, time, height }),
+            _ => None,
+        };
+        let hex = json
+            .get("hex")
+            .and_then(Json::as_str)
+            .expect("Must be present in bitcoind response");
+        let bytes = Vec::from_hex(hex).expect("bitcoind returned a wrong transaction format");
+        let tx: bitcoin::Transaction = bitcoin::consensus::encode::deserialize(&bytes)
+            .expect("bitcoind returned a wrong transaction format");
```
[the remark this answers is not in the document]
edouardparis: Sorry it does not exist for `Transaction`: https://docs.rs/bitcoin/latest/bitcoin/?search=from_str

src/database/sqlite/mod.rs
```diff
@@ -1089,4 +1121,157 @@ mod tests {
...
+                    is_change: false,
+                    spend_txid: Some(
+                        bitcoin::Txid::from_str(
+                            "0c62a990d20d54429e70859292e82374ba6b1b951a3ab60f26bb65fee5724ff7",
+                        )
+                        .unwrap(),
+                    ),
+                    spend_block: Some(SpendBlock {
+                        height: 101_199,
+                        time: 1_123_000,
```
darosior, line 1177: Nice, this tests for the inclusive lower bound :+1:

src/database/sqlite/mod.rs
```diff
@@ -1089,4 +1121,157 @@ mod tests {
...
+                    )
+                    .unwrap(),
+                    block_height: Some(101_099),
+                    block_time: Some(1_122_000),
+                    amount: bitcoin::Amount::from_sat(98765),
+                    derivation_index: bip32::ChildNumber::from_normal_idx(1000).unwrap(),
+                    is_change: false,
+                    spend_txid: Some(
+                        bitcoin::Txid::from_str(
+                            "0c62a990d20d54429e70859292e82374ba6b1b951a3ab60f26bb65fee5724ff7",
```
darosior, line 1171: nit: would have been nice to have another coin have this as its deposit txid. But i'll do that in the functional test anyways.

src/jsonrpc/api.rs
```diff
@@ -136,6 +136,50 @@ pub fn handle_request(control: &DaemonControl, req: Request) -> Result<Response,
                 .ok_or_else(|| Error::invalid_params("Missing 'psbt' parameter."))?;
             update_spend(control, params)?
         }
+        "listconfirmed" => match req.params {
+            Some(Params::Array(params)) => {
+                if params.len() != 3 {
+                    return Err(Error::invalid_params("command requires 3 parameters"));
```
darosior, line 142:
```suggestion
                    return Err(Error::invalid_params("The 'listconfirmed' command requires 3 parameters: 'start', 'end' and 'limit'"));
```

src/jsonrpc/api.rs
```diff
@@ -136,6 +136,50 @@ pub fn handle_request(control: &DaemonControl, req: Request) -> Result<Response,
...
+                    serde_json::Value::as_u64(&params[1]),
+                    serde_json::Value::as_u64(&params[2]),
+                ) {
+                    serde_json::json!(&control.list_confirmed_transactions(
+                        start as u32,
+                        end as u32,
+                        limit
+                    ))
+                } else {
+                    return Err(Error::invalid_params("command requires 3 parameters"));
```
darosior, line 155:
```suggestion
                    return Err(Error::invalid_params("The 3 'start', 'end', and 'limit' parameters must be unsigned integers"));
```

src/jsonrpc/api.rs
```diff
@@ -136,6 +136,50 @@ pub fn handle_request(control: &DaemonControl, req: Request) -> Result<Response,
...
+                }
+            }
+            None => return Err(Error::invalid_params("command requires 3 parameters")),
```
darosior, line 158: Same as above

src/jsonrpc/api.rs
```diff
@@ -136,6 +136,50 @@ pub fn handle_request(control: &DaemonControl, req: Request) -> Result<Response,
...
+            _ => return Err(Error::invalid_params("invalid parameters")),
```
darosior, line 159: This is incorrect. See how it's done for the other commands above. Users must be able to pass parameters either as an array or as a mapping. There a `get()` helper specifically for this purpose.

While you are at it, it's probably better to make those into their own functions like it's done for all other commands that need to parse parameters.

darosior (commented):
Cool, the unit test for the DB query is good. :+1: 

The core of the code looks good to me. I'll be writing a functional test now. Just, the JSONRPC query parsing is slightly incorrect see the comments.

src/database/sqlite/mod.rs
```diff
@@ -490,6 +490,38 @@ impl SqliteConn {
...
+                    ORDER BY blocktime DESC LIMIT (?3) \
+                ) \
+                UNION \
+                SELECT * FROM (
+                    SELECT spend_txid AS txid, spend_block_time AS date FROM coins \
+                    WHERE spend_block_time >= (?1) \
+                    AND spend_block_time <= (?2) \
+                    ORDER BY spend_block_time DESC LIMIT (?3) \
+                ) \
+                ORDER BY date DESC LIMIT (?3) \
```
darosior, line 512: Is there a specific reason for having `DESC LIMIT (?3)` for inner queries? I think this is incorrect because you need to only limit the number of *distinct* txids.

For instance you have coins with deposit txids `[A, A, B, C, C, C]`, as well as coins with spend txids `[A, A, B, C]`. Now you make a query with a limit of `2`. The first subquery will give you  `[A, A]` and the second one `[A, A]`. After the `DISTINCT` in the outer query you will only have a single result: `[A]`.

I wonder whether the correct fix would be:
```suggestion
            "SELECT DISTINCT(txid) FROM ( \
                SELECT * from ( \
                    SELECT txid, blocktime AS date FROM coins \
                    WHERE blocktime >= (?1) \
                    AND blocktime <= (?2) \
                    ORDER BY blocktime \
                ) \
                UNION \
                SELECT * FROM (
                    SELECT spend_txid AS txid, spend_block_time AS date FROM coins \
                    WHERE spend_block_time >= (?1) \
                    AND spend_block_time <= (?2) \
                    ORDER BY spend_block_time \
                ) \
                ORDER BY date DESC LIMIT (?3) \
```

Or:
```suggestion
            "SELECT DISTINCT(txid) FROM ( \
                SELECT * from ( \
                    SELECT DISTINCT(txid), blocktime AS date FROM coins \
                    WHERE blocktime >= (?1) \
                    AND blocktime <= (?2) \
                    ORDER BY blocktime DESC LIMIT (?3) \
                ) \
                UNION \
                SELECT * FROM (
                    SELECT DISTINCT(spend_txid) AS txid, spend_block_time AS date FROM coins \
                    WHERE spend_block_time >= (?1) \
                    AND spend_block_time <= (?2) \
                    ORDER BY spend_block_time DESC LIMIT (?3) \
                ) \
                ORDER BY date DESC LIMIT (?3) \
```

darosior: Actually i think the second solution would wreck the ordering. So let's go for the first one. I'll wait for your opinion though.

darosior:
Pushed a functional test exercising the two commands. I've also pushed a commit to fix https://github.com/revault/liana/pull/99#discussion_r1029619579.

darosior:
The linter still needs fixing though.

darosior:
ACK c39cb07360e5d3a6f528d4b17f1e34fae0da550b

darosior:
I stressed test this by running the functional test hundreds of times. No issue.

PR #126, opened 2022-11-23
Spent coin detection fixes

### LINKED ISSUE #125: Coin spending detection on self-spend
@edouardparis reports a coin is not detected as spent when it is sent to us.

I suspect it is because we don't `include_change` in the `listsinceblock` calls to get the spending transaction id.

darosior:
There were a couple cases in which we would not detect a coin as spent:
1. If the deposit transaction was still unconfirmed
2. If the spend transaction did not contain any external output

This comes with a functional test that will fail on master. It tests both above cases as well as many other.

Fixes #125


edouardparis (approved):
utACK 5721ab0fdeb5927b4a53a90d4820e30d3dcfea6a

PR #137, opened 2022-11-30
Refac spend steps

edouardparis:
- [x] Add remaining sequence to coins
 - [x] Order coins by remaining sequence or amount
 - [x] Move feerate form to coins selection step 
 - [x] Check amount left to select at coins selection step

PR #146, opened 2022-12-02
Add warning to home panel

edouardparis:
when remaining sequence is inferior to timelock*10/100, a warning is displayed in home panel
based on #137

### DIFF
--- a/gui/src/app/view/coins.rs
+++ b/gui/src/app/view/coins.rs
@@ -170,8 +170,14 @@ fn coin_list_view(
                                     Column::new()
                                         .push(
                                             Row::new()
+                                                .align_items(Alignment::Center)
                                                 .push(text("Outpoint:").small().bold())
-                                                .push(text(format!("{}", coin.outpoint)).small())
+                                                .push(Row::new().align_items(Alignment::Center)
+                                                    .push(text(format!("{}", coin.outpoint)).small())
+                                                    .push(Button::new(icon::clipboard_icon())
+                                                        .on_press(Message::Clipboard(coin.outpoint.to_string()))
+                                                        .style(button::Style::TransparentBorder.into())
+                                                    ))
                                                 .spacing(5),
                                         )
                                         .push_maybe(coin.block_height.map(|b| {


gui/src/app/state/mod.rs
```diff
@@ -82,34 +86,50 @@ impl State for Home {
...
-                            })
-                            .sum(),
-                    );
+                    self.balance = Amount::from_sat(0);
+                    let mut recovery_warning = (Amount::from_sat(0), 0);
+                    let mut recovery_alert = (Amount::from_sat(0), 0);
+                    for coin in coins {
+                        if coin.spend_info.is_none() && coin.block_height.is_some() {
+                            self.balance += coin.amount;
+                            let timelock = daemon.config().main_descriptor.timelock_value();
```
darosior, line 116: There is no need to call that in the loop since it does not depend on each specific coin. In addition it's a bit expensive so it can make the update computation heavy on a wallet with a substantial amount of coins.

gui/src/app/state/mod.rs
```diff
@@ -82,34 +86,50 @@ impl State for Home {
...
+                            let seq = remaining_sequence(&coin, cache.blockheight as u32, timelock);
+                            if seq == 0 {
+                                recovery_alert.0 += coin.amount;
+                                recovery_alert.1 += 1;
+                            } else if seq < timelock * 10 / 100 {
```
darosior, line 121: Why `* 10 / 100` instead of just `/ 10`?

Also technically it'd be more correct to
```suggestion
                            } else if seq * 10 < timelock {
```
So that it doesn't truncate. But no big deal.

gui/src/app/view/coins.rs
```diff
@@ -74,36 +74,43 @@ fn coin_list_view(
...
+                                                .spacing(5)
+                                                .push(text(" 0").small().style(color::ALERT))
+                                                .push(
+                                                    icon::hourglass_done_icon()
+                                                        .small()
+                                                        .style(color::ALERT),
+                                                )
+                                                .align_items(Alignment::Center),
+                                        ))
+                                    } else if seq < timelock * 10 / 100 {
```
darosior, line 90: Would be nice to merge this and the above computation so that only one place needs to be updated should a contributor want to modify this in the future?

darosior:
Woops, too late.

PR #148, opened 2022-12-02
descriptors: add a method to get the size of a spending input

darosior:


darosior:
I need to add a fixed value to the unit test

darosior:
Updated the test and manually checked the value given is correct. @edouardparis double check but it should be readily usable by the GUI now.

edouardparis (approved):
utACK eebc96c4d53f0f6a86d5ddf4267b3d4a45aa7e3f

PR #154, opened 2022-12-06
gui: cleanup the "amount left to select" calculation

darosior:
It don't think it was blatantly wrong, but neither was it entirely correct. And it was seriously lacking comments explaining what was going on for such a tricky calculation.

In order to make sure it's correct, and make this more code more maintainable (and readable to future contributors), clean it up by using a more obviously-correct approach: a transaction template that is filled with inputs depending on the coins that were selected.

I DID NOT TEST IT!

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -5,7 +5,9 @@ use std::sync::Arc;
 use iced::{Command, Element};
 use liana::{
     config::Config as DaemonConfig,
-    miniscript::bitcoin::{util::psbt::Psbt, Address, Amount, Denomination, OutPoint, Script},
+    miniscript::bitcoin::{
+        self, util::psbt::Psbt, Address, Amount, Denomination, OutPoint, Script,
+    },
 };
 
 use crate::{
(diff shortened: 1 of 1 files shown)


gui/src/app/state/spend/step.rs
```diff
@@ -223,44 +225,57 @@ impl ChooseCoins {
...
+            output: self
+                .recipients
+                .iter()
+                .map(|(addr, amount)| bitcoin::TxOut {
+                    script_pubkey: addr.script_pubkey(),
+                    value: amount.to_sat(),
+                })
+                .collect(),
+        };
+        const CHANGE_TXO_SIZE: usize = 8 + 1 + 32;
```
darosior: Technically it should be
```suggestion
        const CHANGE_TXO_SIZE: usize = 8 + 1 + 1 + 1 + 32;
```

darosior: Done in latest push.

edouardparis (approved):
ACK 493dbde666b73097848a5a4c7d10ee1ad7cfeab2
tested and approved

PR #166, opened 2022-12-08
deps: make libc optional

edouardparis:


darosior:
Also, how can this ever compile if `libc` isn't feature-gated in the code? https://github.com/revault/liana/blob/295e5a31a06b3d7431bf154c6797ea27ce06b10c/src/daemonize.rs#L8-L23

Also, note that the JSONRPC server needs libc too...
https://github.com/revault/liana/blob/295e5a31a06b3d7431bf154c6797ea27ce06b10c/src/jsonrpc/server.rs#L193-L208

edouardparis:
I updated the commit message

edouardparis:
Do not merge

darosior:
Could you *please* answer my questions? `cargo build --no-default-features` trivially fails now. This change is obviously missing feature gates on calls to libc!

edouardparis:
fixed and rebased

Cargo.toml
```diff
@@ -12,16 +12,17 @@ exclude = [".github/", ".cirrus.yml", "tests/",  "test_data/", "contrib/", "pypr
...
 [[bin]]
 name = "liana-cli"
 path = "src/bin/cli.rs"
 required-features = ["jsonrpc_server"]
 
 [features]
-default = ["jsonrpc_server"]
+default = ["jsonrpc_server", "daemon"]
 jsonrpc_server = []
+daemon = ["libc"]
```
darosior, line 25: I think it would be cleaner to get rid of the `jsonrpc_server` feature then. It implies `daemon` (and `daemon` implies it).

```suggestion
default = ["daemon"]
daemon = ["libc"]
```

darosior (commented):
The newly introduced `daemon` feature is equivalent to `jsonrpc_server` so i think you want to rename it into `daemon` and add `libc` as a requirement for it.

Cargo.toml
darosior: And also the dependency for `jsonrpc_server` is wrong: it needs libc.

darosior:
ACK 4bf5234ce97fa725c076be628c12b7289466c1f1

Tested with `cargo test --no-default-features` locally too.

PR #167, opened 2022-12-08
gui: add sat/vbyte label in feerate form

edouardparis:
Feedback from Kevin, it is not clear for the user which units the feerate form requires.

### DIFF
--- a/gui/src/app/view/spend/step.rs
+++ b/gui/src/app/view/spend/step.rs
@@ -124,10 +124,10 @@ pub fn choose_coins_view<'a>(
             .push(text("Choose coins and feerate").bold().size(50))
             .push(
                 Container::new(
-                    form::Form::new("Feerate", feerate, move |msg| {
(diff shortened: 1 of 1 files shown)


gui/src/app/view/spend/step.rs
```diff
@@ -124,10 +124,10 @@ pub fn choose_coins_view<'a>(
             .push(text("Choose coins and feerate").bold().size(50))
             .push(
                 Container::new(
-                    form::Form::new("Feerate", feerate, move |msg| {
+                    form::Form::new("Feerate sat/vbyte", feerate, move |msg| {
```
darosior, line 127: nit
```suggestion
                    form::Form::new("Feerate (sat/vbyte)", feerate, move |msg| {
```

edouardparis (commented):
Self-ACK 4783dc0f1b72b1c516e6ddb81f895217a292cd88

Issue #69, opened 2022-10-20
`macos-latest` + `rustc 1.48` = ` error: linking with cc failed: exit code: 1`


darosior:
The rust 1.48 build started failing on macOS in #67.

```
   Compiling proc-macro2 v1.0.40
   Compiling cc v1.0.73
   Compiling quote v1.0.20
     Running `rustc --crate-name build_script_build --edition=2018 /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.40/build.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="proc-macro"' -C metadata=a775773b1c4d6d13 -C extra-filename=-a775773b1c4d6d13 --out-dir /Users/runner/work/minisafe/minisafe/target/debug/build/proc-macro2-a775773b1c4d6d13 -L dependency=/Users/runner/work/minisafe/minisafe/target/debug/deps --cap-lints allow`
     Running `rustc --crate-name cc --edition=2018 /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/cc-1.0.73/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 -C metadata=6951f41806db0446 -C extra-filename=-6951f41806db0446 --out-dir /Users/runner/work/minisafe/minisafe/target/debug/deps -L dependency=/Users/runner/work/minisafe/minisafe/target/debug/deps --cap-lints allow`
     Running `rustc --crate-name build_script_build --edition=2018 /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/quote-1.0.20/build.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="proc-macro"' -C metadata=43d0865adc1d5c42 -C extra-filename=-43d0865adc1d5c42 --out-dir /Users/runner/work/minisafe/minisafe/target/debug/build/quote-43d0865adc1d5c42 -L dependency=/Users/runner/work/minisafe/minisafe/target/debug/deps --cap-lints allow`
   Compiling unicode-ident v1.0.2
[...]

PR #168, opened 2022-12-08
ci: re-enable ci on non-1.48 for MacOS

darosior:
So we at least have one MacOS job. Because it looks like #69 isn't going anywhere.


darosior:
self-ACK 0cf6f3214763109a2e2f7096945a40e8397be9a6 -- trivial.

PR #177, opened 2022-12-08
Check for key duplicates when creating a descriptor

### LINKED ISSUE #149: Don't allow duplicate keys
While it's not technically an issue with our single keys template, it is invalid Miniscript. We shouldn't allow that.

darosior:
Fixes #149.

Also contains a drive-by fix for the sat size API in the second commit.


darosior:
self-ACK 9cc748cd2acb3338f230378b827104fee4e56318

@edouardparis note this is a minor break of the API used by the GUI for the tx size calculation.

PR #178, opened 2022-12-08
Cargo: fix the rust-miniscript patched dependency for downstream

darosior:
This new syntax would actually force all downstream users to apply the patch section in their Cargo.toml! Revert to the previous syntax of specifying the git repo directly.


darosior:
self-ACK 293c28570ca28e1375d424245a907c8d0b7e2f85

PR #183, opened 2022-12-08
fix installer: import_descriptor next button

edouardparis:
Next button was grey although the descriptor form
is not empty.


edouardparis (commented):
Self-ACK 2d788d5bffc1b237dc6be3c09ca2e13afcdeca67

Issue #200, opened 2022-12-09
GUI: change is incorrectly displayed


darosior:
From @kloaec:
> Why do i have -0,00438 when i'm actually spending 0.003 (see spend flow video from before)

(see https://github.com/revault/liana/issues/199 for the spend flow mentioned)

![image](https://user-images.githubusercontent.com/22457751/206679504-793ec968-72e4-43fb-bead-3af56bdcfa9d.png)

darosior:
cc @kloaec: i'm not entirely sure of what is the issue here.

edouardparis:
I believe it is all about confusion because of the wrong event selection

PR #201, opened 2022-12-09
fix gui: wrong event selection on home panel

edouardparis:
Because pending events are in a different list
that confirmed event. The wrong list item was
selected when user tried to click on a pending
event.


darosior:
Nice. Does this close any of the issues i opened this morning?

edouardparis:
it's related to #200
In the video the wrong event is opened that's why the amounts are not the same

darosior:
But it doesn't fix it entirely? Or does it?

edouardparis (commented):
Self-ACK ea6d4f9b6374ea9e8873a4eb768feef11caaa48f

PR #207, opened 2022-12-12
Add broadcasttx command

edouardparis:
`createrecovery` creates a psbt that needs
to be broadcasted after being signed.
`broadcastspend` cannot be used because the
recovery transaction is not store as spend tx.


edouardparis:
finally no use for this command, updatespend and broadcastspend can do the trick

PR #221, opened 2022-12-13
just put the whole gpg key

kloaec:
I'm quite sure it does the check for you. So i also removed the extra comment. If i'm doing something wrong plz educate me :D


darosior (commented):
I wouldn't be surprised if gpg was doing something surprising like transforming the key into a fingerprint before pulling from the keyserver. On the other hand it's the only bit of additional information re: checking GPG signatures in this document so being consistent is probably better. ACK 52341db41d719f7cbb224d9164593f5f2992eae7

PR #212, opened 2022-12-13
Windows reproducible build

darosior:
Using Docker for now. I'll add documentation once i've figured out macOS.

PR #224, opened 2022-12-14
Windows watchonly wallet path workaround

darosior:
Turns out that providing an absolute path to bitcoind when loading a wallet on Windows results in surprising behaviour.


darosior:
@edouardparis see the last commit since it is still based on #212.

edouardparis (approved):
ACK 48e4909e7ad83c21f788802953327f8041a69b7d

PR #225, opened 2022-12-14
gui: fix minimal loading sync value

### LINKED ISSUE #217: Roundup the sync progress in both the daemon poller and the GUI
If it's 0.999 consider it is `1.0`.

edouardparis:
bitcoind sync may stuck for a while at
0.999. At this state, gui can start the
application.

close #217

### DIFF
--- a/gui/src/loader.rs
+++ b/gui/src/loader.rs
@@ -143,7 +143,7 @@ impl Loader {
             } => {
                 match res {
                     Ok(info) => {
-                        if (info.sync - 1.0_f64).abs() < f64::EPSILON {
+                        if info.sync >= 0.999_f64 {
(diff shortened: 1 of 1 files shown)


darosior:
It doesn't close https://github.com/revault/liana/issues/217 since i need to do it in the poller too.

darosior:
I tried this but i can't manage to reproduce the bug on master anymore..

darosior:
Actually i think fixing this on the daemon is enough? After #227 checking for 1.0 should be enough? Sorry for suggesting you should modify the GUI in #217.

edouardparis:
> Actually i think fixing this on the daemon is enough?

Certainly yes, I never had the problem on my side.

PR #228, opened 2022-12-15
fix recovery state coins filtering

edouardparis:
Recoverable coins are coins that recovery
path is available *next* block


darosior (commented):
Not familiar with this code, but it appears it now check whether the timelock will be mature in one block or less. If that is the case then it is consistent with the behaviour of the daemon. utACK192072ea257ae50e6c90bda00677b5f07ca9ada9.

edouardparis (commented):
ACK 192072ea257ae50e6c90bda00677b5f07ca9ada9

PR #241, opened 2022-12-16
Clippy fixes and version pin

darosior:
The CI broke again because clippy is configured to target `stable` and introduces breaking changes with each version. Apply the fixes for the latest stable, then pin clippy to this version.

For the GUI i wasn't sure so i just pinned clippy to the latest stable.


darosior:
self-ACK f996e2565797b413870d75fa764ebd207bb264c4 -- trivial

PR #250, opened 2022-12-16
fix various bugs with hws and spend signing process

### LINKED ISSUE #238: GUI?: Bug - hardware is connected but not found after locking/unlocking it
Trying to find the exact issue and reproduce it, will add in comments when i figure out wtf.

Screenshot from 2022-12-15 15-00-42

### LINKED ISSUE #246: GUI: in the spend process we would like a refresh button even if there is already a hardware wallet
image
I wanted to use my legder but there was no way to refresh, we need a refresh button even when another HW is connected.

edouardparis:
Fixes #246 
Fixes #238

### DIFF
--- a/gui/src/app/state/recovery.rs
+++ b/gui/src/app/state/recovery.rs
@@ -118,8 +118,13 @@ impl State for RecoveryPanel {
                     }
                 }
             },
+            // We add the new hws without dropping the reference of the previous ones.
             Message::ConnectedHardwareWallets(hws) => {
-                self.hws = hws;
+                for h in hws {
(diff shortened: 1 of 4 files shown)


darosior:
I think this fixes #246?

darosior:
And also #238

darosior:
I've added the magic links to the OP for the two issues mentioned above. Please chime in if i was wrong.

edouardparis:
It is good

edouardparis (commented):
ACK 73eb1d38e3612110ad467fd608863519838aa988

PR #260, opened 2022-12-20
fix spend change detection

edouardparis:
Use the presence of bip32_derivation
in the spend outputs to detect the
change outputs.


edouardparis (commented):
Self-ACK a56cab73c5c0193a90732f3ba0c2c5be7b993928

PR #261, opened 2022-12-20
Remove change_index from listspendtxs response

edouardparis:
Multiple change indexes may be present in a spend
draft transaction and can be detected instead
in the response psbt with the bip32_derivation
outputs fields.


darosior:
utACK 103c1950ba9af114cadde6f92117b31ec584ac73

PR #262, opened 2022-12-20
contrib: fix windows archive creation in release script

darosior:
It would create an empty archive since it didn't recurse down the directories..


darosior:
ACK a0404e5d69e895467a5a51621e728b78140eae8e

PR #266, opened 2022-12-20
Check the address network during spend creation

edouardparis:


gui/src/app/state/spend/step.rs
```diff
@@ -166,18 +171,22 @@ impl Recipient {
...
-                    // Make the error disappear if we deleted the invalid address
-                    self.address.valid = true;
-                } else if Address::from_str(&self.address.value).is_ok() {
-                    self.address.valid = true;
+                if let Ok(address) = Address::from_str(&self.address.value) {
+                    self.address.valid = if network == Network::Bitcoin {
+                        address.network == Network::Bitcoin
+                    } else {
+                        address.network != Network::Bitcoin
+                    };
```
darosior, line 183: It's sufficient for safety (it's fine if they mixup test networks) but not for correctness: regtest addresses have their own prefix.

edouardparis: I think I made the error somewhere else also in the gui, which is not bad, just wrong.

darosior (commented):
utACK 9e1c16bc3fef5aa01a62cb34304cc161d4623468 check looks good to me

PR #267, opened 2022-12-20
commands: check addresses' network when creating transactions

darosior:


src/commands/mod.rs
```diff
@@ -622,6 +635,12 @@ impl DaemonControl {
         if feerate_vb < 1 {
             return Err(CommandError::InvalidFeerate(feerate_vb));
         }
+        if address.network != self.config.bitcoin_config.network {
```
edouardparis, line 638: beware that address network can be testnet, and bitcoin_config.network signet

darosior: Ah good catch! I had a note to double check that and rushed and forgot about it. Thanks.

src/commands/mod.rs
```diff
@@ -188,6 +194,22 @@ impl DaemonControl {
         };
         desc.derive(coin.derivation_index, &self.secp)
     }
+
+    // Check whether this address is valid for the network we are operating on.
+    fn validate_address(&self, addr: &bitcoin::Address) -> Result<(), CommandError> {
+        // NOTE: signet uses testnet addresses
+        if addr.network == self.config.bitcoin_config.network
+            || (addr.network == bitcoin::Network::Testnet
+                && self.config.bitcoin_config.network == bitcoin::Network::Signet)
```
edouardparis, line 203: does not handle regtest.

darosior: Regtest has its own address prefix, `bcrt`.

edouardparis (approved):
utACK ed6ff9c67b69b08e08e466102656b6765584dfce

PR #280, opened 2023-01-09
gui: display amount with sats in bold

edouardparis:
2023-01-12T09:40:35,488663454+01:00


edouardparis (commented):
Self-ACK ca096c71b52c5843b68dba0ceaeabe2755bf29a7

PR #286, opened 2023-01-13
Make daemon config optional in gui

### LINKED ISSUE #147: GUI: *actually* allow to connect to a running daemon (ie do not require lianad's config)
If there is already an existing deployment.

```
$ ls ~/.liana/testnet/
lianad_rpc  lianad.sqlite3  lianad_watchonly_wallet
$ cargo run -- --testnet
# This spawns the installer, which in turn says there already is an existing datadir for testnet...
```

edouardparis:
Because by making the daemon config optional,
the GUI must store the main_descriptor from
the getinfo request.

A new module wallet is introduced that
contains the main_descriptor.

close #147

### DIFF
--- a/gui/README.md
+++ b/gui/README.md
@@ -13,7 +13,7 @@ In addition, if you want to build the project from source, you will need:
 - [`pkg-config`](https://www.freedesktop.org/wiki/Software/pkg-config/) (On Debian/Ubuntu `apt install pkg-config`)
 
 
-## Usage
+## Common usage
 
 *For a quick guide to try out the software see [../doc/TRY.md](../doc/TRY.md).*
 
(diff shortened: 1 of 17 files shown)


darosior:
So my thinking when opening #147 was to re-enable the original usage of the GUI back on Revault when it connected to the `revaultd`. I don't remember it needing a config file with the path to `revaultd`'s RPC?

Actually, the whole point of #147 was to be able to "just run the GUI" if you have a `lianad` running already. But if we are going to require a config file before running the GUI, being able to set the RPC socket path instead of the daemon configuration file path seems only be a very mild improvement?

You started explaining it to me earlier but my brain wasn't ready, could you tell me why the GUI couldn't just check if a daemon is running at the standard location for `lianad`'s datadir before requiring a configuration file to be created for it too?

darosior:
Hmm there may be a point for not trying to connect to anything that's running by default. But i don't see one if you can opt out by setting up a config.

> but without config GUI may not be able to sign PSBTs with ledger because the GUI store the ledger hmac in its config

Wait are you saying the GUI writes into its own config? It's very bad practice: you should have a config that is written by the user and read-only for the software, and a data directory (or a DB access or whatever) where the software actually writes. In my opinion the GUI should store the HMAC as part of its data directory.

darosior:
Alright i'm starting to understand why you need the config. Indeed it's an issue for the signing device setup.

At least let's say if there is already a GUI datadir at the default location the GUI could start with the config that's stored there?

Now that i think about it by the way, it would have been nice for the GUI to create the daemon configuration at its standard path: https://github.com/wizardsardine/liana/blob/22bd8326dbdd1afb0efab6c43b5cf8946042a02d/src/config.rs#L179.

edouardparis:
This PR is deprecated and issue are created

PR #289, opened 2023-01-18
Gui multisig

edouardparis:


darosior:
nit: in the installer for the keys "Edit" button should be "Fill in" (or "Set") when it's empty, and only "Edit" after it's been set.
nit: "Name is a duplicate" -> "Duplicate alias"; nit: "Key is a duplicate" -> "Duplicate key";

In "participate in a new wallet", we should make it clear that they can click multiple times on the device button to generate more keys.

Not for this PR but in "register descriptor on signing device", we should really make it clear they *need* to click on the signing devices tabs. We could also have a checkbox here (like in descriptor backup).

In the Spend transaction details, i'm not a fan of the display of the signatures. I think it could and should be greatly simplified. We want to display three pieces of information for each spending path: the number of signatures provided, the number of signatures required and what keys can be used to provide signatures.
First of all, the information for each path should each be on its own line. Then we can reuse what you did here and in revault-gui for the keys available, that is a number of grey icons next to each other, one for each key, that turn green when a signature was provided for them. I like also that you get the alias when hovering the icon. Maybe the alias could even be made a first class citizen here. Finally, the required number of keys can simply be within a sentence. Here is an instance of what i'm thinking about with a 3-of-3 between Alice, Bob and Carol that turns into a 1-of-3 between Alice, Bob and Dave:
- 3 signatures out of `Alice`, `Bob` and `Carol`; or
- 1 signature out of `Alice`, `Bob` and `Dave`
 
Note how this generalizes well to multiple paths. For instance if we add a third path in between the two it just becomes:
- 3 signatures out of `Alice`, `Bob` and `Carol`; or
- 2 signatures out of `Alice`, `Bob` and `Carol`
- 1 signature out of `Alice`, `Bob` and `Dave`

Ideally, i think we should also optimize out the very common cases. For instance, a single key is used and a single path is available: in this case just display whether the transaction is signed or not.

edouardparis (commented):
Self-ACK ca39b15edd366d23d4bcf7433273663807fa2aab

PR #294, opened 2023-01-20
gui: store descriptor in wallet module

edouardparis:
A new module wallet is introduced that contains the main_descriptor retrieved from `getinfo`.
It removes the requirement for the `Daemon` trait to provide the demon config in the  other components of the application.
Actually BitcoindSettings cannot be edited if the gui is connected to an external daemon, this PR just remove the display of the Bitcoind Settings section.


edouardparis (commented):
Self-ACK b177f190d144986c1268c640075c89399f4c599d

PR #302, opened 2023-02-01
contrib: update the Iced patch for reproducible builds

darosior:
The Iced dependency was updated, and the patch wasn't valid anymore.


darosior:
Arg.. We need to add more patches :sob:

darosior:
ACK 7a3eeb741ea42e3af4d5414e6778bb3fa585f762 -- tested a reproducible build with this.

PR #303, opened 2023-02-01
Add liana version in gui settings

### LINKED ISSUE #243: Display liana version in the gui
Required in order to provide support

edouardparis:
close #243

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -1640,7 +1640,7 @@ dependencies = [
 [[package]]
 name = "liana"
 version = "0.1.0"
-source = "git+https://github.com/wizardsardine/liana?branch=master#f433002e91b09ab700d06026e1d94c462aca9756"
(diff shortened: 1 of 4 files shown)


darosior (commented):
Why having a different version for the gui than the one of the daemon? I see the point of displaying both the daemon and GUI version if the GUI is to connect to a running daemon of a previous version (not sure we can guarantee compatibility here though..), but why should it be a new `VERSION` than the one already existing?

https://github.com/wizardsardine/liana/blob/9510e4e06951d0d41a19b2ad995346b26b56d2d9/src/lib.rs#L66-L78

edouardparis:
I prefer to show both, because we may have to release a patch for the gui without touching the daemon for example.

darosior:
Makes sense. Let's reuse the `struct` definition at least then? Feel free to add a third leg to the existing one.

PR #304, opened 2023-02-02
gui: reorganize imports and mod

trigger67:
To make it easy to read, and to code, I followed theses rules :
 - Start with modules, then import crates
 - Start with public then go with private
 - No empty lines, except between the previous groups
 - Group imports when possible
 
Rustfmt will order each group of code alphabetically, prioritizing rust specific keywords in imports (ex: use super::, use crate::)

Example:

```
pub mod home;
pub mod settings;

mod message;
mod util;

pub use message::*;

use crate::{
    app::{cache::Cache, error::Error, menu::Menu, wallet::Wallet},
    daemon::Daemon,
};
use iced::{
    widget::{self, scrollable, Button, Column, Container, Row},
    Element, Length,
};
```

### DIFF
--- a/gui/src/app/error.rs
+++ b/gui/src/app/error.rs
@@ -1,7 +1,6 @@
 use crate::daemon::DaemonError;
 use liana::config::ConfigError;
-use std::convert::From;
-use std::io::ErrorKind;
+use std::{convert::From, io::ErrorKind};
 
 #[derive(Debug)]
 pub enum Error {
--- a/gui/src/app/message.rs
+++ b/gui/src/app/message.rs
@@ -1,15 +1,14 @@
+use crate::{
+    app::{error::Error, view},
+    daemon::model::*,
+    hw::HardwareWallet,
+};
 use liana::{
     config::Config as DaemonConfig,
     miniscript::bitcoin::{
         util::{bip32::Fingerprint, psbt::Psbt},
         Address,
     },
-};
-
-use crate::{
-    app::{error::Error, view},
-    daemon::model::*,
-    hw::HardwareWallet,
 };
 
 #[derive(Debug)]
--- a/gui/src/app/mod.rs
+++ b/gui/src/app/mod.rs
@@ -9,24 +9,17 @@ pub mod wallet;
 
 mod error;
 
-use std::fs::OpenOptions;
-use std::io::Write;
-use std::sync::Arc;
-use std::time::Duration;
-
-use iced::{clipboard, time, Command, Element, Subscription};
-
-pub use liana::config::Config as DaemonConfig;
-
 pub use config::Config;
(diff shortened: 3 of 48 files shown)


edouardparis:
Can you share a link to a documentation explaining these guideline ?  thanks

darosior:
Hey, thank you for your pull request.

I guess what Edouard is getting at here is about the rationale for this PR. Changing 48 (!!) files in a single PR incurs a considerable burden to contributors of an Open Source project in general, as it has a good chance of conflicting with every single other branch out there people might be working on (whether it is published as a PR or not). A large set of similar changes is also very error prone and makes review difficult.

Granted, we are still a small project so the impact is limited. But still, i'd like us to hold ourselves to a high bar. Such invasive changes should come with a strong rationale. Therefore i concur with Edouard: is the motivation only that you find it clearer this way or is it a widely agreed-upon guideline across Rust projects? If the latter it might be worth biting the bullet once here along with adding it to the contributing guidelines as Edouard suggests. Otherwise, i'm afraid it's not worth it. We can't afford having everyone applying their style nits to the whole project when they start contributing. :)

trigger67:
Hey I totally get I, I hoped there is not too many different people and branches at the moment, so I thought it could be better to do it now than later. Also I understand if that's not a priority and you want to skip/shift this PR.

Unfortunately I did not find a lot of guidelines about that topic (more below), but what I saw is that rustfmt is reordering imports and modules by default (see [here](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#reorder_imports)).
So to take advantage of this, I think it's good to groups imports (not put empty lines between them, as today). With that developers do not have to do it manually, and can just let rustfmt order each groups individually.

But what I find disturbing, is that rustfmt will mix modules, use statements, public modules, and public use. So I thought just adding some small rules to split them in different categories will make things better.

I thought modules first, and imports second, could be a good option. Then I looked on the internet and took the following advice from [this post](https://www.reddit.com/r/rust/comments/wwbxhw/how_do_you_organize_imports_reexports_structs/ ) : 
- place all mod before any use
- pub is more important than private, so pub things go first


Having a standard will make things easier in the future, especially when there are conflicts to manage.
But I agree it would be good if that standard is common between many rust projets (ex: there is no option to put the public before private in rustfmt).
Looking at many different open source projects, I did not find any standard, but I saw that many projects are at least creating "groups of imports", sometimes without logic, or with a different logic for every files, but at least there are some groups that are correctly ordered.

Not to have a big PR like this, I can remove my changes and work on CONTRIBUTING.md if you'd like

edouardparis:
Sorry, I have to close this PR as the conflicts are way too big. But we can move to discuss about style guideline in an issue :smiley:

PR #305, opened 2023-02-02
Add patch to liana version

edouardparis:
The struct will be imported by the gui which
may need one day to increase the patch version number.


darosior (commented):
ACK 5e79c05c544a573a9030e64564c412d95ac34f57

PR #307, opened 2023-02-03
Change HardwareWallet for (Un)Supported enum

### LINKED ISSUE #252: GUI: sanity check the version of the Bitcoin app on the Ledger device
This should be a priority for the next version, as i expect most users won't have the latest app installed and will be confused about why they were able to install the wallet (if they missed the descriptor registration step (see #187, #247), the installer wouldn't fail with an earlier version of the Bitcoin app on the Ledger) but not able to use the wallet.

Not specific to this but generally we should thrive to have failures as early as possible in the process: if a user won't be able to use the wallet, the installation should fail in the first place.

edouardparis:
close #252

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -145,9 +145,9 @@ dependencies = [
 
 [[package]]
 name = "async-hwi"
-version = "0.0.2"
+version = "0.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


edouardparis:
2023-02-03T11:35:17,007212521+01:00

edouardparis (commented):
Self-ACK 2a9bcf92d031510e8c1953a80ef8cd3e13b39a7d

PR #308, opened 2023-02-03
Handle unconfirmed deposits that were replaced

### LINKED ISSUE #72: Handle replacement of deposit transactions
I think at the moment if a deposit transaction gets RBF'd for instance, we'd just keep an unconfirmed coin forever. We can do better.

darosior:
Previously we would not detect whether the transaction for an unconfirmed deposit was still in our mempool. For instance a deposit that was RBF'd would result in us storing two coins, the replaced and the replacement. Keeping the unconfirmed replaced coin forever.

Fix this by dropping unconfirmed coins from our database if their transaction isn't in our mempool (anymore).

Fixes #72.

### DIFF
--- a/src/bitcoin/mod.rs
+++ b/src/bitcoin/mod.rs
@@ -58,11 +58,12 @@ pub trait BitcoinInterface: Send {
         descs: &[descriptors::InheritanceDescriptor],
     ) -> Vec<UTxO>;
 
-    /// Get all coins that were confirmed, and at what height and time.
+    /// Get all coins that were confirmed, and at what height and time. Along with "expired"
+    /// unconfirmed coins (for instance whose creating transaction may have been replaced).
     fn confirmed_coins(
         &self,
         outpoints: &[bitcoin::OutPoint],
-    ) -> Vec<(bitcoin::OutPoint, i32, u32)>;
+    ) -> (Vec<(bitcoin::OutPoint, i32, u32)>, Vec<bitcoin::OutPoint>);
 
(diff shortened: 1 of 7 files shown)


darosior:
I stressed test a bit the functional tests to make sure this didn't introduce any flakiness, especially in the reorg tests. Just force-pushed to correct the title of a commit.

src/bitcoin/mod.rs
```diff
@@ -58,11 +58,12 @@ pub trait BitcoinInterface: Send {
         descs: &[descriptors::InheritanceDescriptor],
     ) -> Vec<UTxO>;
 
-    /// Get all coins that were confirmed, and at what height and time.
+    /// Get all coins that were confirmed, and at what height and time. Along with "expired"
+    /// unconfirmed coins (for instance whose creating transaction may have been replaced).
     fn confirmed_coins(
```
edouardparis, line 63: Maybe rename the fn now for confirmed_and_expired_coins
or maybe we can find a way to have `confirmed_coins` and `expired_coins` sharing the same cache

edouardparis (approved):
utACK 3a573b695e07d357e844dd9d573d578e05ecebed

src/bitcoin/mod.rs
darosior: Yeah the name could be updated. I'm going to keep it as is for now since it's documented at least in the comment.

For the cache it could be nice to have but then you get into managing its size, evictions, etc.. Until it becomes a bottleneck i don't think it's worth the effort / complexity.

PR #313, opened 2023-02-03
gui: fix hws list reloading

edouardparis:
I wanted the refresh button to add more harware wallet to the list without erasing the previous list
in order to not cut the existing communication channels.

But with the new display of "unsupported wallet" this behaviour can corrupt the list. An "unsupported wallet" can become "supported" because user switch app or unlock.

It is preferable to reset the full list and drop the objects after a refresh.


edouardparis (commented):
Self-ACK ea4f02172824371ef162db9ee0b974ef4597a541

PR #314, opened 2023-02-03
gui: change path signature details

edouardparis:
2023-02-03T21:50:42,669968940+01:00
2023-02-03T21:50:50,459471889+01:00
2023-02-03T21:58:48,281389762+01:00

### DIFF
--- a/gui/src/app/view/spend/detail.rs
+++ b/gui/src/app/view/spend/detail.rs
@@ -287,26 +287,54 @@ pub fn signatures<'a>(
     keys_aliases: &'a HashMap<Fingerprint, String>,
 ) -> Element<'a, Message> {
     Column::new()
-        .push(Collapse::new(
+        .push(
+            if let Some(sigs) = tx.path_ready() {
+            Container::new(
+                Scrollable::new(
+                    Row::new()
+                        .spacing(5)
+                        .align_items(Alignment::Center)
+                        .push(icon::circle_check_icon().style(color::SUCCESS))
+                        .push(text("Ready").bold().style(color::SUCCESS))
+                        .push(text(", signed by"))
+                        .push(
+                            sigs.signed_pubkeys
+                            .keys()
+                            .fold(Row::new().spacing(5), |row, value| {
+                                row.push(if let Some(alias) = keys_aliases.get(value) {
+                                Container::new(
+                                    tooltip::Tooltip::new(
+                                        Container::new(text(alias))
+                                            .padding(3)
+                                            .style(badge::PillStyle::Simple),
+                                            value.to_string(),
+                                            tooltip::Position::Bottom,
+                                    )
+                                    .style(card::SimpleCardStyle),
+                                )
+                            } else {
+                                Container::new(text(value.to_string()))
+                                    .padding(3)
+                                    .style(badge::PillStyle::Simple)
+                            })
+                            }),
+                    )
+                ).horizontal_scroll(scrollable::Properties::new().width(2).scroller_width(2))
+            ).padding(15)
+        } else{
+            Container::new(
+            Collapse::new(
             move || {
                 Button::new(
                     Row::new()
                         .align_items(Alignment::Center)
-                        .push(if tx.is_ready() {
-                            Row::new()
-                                .spacing(5)
-                                .align_items(Alignment::Center)
-                                .push(icon::circle_check_icon().style(color::SUCCESS))
-                                .push(text("Ready").bold().style(color::SUCCESS))
-                                .width(Length::Fill)
-                        } else {
-                            Row::new()
+                        .push(Row::new()
                                 .spacing(5)
                                 .align_items(Alignment::Center)
                                 .push(icon::circle_cross_icon())
                                 .push(text("Not ready").bold())
                                 .width(Length::Fill)
-                        })
+                        )
                         .push(icon::collapse_icon()),
                 )
                 .padding(15)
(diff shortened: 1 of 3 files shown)


darosior:
Thanks. It's much nicer now.

From my suggestion i see you dropped the "or" at the beginning of each line. Didn't it render nicely?

What about dropping the green color for signers that signed? It was useful when it wasn't explicited they signed but now it's literally written next to it so it seems unnecessarily heavy?

Maybe we should also qualify the aliases. Something like "2 more signatures from signers `Specter` and `Edouard Ledger`" rather than "2 more signatures from `Specter` `Edouard Ledger`". This is kinda getting in nit territory but my concern is that we shouldn't assume these labels are meaningful aliases set by the user, if you replace `Specter` by `326475a9` and `Edouard Ledger` by `09cc411a` does the sentence still hold?
> 2 more signatures from `326475a9` `09cc411a`

With my suggestion it'd become
> 2 more signatures from signers `326475a9` and `09cc411a`

Yeah i don't know it's not a bunch clearer either. It looks like nothing will make fingerprints clear in the UI (apart maybe a whole and heavy "signer with id `326475a9`" instead of just showing the fingerprint.

I have more suggestions that we won't be able to do before the next release. Let's discuss only the pressing fixes here.

edouardparis:
maybe `key:326475a9`

darosior:
I'd rather use "signer" thank "key" because the fingerprint does identifiy a single key.

darosior:
So we are grouping keys by signers, but not signatures. So it can lead to confusing displays like "need 5 signatures from Alice, Bob and Carol" if a device is used multiple times in the same spending path. For instance here (ignore my retarded third alias):
image

So maybe we could instead display something like "Need X more signers to participate in the transaction" instead of "Need Y signatures".

On the other hand it would be an invasive change at this point and for a simple multisig it's probably useless in the first place to have one device several times on a single spending path.

kloaec:
Ok there is a lot to cover here. I'll try to go chronologically from previous comments, then add my own.

`From my suggestion i see you dropped the "or" at the beginning of each line. Didn't it render nicely?`
I personally don't care, either works too. We can put that in later UX improvements.

`Something like "2 more signatures from signers`
`I'd rather use "signer" thank "key" because the fingerprint does identifiy a single key.`
Something i repeatedly said at the last retreat, we have a big wording issue with the Fingerprint identifiers. I absolutely think we are doing the right thing by using Fingerprint labels, but indeed the words "keys", "signers", "device" aren't good for describing it.
I absolutely NACK the change for "Signer", as a single HW can have multiple fingerprints, typically for any password protected mnemonic. Some other HW can have multiple mnemonics loaded to them, even outside of the realm of password derivation.
If you need a word, we first need to find one. "Fingerprint", "mnemonic", "seed" are all correct but at the same time bad/confusing. Signer, Key, Account etc are just not describing the actual thing.

`maybe key:326475a9 or key#326475a9`
NACK for the same reason.

`So we are grouping keys by signers, but not signatures. So it can lead to confusing displays like "need 5 signatures from Alice, Bob and Carol" if a device is used multiple times in the same spending path.`
I cannot see any reason why a single mnemonic should be reused in a single path. This should actually be prevented IMHO.
(again, Fingerprint is mnemonic, not device. I could imagine using the same device, requiring multiple keys that you load separately (ledger with passphrase, Specter with different mnemonics, etc etc)

For my own comments:
- if it's ready, the tx doesn't need anything more. So I'd just display: "Ready, signed by `bob` `carol`
- Nit: i'd replace `Ed Ledger` `Specter` already signed by "already signed by `Ed Ledger` `Specter` . We might even find better wording, i just don't have it now.
- i'm getting very triggered by the lack of consistency between buttons. Ignore this comment for now but please take this into account in the future, i keep saying it.
- We can keep the "1 signature(s) from `ed ledger` " for the single case, we can change that later.
- We can keep the "2 signature(s) from `ed ledger` `potato` `chips` for the multi case, might change that later.
- I would change the "finalizing this transaction needs either" by "finalizing this transaction requires". Not I don't mind about keeping the Either there or using Or later.

Thanks for the hard work, we are getting there. It's hard to design something like that!

darosior:
Kevin, i agree with regard to the words "device" or "key" not being representative of a fingerprint. But "signer" is. It is a common abstract role in Bitcoin terminology. For instance, see the [signer role in BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#user-content-Signer) (PSBTs). The word "mnemonic" (contrary to "seed", which do keep in mind is not the same thing) you then bring up does not always represent a signer since you may derive multiple signers by using no, or different, passwords. I also don't understand why you then throw vaguely described unsupported signing devices ("Some other HWs") into the mix.

Anyways, i agree with you "signer" is not perfect and maybe not the right word to use in a user interface. But to be frank we are fighting a time-wasting useless battle here. It was just an off-hand note on displaying the fingerprint being confusing. We are not going to change that in this PR, one day from the release.

> I cannot see any reason why a single mnemonic should be reused in a single path. This should actually be prevented IMHO.

Me neither. But again as a general rule of thumb i don't think we should just forbid every single thing we can't come up with a use for. For the simple reason that we are not omniscient, but also because it can come back to bite us in the future.

> if it's ready, the tx doesn't need anything more. So I'd just display: "Ready, signed by bob carol

I agree (as i posted on Mattermost). We shouldn't display the details if the transaction is fully signed. At least for now it saves us some things to figure out. Are you fine with that @edouardparis?

> Nit: i'd replace Ed Ledger Specter already signed by "already signed by Ed Ledger Specter . We might even find better wording, i just don't have it now.

+1 for this and the rest of your comment basically.

edouardparis:
revault-gui-stakeholder
2023-02-06T11:05:25,672443226+01:00
2023-02-06T11:05:37,040872138+01:00
2023-02-06T11:05:48,257467942+01:00

darosior:
Screenshots look good to me!

edouardparis:
In order to not slow down the release process, a first state of the work is merged.
People are invited to give feedback and continue discussions at #327

Self-ACK 56d3522f8c74ff774cb3ecdbf76b48dd216497f9

PR #316, opened 2023-02-04
fix: spend minimal amount in btc

edouardparis:
"> 5000 sats" message was confusing with the actual amount in btc requirement


edouardparis (commented):
Self-ACK 263f00150d37908d97bb491cd0dbc2180f36b654

PR #318, opened 2023-02-04
fix gui: use address.is_valid_for_network

edouardparis:
Use this nice helper from rust-bitcoin, instead
of wrong checks.


edouardparis (commented):
Self-ACK c7dd94b7488ddfbe96b91a9eb1f0430c7c0b3f3a

PR #319, opened 2023-02-04
fix: drop connections to hws in share xpub panel

edouardparis:
Use needs to be able to start new connections
in the register descriptor step.


edouardparis (commented):
Self-ACK 42940f638598b02acf1cc4f175703725a2ef3cf2

PR #332, opened 2023-02-06
contrib: fix dmg creation in the release script

darosior:
The binary need to be named the same as the configured application name.


darosior:
self-ACK fa518af53d8b8a0b2f8a67072d173dce662e8a5b

PR #341, opened 2023-02-08
Add wallet path in bitcoind errors related to wallet

edouardparis:


src/bitcoin/d/mod.rs
```diff
@@ -51,13 +51,10 @@ pub enum BitcoindError {
     Server(jsonrpc::error::Error),
     /// They replied to a batch request omitting some responses.
     BatchMissingResponse,
-    WalletCreation(String),
-    DescriptorImport(String),
-    WalletLoading(String),
-    MissingOrTooManyWallet,
+    /// Error while managing wallet.
+    Wallet(String, WalletError),
```
darosior, line 55: nit
```suggestion
    Wallet(String /* watchonly wallet path */, WalletError),
```

src/bitcoin/d/mod.rs
```diff
@@ -98,15 +95,7 @@ impl std::fmt::Display for BitcoindError {
...
-            BitcoindError::WalletCreation(s) => write!(f, "Error creating watchonly wallet: {}", s),
-            BitcoindError::DescriptorImport(s) => write!(
-                f,
-                "Error importing descriptor. Response from bitcoind: '{}'",
-                s
-            ),
-            BitcoindError::WalletLoading(s) => {
-                write!(f, "Error when loading watchonly wallet: '{}'.", s)
-            }
+            BitcoindError::Wallet(_, e) => e.fmt(f),
```
darosior, line 98: We should not ignore the path of the watchonly wallet file. It's always useful if there is an error with it.

```suggestion
            BitcoindError::Wallet(path, e) => write!(f, "Watchonly wallet (path: {}) error: {}", e),
```

darosior (commented):
Looks good to me (untested) if that works for what the GUI needs. Just please always log the path in the error.

PR #347, opened 2023-02-13
daemonize: switch the pid and log file arguments

darosior:
The caller and the definition had them swapped.. So it was logging to the PID file and writing the PID to the log file..


darosior:
@edouardparis in case you had noticed and were wondering

darosior:
self-ACK f10c35e964c74a9b60df8cd03ef084f6a47c53e6 -- trivial

PR #351, opened 2023-02-16
Gui unconfirmed badge

edouardparis:
2023-02-16T10:39:22,026408712+01:00


edouardparis (commented):
Self-ACK 9535b802584a153fd0e6fcb808d8af7aa067ec95

darosior (commented):
Cool. The tooltip is not appropriate for an outgoing transaction though:
> Do not treat this as a payment until it is confirmed

But it's still an improvement

PR #352, opened 2023-02-16
Transaction creation sanity checks fixes

### LINKED ISSUE #343: Sanity checks are hit when creating a tx with a too large feerate
Reported by @kloaec 
> Happens with 555 feerate. I think I remember such an error before
image

darosior:
Fixes #343

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -54,6 +54,7 @@ pub enum CommandError {
         /* out value */ bitcoin::Amount,
         /* target feerate */ u64,
     ),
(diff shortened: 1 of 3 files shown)


darosior:
Now that i coded it up i'm actually thinking it's not worth having the second commit. It's redundant with the check in the sanity checks. We could just have the more precise error be returned by the sanity checks. What do you think?

darosior:
Alright, i did that.

edouardparis (approved):
utACK 518eb91064ac4120792b71cec8f52eb26e359b89

PR #355, opened 2023-02-17
Clippification: a little bit of housekeeping

darosior:
FIx clippy lints and bump the clippy version for the daemon. Also fix some lints for the GUI (there are some false positives that couldn't be fixed though).


darosior:
self-ACK ebf5a294ad5eeb9359625e941013809f5544672b -- trivial

PR #374, opened 2023-03-23
Cleanup the `descriptors` module

darosior:
We've been piling a bunch of new features since this module was first architectured, and it has become messy. This led to duplicate code, a confusing interface (`InheritanceDescriptor`, `LianaDescInfo`, ..) and more importantly bugs.

This is a complete re-organization of the module in view of introducing multi-paths descriptors soon. This PR contains a bugfix but aside from that it should not change (correct) behaviour. It does however completely break the interface.

The new interface makes a lot more sense:
- A `LianaPolicy` representing a Liana spending policy, from which you can get the parameters for the various spending path, and you can create from those parameters.
- A `LianaDescriptor` which can be created from a `LianaPolicy`, and from which you can infer a `LianaPolicy` to retrieve the parameters of each spending path.

This bijection (although it will soon become a surjection as we'll introduce the Miniscript policy compiler to create a `LianaDescriptor` from a `LianaPolicy`) makes the life of a client of the API easier, but it also harmonizes the code: we've centralized the Miniscript Semantic Policy checks of a descriptor in a single place to make sure that we can parse only what, and all, descriptors we can create.


src/descriptors/keys.rs
```diff
@@ -0,0 +1,137 @@
...
+            DescKeyError::DerivedKeyParsing => write!(f, "Parsing derived key"),
+        }
+    }
+}
+
+impl error::Error for DescKeyError {}
+
+/// A public key used in derived descriptors
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DerivedPublicKey {
```
edouardparis, line 29: Maybe later we can user https://docs.rs/miniscript/latest/miniscript/descriptor/struct.SinglePub.html
but for now I am not sure it is implementing enough traits like fmt::Display

darosior: Good call. Another advantage of our own `DerivedPublicKey` is that the type enforces an origin be always present.

darosior:
Added a fix for a bug found by @edouardparis.

edouardparis (approved):
ACK 9394be645c698591da9c477dd77363010cb3298e
Tested with gui on branch #385

PR #384, opened 2023-03-27
gui settings: truncate file

### LINKED ISSUE #383: Alias update to remove it failed but is shown as both succeeded and errored
image

It shouldn't be shown as "Updated" if it wasn't. We should probably have support to remove it, or have an error and be clear you cannot set an empty alias?

edouardparis:
We want to override preexisting settings file.

fixes #383

### DIFF
--- a/gui/src/app/settings.rs
+++ b/gui/src/app/settings.rs
@@ -47,6 +47,7 @@ impl Settings {
 
         let mut settings_file = OpenOptions::new()
             .write(true)
+            .truncate(true)
             .open(path)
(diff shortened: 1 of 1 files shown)


edouardparis:
I think that we can allow user to have empty aliases, as it falls back on the fingerprint of the key.
This PR fixes #383 because it fixes the settings file update by making it clear to override the previous settings and removes risks of wrong trailing characters

darosior:
image
image


ACK 76f3c1a876c99014bee0d4483d01ceda755df7da

PR #386, opened 2023-03-27
gui: filter unused events

edouardparis:
It may removes the annoying logs that occurs after a while.
```
[1669885771][iced_futures::subscription::tracker][WARN] Error sending event to subscription: TrySendError { kind: Full }
[1669885771][iced_futures::subscription::tracker][WARN] Error sending event to subscription: TrySendError { kind: Full }
[1669885771][iced_futures::subscription::tracker][WARN] Error sending event to subscription: TrySendError { kind: Full }
```
By filtering the events to use only the events listened by the application, we may not reach some buffer size limit


darosior:
Doesn't this risk hiding potential bugs in the future? Or making them harder to debug?

edouardparis:
I do not think so, the native events were listened only to get the `CloseRequestedEvent`, so they were ignored anyway.

Issue #281, opened 2023-01-11
GUI: sort Spend transactions drafts


darosior:
Similar to  #193.

If we want to sort in chronological order, we'll need to have the creation date to be stored in the DB too.

edouardparis:
I am thinking to name the new field: `stored_at`
Not `created_at` because if the user imported the psbt from another user, the time stored is not the real time of creation of the PSBT

darosior:
Let's think of other sorting strategy before to make sure which one we wanna go for?
------- Original Message -------

darosior:
Good suggestions. Let's split this between going from no sorting (current situation) to an acceptable sorting, and then improving the sorting.

So yeah i agree Edouard latest at the top seems the only reasonable small fix.

-------- Original Message --------

edouardparis:
> So yeah i agree Edouard latest at the top seems the only reasonable small fix.

So a new field `stored_at` to listspendtxs response is required for the gui to sort the list

darosior:
Yes. Can you take care of that? It shouldn't take long to add.

darosior:
We decided to push this to v2.

darosior:
Let's see if we can have this in 0.3.

PR #392, opened 2023-03-29
Store in DB the time a Spend transaction draft was last updated

darosior:
This is an alternative to #390. 

We add a new column in the `spend_transactions` SQLite table, and as such introduce a straightforward migration system. This field is in turn always set in `updatespendtx` when we store the new PSBT, and listed in `listspendtxs` entries. It is useful for instance to sort spend transactions on the GUI (#281).

### DIFF
--- a/src/database/sqlite/schema.rs
+++ b/src/database/sqlite/schema.rs
@@ -74,7 +74,8 @@ CREATE TABLE addresses (
 CREATE TABLE spend_transactions (
     id INTEGER PRIMARY KEY NOT NULL,
     psbt BLOB UNIQUE NOT NULL,
-    txid BLOB UNIQUE NOT NULL
+    txid BLOB UNIQUE NOT NULL,
+    updated_at INTEGER
 );
 ";
 
@@ -253,6 +254,7 @@ pub struct DbSpendTransaction {
     pub id: i64,
     pub psbt: Psbt,
     pub txid: bitcoin::Txid,
+    pub updated_at: Option<u32>,
 }
 
 impl TryFrom<&rusqlite::Row<'_>> for DbSpendTransaction {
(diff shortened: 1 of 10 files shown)


darosior:
I'm working on a functional test for the migration, but opened so you can start working on the GUI side of the sorting of Spend transactions drafts. This can also be done in a follow-up.

src/database/sqlite/schema.rs
```diff
@@ -74,7 +74,8 @@ CREATE TABLE addresses (
 CREATE TABLE spend_transactions (
     id INTEGER PRIMARY KEY NOT NULL,
     psbt BLOB UNIQUE NOT NULL,
-    txid BLOB UNIQUE NOT NULL
+    txid BLOB UNIQUE NOT NULL,
+    updated_at INTEGER
```
edouardparis, line 78: What do you think to make it NOT NULL, and populate the db with current time in case of preexisting spend_transactions ?

darosior: I thought about doing this, but my thinking is that setting it to the current time (or `0` for that matter) is arbitrary and as meaningless as just being explicit that it was not set and going for `NULL`.

In addition, setting `NOT NULL` here would make the migration much more involved since we wouldn't be able to use `ALTER TABLE` (we'd have to create a new table, copy the content over, delete the old one and rename the new into the old).

Finally, this is only going to affect a tiny number of users anyway. (Those that will ever have used Liana <=0.3 and still have spend drafts from this time.)

darosior:
Rebased on master to facilitate GUI integration.

edouardparis:
tested with GUI.
Satisfied with the change, we may later move things in their own `migration` module.

darosior:
ACK f262ca2d1ca864e37704ab2a35d9426f85a31c6c -- i've tested it, and Edouard too with the GUI

PR #393, opened 2023-03-29
Backward compat functional test

darosior:
Based off #392.

This introduces a backward-compatibility functional test. For now it really is targeted at testing the DB migration from 0.3 to the upcoming 0.4, but the concept is also useful more generally to automatically test backward compatibility between our releases.


darosior:
self-ACK 6cc4efd

PR #394, opened 2023-03-30
fix installer: wrong detection of hot signer usage.

edouardparis:
We did not check that the keys in the primary path were using the hot signer. In the case of a hot signing key only used for the primary path,
the hot signer was not advertised to the context
passing to each installer step and was not stored after install.


darosior:
tested ACK 789ba5066e8e2dfdd7b86a1fad359b93dbacbdf9

PR #395, opened 2023-03-30
fix installer: wrong detection of hot signer usage.

edouardparis:
We did not check that the keys in the primary path were using the hot signer. In the case of a hot signing key only used for the primary path,
the hot signer was not advertised to the context
passing to each installer step and was not stored after install.

### DIFF
--- a/gui/src/installer/step/descriptor.rs
+++ b/gui/src/installer/step/descriptor.rs
@@ -357,6 +357,9 @@ impl Step for DefineDescriptor {
                         master_fingerprint,
                         name: spending_key.name.clone(),
                     });
+                    if master_fingerprint == self.signer.fingerprint() {
+                        signer_is_used = true;
+                    }
                 }
                 let xpub = DescriptorMultiXKey {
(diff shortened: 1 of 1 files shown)


darosior:
Successfully tested i'm now asked for the mnemonic even if i pick a hot key only for the primary path. However i wonder if there isn't a less brittle way of checking if we need to store / display the mnemonics than relying on a `signer_is_used` to be set. It seems to me to be a huge footgun: we may be missing other (potentially intricate!) instances where it is set now, and we might forget to set it in the future. I'm fine with a patch like this for 0.3, but for 0.4 let's try to make this code more robust?

darosior:
Yeah i think it's best if we always use the descriptor as the source of truth. The computation should be reasonable.

edouardparis:
Can we merge this fix and think about a way to harden the installer flow after the multipath feature ? I would be easier for me

darosior:
ACK 2cbdf5bb32b711ea6057141c553b74ec810d9f0d

PR #402, opened 2023-04-03
gui: do not check empty addresses for duplicates

### LINKED ISSUE #190: GUI: don't check for duplicate addresses if there is no address
image

jp1ac4:
This is to resolve #190.

The check for a duplicate is skipped if the recipient address is an empty string.

I was able to run the GUI locally and it seemed to work as required.

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -77,7 +77,7 @@ impl ChooseRecipients {
             if !recipient.valid() {
                 self.is_valid = false;
             }
-            if !self.is_duplicate {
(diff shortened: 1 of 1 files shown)


darosior:
Cool, thanks for tackling this. Will let the call to @edouardparis.

jp1ac4:
Thank you both for your feedback.

I've now fixed the clippy warning. I can squash my commits if required (assuming all checks pass).

edouardparis (approved):
ACK f508d4a1ce9739af92b99be479dcfeeda2d263d9

PR #403, opened 2023-04-03
Add unregistered policy warning on hw list

### LINKED ISSUE #339: GUI: Label devices instead of displaying fingerprint when label is available
It's annoying when dealing with multiple similar devices to have to remember the fingerprint of which is which, when saving descriptors, signing, etc.

The issue is exacerbated by the current lack of support for multiple Ledger devices connected at the same time. I would like to know which is recognized, i might think it's the one I have in hand when it's actually the other one being "seen".

### LINKED ISSUE #369: GUI: display a warning if a signer is not registered.
Not sure what should be the wording. "Not registered"? Not very self-explanatory but i don't have anything better atm.

edouardparis:
close #369
close #339

### DIFF
--- a/gui/src/app/view/hw.rs
+++ b/gui/src/app/view/hw.rs
@@ -10,13 +10,116 @@ use liana_ui::{
 
 use crate::{app::view::message::*, hw::HardwareWallet};
 
-pub fn hw_list_view<'a>(
+pub fn hw_list_view(
     i: usize,
-    hw: &'a HardwareWallet,
+    hw: &HardwareWallet,
     chosen: bool,
     processing: bool,
-    status: Option<&'a str>,
-) -> Element<'a, Message> {
+    signed: bool,
+) -> Element<Message> {
+    let mut bttn = Button::new(
+        Row::new()
+            .push(
+                Column::new()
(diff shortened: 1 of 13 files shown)


gui/src/app/view/hw.rs
```diff
@@ -26,14 +26,32 @@ pub fn hw_list_view<'a>(
...
+                            .push_maybe(registered.and_then(|registered|
+                                    if !registered {
+                                        Some(Row::new()
+                                            .spacing(5)
+                                            .align_items(Alignment::Center)
+                                            .push(text("unregistered").small())
+                                            .push(
+                                                tooltip::Tooltip::new(
+                                                    icon::warning_icon(),
+                                                    "Policy is not registered on the device.\n You can register it in the settings.",
```
darosior, line 49:
```suggestion
                                                    "The wallet descriptor is not registered on the device.\n You can register it in the settings.",
```

We've been using "descriptor" and not policy elsewhere, let's be consistent.

edouardparis:
Self-ACK f1549e1e976a30105bc6f36488a7e90beac700e5

PR #405, opened 2023-04-04
fix not found settings file error for old gui versions.

edouardparis:
After migrating from a 0.1 or 0.2 version of the gui user may want to update the keys aliases in the settings This commit makes sure that a settings file exists when wallet is loaded.


edouardparis (commented):
Self-ACK 3eeba082483c06318ca7a1f018ce2c5da47f5f32

PR #407, opened 2023-04-05
gui: check connection to bitcoind in installer

### LINKED ISSUE #115: GUI: Installer: test the connection to bitcoind
Before moving forward we should test a bitcoind is indeed listening

edouardparis:
2023-04-05T11:31:04,960061967+02:00
2023-04-05T11:31:14,946662174+02:00
close #115

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -1757,6 +1757,7 @@ dependencies = [
  "iced",
  "iced_lazy",
  "iced_native",
+ "jsonrpc",
  "liana",
  "liana_ui",
  "log",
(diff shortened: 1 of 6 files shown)


darosior:
The screenshot looks perfect to me. Can the check be made automatically when we land on this page? So that if they have a bitcoind running on this port already (the usual case) they can just click next immediately?

edouardparis:
>  Can the check be made automatically when we land on this page? So that if they have a bitcoind running on this port already (the usual case) they can just click next immediately?

It is the case, when arriving on the page the check is done immediately with the default values.

darosior:
Awesome.

edouardparis (commented):
Self-ACK ec228485e558ff2e04052e3a6fbc81c3fc55997b

PR #415, opened 2023-04-06
fix gui jsonrpc client: wrong createspend args

edouardparis:
Wrong order.
We should find a way to test the jsonrpc client agains the liana rpc api.


edouardparis (commented):
Self-ACK 54fd0d8f50836d5d414dc43de88cfd86d3a210de

PR #419, opened 2023-04-07
gui: fix a misleading tooltip on hw connection error

### LINKED ISSUE #418: Misleading error on locked signing device
image

It's not a version issue, it's just that it's locked! If we can't know for sure we shouldn't be making random guesses that could mislead users into thinking they did something wrong while they haven't.

darosior:
The error may not be due to an unsupported version. For instance the device may just be locked. In these cases telling the user to update their device is very misleading.

Fixes #418.


edouardparis (approved):
ACK 3314b7e597a11b5c6fa653ec9c18058aff64adde
Simple and effective

PR #421, opened 2023-04-07
Add alias to hotsigner component

edouardparis:
When a hot signer is suggested to user and an alias was entered then the alias must be displayed with the "This computer" label as signer kind.


edouardparis (commented):
Self-ACK 997829bdfe30a950ca40412fc1d273c2f1ca7e43

PR #430, opened 2023-04-10
Reset the daemon and GUI versions post 0.4 branch-off

darosior:
Let's use `u32::MAX` as patch version to denote a development build.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,7 +73,7 @@ pub struct Version {
 
 impl fmt::Display for Version {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}.{}.{}", self.major, self.minor, self.patch)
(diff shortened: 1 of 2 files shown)


darosior:
What do you think? Eventually it'd be nice to have the commit somehow but it's already a first step.

edouardparis:
Maybe I would prefer something like `.9999+` but I am ok with the change if we document it

darosior:
Hmm i can make it `9999`, we won't use 10 thousands patch versions anyways. Documented where?

I'm thinking also of just changing the serialization with a `-dev` suffix.

edouardparis:
Maybe top of the CHANGELOG.md file, like `yet unreleased code is marked as .9999`

edouardparis (commented):
ACK a076a83cbac37e15e9f25c0cf5ba6a62233fea86
(maybe  a boolean somewhere to set to true or false would be easier to change each release)

PR #444, opened 2023-04-12
gui: reduce granularity of slider

### LINKED ISSUE #420: UI: Reduce granularity of slider when selecting a timelock in blocks
Currently, the granularity is 1 block, which makes it hard to get the number yout want (e.g.: 6 months) by dragging the slider.

Probably a granularity of 144 blocks would be more enjoyable (ideally, you could still leave the freedom of choosing a different number in the textbox).

jp1ac4:
This is to resolve #420.

I've added a step of 144 to the slider.


edouardparis (approved):
ACK dd84e76c82b0426daa21ebf913d9dccef65e8079
Very good, I did not know the step method

PR #455, opened 2023-04-20
Gui new coins theme

### LINKED ISSUE #441: Use time instead of number of blocks everywhere for coins expiry
The durations before a coins expires need to be in time, not blocks. Below one day we shouldn't try to be precise but rather say "in less than a day".

edouardparis:
close #441


edouardparis (commented):
Self-ACK d4c2f702abc0d72045734b824c6f69e3fcb92381

PR #470, opened 2023-04-29
gui: skip register step if no device used

### LINKED ISSUE #400: Installer: do not make them check device registration of the descriptor if there is no device
It keeps confusing people that use the wallet without a signing device.

jp1ac4:
This is to resolve #400.

Thanks to @edouardparis for the suggestions and code snippets provided that form the basis of these changes.

I'm opening this PR as draft so you can see my progress and check in case I've missed anything or you'd like some tests added.

Based on my own tests with and without devices, this is working as expected.

### DIFF
--- a/gui/src/installer/context.rs
+++ b/gui/src/installer/context.rs
@@ -31,6 +31,7 @@ pub struct Context {
     )>,
     pub data_dir: PathBuf,
     pub signer: Option<Arc<Signer>>,
+    pub hw_is_used: bool,
 }
 
 impl Context {
@@ -46,6 +47,7 @@ impl Context {
             descriptor: None,
             data_dir,
             signer: None,
+            hw_is_used: false,
         }
     }
 
--- a/gui/src/installer/message.rs
+++ b/gui/src/installer/message.rs
@@ -6,6 +6,7 @@ use std::path::PathBuf;
 
 use super::Error;
 use crate::hw::HardwareWallet;
+use async_hwi::DeviceKind;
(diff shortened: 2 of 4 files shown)


jp1ac4:
OK sure, I'll give that a try. Would it also be worth adding some assertions to the existing test?

jp1ac4:
I've made an initial attempt at a test to check that the bool `hw_is_used` is being set correctly in a few different cases. Please let me know if this is along the right lines.

I've also rebased on master.

Thanks.

edouardparis:
Tested and everything seems fine.
However when importing a descriptor while participating in a shared setup or recovering a wallet, the registration step is not shown anymore. Can you modify ImportDescriptor.apply by setting ctx.hw_is_used to true with a comment please ?
```
// Set to true in order to force the registration process to be shown to user.
ctx.hw_is_used = true;
```

jp1ac4:
Thanks for your feedback.

I've now added those lines before the `if` expression in the `apply` method, but can move it within the `if` expression if you prefer. I checked that the registration step now appears when importing a descriptor.

I've squashed my commits to a single commit and rebased on master.

edouardparis (approved):
ACK 38340c01898a8b19f84abe5e913cd0316d93a9c4

edouardparis:
Good job :clap: 
The installer has sadly a lot of moving parts, but it went great.

jp1ac4:
Thanks very much for all your help :pray:

PR #477, opened 2023-05-04
commands: allow to create send-to-self transactions with 'createspend'

darosior:


darosior:
Turns out it was straightforward. But i'm wary as these calculations are always pretty intricate.

edouardparis (approved):
ACK c6c312a01171460446ab41e22b9832edb95fa83b
tested with GUI

PR #482, opened 2023-05-07
Some wording cleanups all around the GUI

### LINKED ISSUE #446: Wording for v1
We should go through all the wording all around the GUI and update it if necessary. I'm thinking for instance about tooltips, rescan, recovery, spend creation process.

### LINKED ISSUE #460: Homepage coins expiry warning typo
image

I don't think it's correct to say "to be expired". Instead i'd say:
> 2 of your coins are or will soon be available through the recovery path

If we want to be even more correct we could detect if they've got more than 1 recovery path and if that's the case say "through the *first* recovery path", but i don't think it's really important.

cc @kloaec if you can chime in on the wording here.

darosior:
Fixes #460.
Fixes #446.

### DIFF
--- a/gui/src/app/view/home.rs
+++ b/gui/src/app/view/home.rs
@@ -57,7 +57,7 @@ pub fn home_view<'a>(
                         .push(
                             h4_regular(format!(
                                 "≈ {} left before first recovery path becomes available.",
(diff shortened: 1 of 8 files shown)


edouardparis (approved):
ACK 6bf29123670a4304127859c98e7e9bf689721789

PR #486, opened 2023-05-07
Make home warning message relative to timelock

### LINKED ISSUE #461: Home expiry warning should be relative to the configured timelock
It seems that #455 made it so that the warning will be displayed if there is less than a day worth of blocks before a recovery path is available. (@edouardparis please correct me if i'm wrong.)

But this arbitrary threshold doesn't generalize well. For instance if one set their recovery path to 1 year, we should definitely warn them sooner than a day before their coin(s) expire(s).

That's why i suggested using 10% of the configured timelock value in #439. But it can also be a smaller percentage. 1%? 5%?

5% generalizes fine: for a timelock of 1year it'll display the warning 2 weeks and a half before it becomes available. For a timelock of 3 months it'll be a bit less than 5 days before. For a maximum timelock of a year and (around) 3 months, it'll warn a bit more than 3 weeks before the recovery path becomes available.

What do you think @edouardparis @kloaec @mariabmont?

edouardparis:
close #461

### DIFF
--- a/gui/src/app/state/mod.rs
+++ b/gui/src/app/state/mod.rs
@@ -133,8 +133,9 @@ impl State for Home {
                                 let timelock = self.wallet.main_descriptor.first_timelock_value();
                                 let seq =
                                     remaining_sequence(&coin, cache.blockheight as u32, timelock);
-                                // number of block in a day
-                                if seq <= 144 {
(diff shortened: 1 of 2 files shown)


edouardparis (commented):
Self-ACK 25fffb2c52f87715a69cee3fd0f6cda64e2185f2

PR #487, opened 2023-05-08
Check hot signer usage using descriptor

### LINKED ISSUE #397: Installer: check hot signer presence and backup multiple time using created descriptor
See #395

Another way also is to refont installer flow to use less optional components in the installer context.

edouardparis:
It is a refont of the installer flow
to use less optional components in the installer context.

Hot signer is stored a the install step if its fingerprint is present in the descriptor string

close #397


edouardparis (commented):
Self-ACK 307d12d980b618a47f5a929046d0643e688cd1ff

PR #492, opened 2023-05-10
fix installer final step view

edouardparis:
The height of the content was filling too much and user can scroll down to the abyss


edouardparis (commented):
Self-ACK 200fa9e7b4497687890d91ca97add45aea88c1e2

PR #494, opened 2023-05-11
fix psbt view: wrong signer fingerprint padding

edouardparis:


edouardparis (commented):
Self-ACK 2d7f2254cac8a7d0246e52fcbebc90f470a34aff

PR #498, opened 2023-05-12
bitcoind: do not assert presence of 'warning' field in unloadwallet response

darosior:
We should really be less picky in the bitcoind interface... Too many unnecessary assertions.


darosior:
Alright so https://github.com/bitcoin/bitcoin/pull/27279 actually broke the interface. Added a commit to adapt our `warning` field detection in all wallet commands..

darosior:
self-ACK 0f07361d9b6a9d118360607479ef9f7e5d9278e1

PR #499, opened 2023-05-12
sqlite: prevent races when updating the next derivation index

darosior:
You can check the unit test triggering the race fails on master. I'm pretty sure it's the reason for the address generation bug i experienced yesterday when testing.


edouardparis (commented):
utACK 039b82d9b3e11786f09c761a6d0a6d137cdff8ad

PR #535, opened 2023-05-20
add udev dependency to TRY.md

pythcoiner:
i just found out that on ubuntu 22.04,  libudev-dev cannot be installed if udev is not installed first


darosior:
Thanks. utACK 4dcac426a941dd62b27d4df1379ef836e394a789.

PR #536, opened 2023-05-21
Make compilable on aarch64 (M1)

raphjaph:
I ran into some issues while trying to build this on an M1 Mac. First, the `bip39` crate was somehow not being imported correctly and secondly there wasn't a way to just use the `hardware_randomness()` if `cpu_randomness()` wasn't available. Not sure how to do the `#[cfg]` stuff correctly so feel free to close and do another way. Just wanted to post somewhere in case someone else runs into this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,11 +60,11 @@ base64 = "0.13"
 # Used for generating mnemonics
 getrandom = "0.2"
 
+# Used for the hot signer
+bip39 = "2.0"
+
 # Additional entropy for generating mnemonics
 [target.'cfg(target_arch = "x86")'.dependencies]
 rdrand = "0.8"
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 rdrand = "0.8"
-
-# Used for the hot signer
-bip39 = "2.0"
--- a/src/random.rs
+++ b/src/random.rs
@@ -41,11 +41,6 @@ fn cpu_randomness() -> Result<Option<[u8; 32]>, RandomnessError> {
     }
 }
 
-#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
-fn hardware_randomness() -> Result<Option<[u8; 32]>, RandomnessError> {
-    Ok(None)
-}
-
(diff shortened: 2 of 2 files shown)


src/random.rs
```diff
@@ -82,6 +82,12 @@ fn additional_data() -> Result<[u8; 32], RandomnessError> {
 pub fn random_bytes() -> Result<[u8; 32], RandomnessError> {
     let mut engine = sha256::HashEngine::default();
 
+    #[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+    if let Some(bytes) = hardware_randomness()? {
+        engine.input(&bytes);
+    }
```
darosior, line 88: This is a noop.

https://github.com/wizardsardine/liana/blob/32dbb9490e044f15fb62a9f86815aaaa5b435d80/src/random.rs#L44-L47

src/random.rs
```diff
@@ -82,6 +82,12 @@ fn additional_data() -> Result<[u8; 32], RandomnessError> {
...
+
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
darosior, line 90: Nice, i had forgotten this. I think this is actually what fixes your compilation on M1.

Cargo.toml
```diff
@@ -60,11 +60,11 @@ base64 = "0.13"
...
+bip39 = "2.0"
+
 # Additional entropy for generating mnemonics
 [target.'cfg(target_arch = "x86")'.dependencies]
 rdrand = "0.8"
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 rdrand = "0.8"
-
-# Used for the hot signer
-bip39 = "2.0"
```
darosior, line 70: :facepalm:. Thanks.

src/random.rs
raphjaph: True, then I just leave it in? I can't see it used anywhere

raphjaph:
Force pushed to make it one commit

darosior:
utACK 700f6e1d6fc9f4fcbfeeffc00a49c655be8ce671

PR #540, opened 2023-05-26
gui: clear fingerprint alias when changing device type

### LINKED ISSUE #531: GUI: Clear fingerprint value when changing device type on installer
I noticed the following inconsistent fingerprint behaviour when setting the primary path on the installer:

Case 1: Change Specter to hot signer
- Set primary path -> choose Specter -> enter alias "spec" -> Apply
- Edit primary path -> Choose hot signer -> fingerprint value is cleared and placeholder shown:
image

Case 2: Change hot signer to Specter
- Set primary path -> choose hot signer -> enter alias "hot_signer" -> Apply
- Edit primary path -> choose Specter -> "hot_signer" is still shown as value for fingerprint:
image

I was expecting Case 1 behaviour to apply also in Case 2 so that the fingerprint value would be cleared and the placeholder shown.

Possibly this can be made consistent with the following change:
```diff
diff --git a/gui/src/installer/step/descriptor.rs b/gui/src/installer/step/descriptor.rs
index 88ca8cf..58172f8 100644
--- a/gui/src/installer/step/descriptor.rs
+++ b/gui/src/installer/step/descriptor.rs
@@ -862,6 +862,7 @@ impl DescriptorEditModal for EditXpubModal {
                                 self.edit_name = false;
                             } else {
                                 self.edit_name = true;
+                                self.form_name.value = String::new();
                             }
                             self.form_xpub.valid = true;
                             self.form_xpub.value = key.to_string();
```

jp1ac4:
This is to resolve #531.

### DIFF
--- a/gui/src/installer/step/descriptor.rs
+++ b/gui/src/installer/step/descriptor.rs
@@ -862,6 +862,7 @@ impl DescriptorEditModal for EditXpubModal {
                                 self.edit_name = false;
                             } else {
                                 self.edit_name = true;
+                                self.form_name.value = String::new();
                             }
                             self.form_xpub.valid = true;
                             self.form_xpub.value = key.to_string();


edouardparis (approved):
ACK c55b992ace93da5b334a878b92a62a5d5325a0cd

Issue #517, opened 2023-05-13
Show 'level' of expiration in Send page


pythcoiner:
![image](https://github.com/wizardsardine/liana/assets/124568858/82b3aafa-fb42-4899-9e06-81f1d5d4ab8d)

It can be nice to add info about the expiring level of a coin, by example 'Expired 2/3' in the coin selection area

darosior:
You mean like for #516, display how many of the recovery paths are available?

pythcoiner:
my thought was more simple just a label: (#516 i think about something more 'graphical')

![coin_selection](https://github.com/wizardsardine/liana/assets/124568858/ddeb6273-790f-4943-bebf-00bbd791cdbc)

VeronicaTee:
I'd like to work on this issue, if it's up for grabs, that is

edouardparis:
I think the 2/3 expired label can confuse the user that thinking the expiration is not yet done. We use this kind of ratio for psbts with signatures missing for example.

If we want to display number of path available, it would require more modification of the liana library side first before integrating into the GUI.
The GUI use the helper [first_timelock_value](https://github.com/wizardsardine/liana/blob/master/src/descriptors/mod.rs#L182)
So we should certainly first change for an helper providing an ordered list of timelocks.

edouardparis:
> So, a 'timelock_value' function (for instance), that provides an ordered list of timelocks instead of a single value would suffice?

To replace first_timelock_value, as a first step in the liana lib code yes :)

darosior:
So there is two aspects to this issue. The first is to expose this information at all from the daemon side. The second is about making use of this information on the GUI side once we have it.

Regarding exposing the data i think you are on the right track @VeronicaTee. We could make `first_timelock_value` into a `timelocks_values` method that returns an ordered list of timelocks. Note internally we store the recovery paths as a `BTreeMap` for which the keys are sorted. Then you'd need to update the call sites on the GUI side where it assumes a single timelock value.
The initial reason for not making this value a list in the first place was to convey there is always at least one timelock. I didn't want the caller (here the GUI) to have to handle the case where the list is empty, as it would make an assumption already made internally by the daemon bleed into the caller code. Maybe we could have a newtype that represents a never-empty list here? It seems like a somewhat common pattern, is there some cleverer Rust trick to achieve this?

[...]

PR #547, opened 2023-05-29
Get ordered list of recovery paths

VeronicaTee:
The new function gets the ordered list, as discussed as part of #517

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -43,10 +43,10 @@ task:
     set -xe
 
     # Download the bitcoind binary
-    curl -O https://bitcoincore.org/bin/bitcoin-core-24.1/bitcoin-24.1-x86_64-linux-gnu.tar.gz
-    echo "c112af3d19cca7f5b5f942708ca4d522110d8e1dc1c0f8d3077f531c94e5f00f bitcoin-24.1-x86_64-linux-gnu.tar.gz" | sha256sum -c
-    tar -xzf bitcoin-24.1-x86_64-linux-gnu.tar.gz
-    export BITCOIND_PATH=bitcoin-24.1/bin/bitcoind
+    curl -O https://bitcoincore.org/bin/bitcoin-core-25.0/bitcoin-25.0-x86_64-linux-gnu.tar.gz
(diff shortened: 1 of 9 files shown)


darosior (commented):
Thanks for the PR. I think it's the right first step toward #517.

Regarding the code. As i mentioned in https://github.com/wizardsardine/liana/issues/517#issuecomment-1562468285 i think we should convey somehow that the vector will never be empty, if possible.
This new interface should also be tested. For instance with a unit test against a descriptor that comports multiple timelocks.

Regarding the PR, please [squash your commits](https://github.com/wizardsardine/liana/blob/master/CONTRIBUTING.md#workflow).

VeronicaTee:
Regarding conveying somehow that the vector will never be empty, I can look into a way of ensuring that.

For the new interface, this will be effected and tested as suggested, and added in another PR, once I get the go-ahead to implement this function for it.

.cirrus.yml
```diff
@@ -43,10 +43,10 @@ task:
     set -xe
 
     # Download the bitcoind binary
-    curl -O https://bitcoincore.org/bin/bitcoin-core-24.1/bitcoin-24.1-x86_64-linux-gnu.tar.gz
-    echo "c112af3d19cca7f5b5f942708ca4d522110d8e1dc1c0f8d3077f531c94e5f00f bitcoin-24.1-x86_64-linux-gnu.tar.gz" | sha256sum -c
-    tar -xzf bitcoin-24.1-x86_64-linux-gnu.tar.gz
-    export BITCOIND_PATH=bitcoin-24.1/bin/bitcoind
+    curl -O https://bitcoincore.org/bin/bitcoin-core-25.0/bitcoin-25.0-x86_64-linux-gnu.tar.gz
+    echo "33930d432593e49d58a9bff4c30078823e9af5d98594d2935862788ce8a20aec bitcoin-25.0-x86_64-linux-gnu.tar.gz" | sha256sum -c
```
darosior, line 47: Unrelated change? You must have messed up the rebase. :)

darosior (commented):
> For the new interface, this will be effected and tested as suggested, and added in another PR, once I get the go-ahead to implement this function for it.

What go-ahead? I agree it's better to do the GUI integration in another PR but this one introduces a new API that needs to be tested.

pythcoiner:
This PR is stale for ~3years, I'm closing, feel free  if you re-gain some interest in it.

Issue #519, opened 2023-05-13
Installer packages for linux


pythcoiner:
Should we build install packages for linux, for user can easily install liana (apt, dnf, yum, etc.... maybe snap?)
which one do you think is needed?

darosior:
I think we could at least do a `.deb` and an arch package. Maybe an RPM as well?

pythcoiner:
i made a try with .deb, i got something not bad...

 which task do we need to execute prior to install the package? (only remove old versions liana binaries if present? have a check if bitcoind installed on some standard location? check if bitcoind running for some kind autosettings about the node?)

during install of package i do this:
- copy binaries in /usr/bin/
- copy liana logo in opt/liana/
- build liana.desktop in /usr/share/applications/
- anything else to do?

do we need to do some cleanup when removing liana?

darosior:
I've never created a `.deb` but that looks correct. I think you need to make sure the system's got the necessary dependencies (list [here](https://github.com/wizardsardine/liana/tree/master/gui)) installed. Another thing is maybe installing it for the user instead of system-wide? Like in `~/.local/bin` instead of `/usr/local/bin` and so on? Not sure. Is there any recommendation from the Debian doc?

pythcoiner:
about the dependencies my though was they are needs at compile time, will check that.
about install the binaries into user's home folder i'll have a look on debian doc, but if i don't misunderstand, i read something that .deb installer cannot acces home folder (at least on uninstall process).

pythcoiner:
> Another thing is maybe installing it for the user instead of system-wide? Like in `~/.local/bin` instead of `/usr/local/bin` and so on? Not sure. Is there any recommendation from the Debian doc?

I take little time looking at the debian doc about .deb packages, they don't mention locations, i think it's more about linux standards, i got a check about the app installed with apt on my ubuntu, they almost have the binaries (or symlink to the binaries) in /usr/bin.

In my mind, the home folder (~/) is reserved for user storage, i dont know if its the same thought for almost linux users?

pythcoiner:
i'm thinking about to add a glibc version check prior to install package, i think it's a must?

darosior:
You should be able to specify dependencies and their versions in the `.deb`, GLIBC is just one of them.
------- Original Message -------

pythcoiner:
> You should be able to specify dependencies and their versions in the `.deb`, GLIBC is just one of them.

yes, found out that libc6 (>= 2.34) is the dependency we need

[...]

PR #550, opened 2023-05-30
Deb package

pythcoiner:
following #519 

should i locate the script in some specific location instead of at the repo root?

is it a way in the CI workflow to automate the test for .deb installs on different versions?


darosior:
Let's have it in `contrib/` for now. Eventually it should be integrated to the release script (which could be ported to Python). I don't think we should have it in CI at this point, but this should be doable.

Let me know when this is ready for review.

pythcoiner:
> Let me know when this is ready for review.

ready

pythcoiner:
> Eventually it should be integrated to the release script (which could be ported to Python)

I can handle that, it looks trivial

darosior (commented):
ACK 7c3511e1e4516bef26f525b7280aeeda9299d206 -- can be refined and integrated into the release process in a follow-up.

Thanks!

PR #565, opened 2023-07-12
Update `rust-miniscript` (and thereby `rust-bitcoin`) to latest version

darosior:
It was a big chunk, especially in making sure we don't introduce any silent bug with all the upstream recent code movements.

Also, we no longer depend on my custom `rust-miniscript` branch! :tada:

### DIFF
--- a/src/bitcoin/poller/looper.rs
+++ b/src/bitcoin/poller/looper.rs
@@ -31,16 +31,28 @@ fn update_coins(
     descs: &[descriptors::SinglePathLianaDesc],
     secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
 ) -> UpdatedCoins {
+    let network = db_conn.network();
     let curr_coins = db_conn.coins(CoinType::All);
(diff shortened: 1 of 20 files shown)


src/bitcoin/poller/looper.rs
```diff
@@ -31,16 +31,22 @@ fn update_coins(
...
     log::debug!("Current coins: {:?}", curr_coins);
 
     // Start by fetching newly received coins.
     let mut received = Vec::new();
     for utxo in bit.received_coins(previous_tip, descs) {
         // We can only really treat them if we know the derivation index that was used.
-        if let Some((derivation_index, is_change)) =
-            db_conn.derivation_index_by_address(&utxo.address)
-        {
+        let address = match utxo.address.clone().require_network(network) {
```
darosior, line 42: Unnecessary clone, could just move it out as is done below with other fields.

darosior: Done.

darosior:
Self-ACK a19f2c1536cab6f6a3e85eedcba9be4b907dc33b.

Issue #542, opened 2023-05-29
Improve UX for creating a new wallet with remote participants


pythcoiner:
When you create a wallet importing Xpub from other remotes participants, it would be nice that after we already register one xpub (by example in the primary path) we can just have to 'select' the key in the signing devices list instead of paste its Xpub (maybe need to have a color scheme for differentiate hot/hardware/remote wallets?) .

here a quick example:
![image](https://github.com/wizardsardine/liana/assets/124568858/4b3bb3d8-f5e7-46dd-8b4e-498bff14eac4)
 `liana2` is my 'local' Liana hot key
`liana3` is a remote (hot) key (i already import xpub in primary path)
`liana4` is another remote (hot) key (i already import xpub in primary path)

when i add a key to the recovery path:
![image](https://github.com/wizardsardine/liana/assets/124568858/0f4f7cf4-d3b1-44a9-93f4-567c4dbc068d)
only liana2 is present in the signing device list (i think we need to display all the already imported key here)

pythcoiner:
ok i think i got the point, because we cannot reuse the same derivation path (we cannot use twice the same xpub in the descriptor), but is it any way to derive only from a known Xpub (so that derivation can be done on the 'coordinator' side instead the remote guys need so send a batch of keys)?

kloaec:
Great point. I just want to make sure we don't do anything stupid, but it seems perfectly doable, and a big improvement on the current UX for participants, and creator.

We would need to either to derive one more depth to have the different xpubs, or to have a way to "go back up" one step in the tree, to derive siblings to the first one.

We may want to get a (new) specific path for the xpub to be extracted/shared, to make sure one step higher isn't hardened if we go the siblings route.

Let's discuss this more, I think it's a very good ux improvement and should be implemented as soon as possible.

darosior:
Yes indeed! Next time you should link it with using "Fixes #XXX" in the PR OP.

PR #575, opened 2023-07-19
descriptor: make it possible to reuse xpubs at different derivation paths

darosior:
This is unnecessary to bump a hardened step to get another xpub for the same participant:
- It's more hurdle when sharing it
- It's more hurdle when verifying it on the device's screen
- It's for the same person within the same script anyways

Fix this by allowing to use the same xpub multiple times in a descriptor as long as the derivation path is different. Then the GUI will be able to use a new unhardened derivation step instead of bumping the BIP48 "accounts". A unit test is introduced that showcases this.

See also https://github.com/wizardsardine/liana/issues/542.

### DIFF
--- a/src/descriptors/analysis.rs
+++ b/src/descriptors/analysis.rs
@@ -67,7 +67,7 @@ fn is_single_key_or_multisig(policy: &SemanticPolicy<descriptor::DescriptorPubli
 }
 
 struct DescKeyChecker {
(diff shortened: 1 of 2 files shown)


darosior:
Fucking flaky tests :sob:

edouardparis (approved):
ACK 28049d4f7cd2001207cc2b204c31b33d9640077d 
tested with GUI #581

Issue #576, opened 2023-07-19
GUI: disallow xpubs with appended derivation paths


darosior:

darosior:
Actually it's also a failure of the daemon.

PR #577, opened 2023-07-19
descriptor: refuse non-normalized xpubs

darosior:
We wouldn't be able to derive them. See https://github.com/wizardsardine/liana/issues/576#issuecomment-1642429179.


darosior:
self-ACK f4d184f6f7fb748bff17f8a846616c589a432342

PR #578, opened 2023-07-20
Fix the tracking of immature coinbase deposits

### LINKED ISSUE #567: Flaky `test_coinbase_deposit` functional test
It sometimes randomly fails on unrelated PRs in ci. Although i'm struggling to reproduce the failure locally.

darosior:
#567 uncovered that we were actually not tracking coinbase deposits correctly. In fact we would most likely miss them all in any real situation. This is because we would filter out immature coinbase deposits from the result of `listsinceblock` and not consider them newly received coins. But they would be confirmed, and as the chain moves forward we'd not scan this range anymore even once they've become mature.

This PR fixes it by the simplest possible manner: record immature coinbase deposits as unconfirmed and only mark them as confirmed once they've become mature. This is a bit clumsy, but should be fine for the number of users that would receive payouts from coinbase transactions (ie most likely 0). Also, we don't accurately update coinbase coins on reorg. This is unnecessary as it'd be very unlikely that a mature coinbase would become immature and if there is a 100 blocks reorg that would invalidate it altogether we'd have bigger problems.

Fixes #567.
Still as draft as i want to go over this one more time before asking for review, as this if pretty intricate and touches core parts of our codebase.

### DIFF
--- a/src/bitcoin/poller/looper.rs
+++ b/src/bitcoin/poller/looper.rs
@@ -66,6 +66,7 @@ fn update_coins(
             if !curr_coins.contains_key(&utxo.outpoint) {
                 let coin = Coin {
                     outpoint,
+                    is_immature: false,
                     amount,
                     derivation_index,
                     is_change,
--- a/src/bitcoin/poller/looper.rs
+++ b/src/bitcoin/poller/looper.rs
@@ -24,6 +24,8 @@ struct UpdatedCoins {
 // or spent.
 // NOTE: A coin may be updated multiple times at once. That is, a coin may be received, confirmed,
 // and spent in a single poll.
+// NOTE: Coinbase transaction deposits are very much an afterthought here. We treat them as
+// unconfirmed until the CB tx matures.
 fn update_coins(
     bit: &impl BitcoinInterface,
     db_conn: &mut Box<dyn DatabaseConnection>,
@@ -42,6 +44,7 @@ fn update_coins(
             outpoint,
             amount,
             address,
+            is_immature,
             ..
         } = utxo;
         // We can only really treat them if we know the derivation index that was used.
@@ -66,7 +69,7 @@ fn update_coins(
             if !curr_coins.contains_key(&utxo.outpoint) {
                 let coin = Coin {
                     outpoint,
-                    is_immature: false,
+                    is_immature,
                     amount,
                     derivation_index,
                     is_change,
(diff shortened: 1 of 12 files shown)


darosior:
I'd like to this in master rather earlier than later. I think i'll have another pass at reviewing tomorrow and merge it if i don't find anything.

@edouardparis please do still review it, it's pretty critical that we get this right.

darosior:
self-ACK 097d5e71c19c0c5fc53d1e91087d91edfdd218f3. Didn't have the chance to re-review it but getting it in is best at this time. Edouard will still have a look post-merge.

src/database/sqlite/utils.rs
```diff
@@ -165,6 +165,25 @@ fn migrate_v0_to_v1(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError
...
+fn migrate_v1_to_v2(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError> {
+    db_exec(conn, |tx| {
+        tx.execute(
+            "ALTER TABLE coins ADD COLUMN is_immature",
+            rusqlite::params![],
+        )?;
+        tx.execute(
+            "UPDATE coins SET is_immature = 0 WHERE is_immature IS NULL",
+            rusqlite::params![],
+        )?;
```
edouardparis, line 179: After the update, I would alter the column to make the is_immature `not null`

darosior: Wouldn't that mean we'd have to ALTER it also when creating it for the first time? The schema currently doesn't have `NOT NULL`.

edouardparis: https://github.com/wizardsardine/liana/pull/578/files#diff-9afc8c965492b82a57677899c77edda2d3c6ec9da5a31863b3550ba3d072da88R60
I may misunderstood this line then

darosior: oh, no, you're right. It doesn't match what i did for the first migration..

darosior: I guess i could even go with the one liner
```sql
ALTER TABLE coins ADD COLUMN is_immature NOT NULL DEFAULT 0
```

Or maybe even with the `CHECK (is_immature IN (0,1))`.

darosior: So i decided to not go with my approach to have exactly the same schema as if the database was freshly created (without any `DEFAULT`). But it's not possible to use your approach in SQLite: https://www3.sqlite.org/matrix/lang_altertable.html.

I'll just go with my one liner then.

src/bitcoin/poller/looper.rs
```diff
@@ -42,6 +44,7 @@ fn update_coins(
             outpoint,
             amount,
             address,
+            is_immature,
```
darosior, line 47: This may trigger the assertion and the SQL constraint. My bad. See #1001.

PR #551, opened 2023-05-31
gui: update async-hwi 0.0.7

edouardparis:

PR #585, opened 2023-07-31
1.1 backports

darosior:
Some bug fixes for 1.0 to be released before 2.0 is out.

The first commit is from #551 to fix a connection bug to the Specter DIY.


darosior:
self-ACK 81a40b84778a36544c8269a0f834de5a629b9ee1

Issue #570, opened 2023-07-13
Automatically startup and download `bitcoind` when starting up the GUI


darosior:
If `bitcoind` is not already running when the GUI is started, we should run it. If we can't find any `bitcoind` binary, we should download it from https://bitcoincore.org/bin/bitcoin-core-25.0/.

There is a number of things to be taken into account. Here is a non exhaustive list:
- What options should we start `bitcoind` with? Pruned? By how much? Lower cache?
- What warning should appear about those in the GUI (like "around 20G of space will be used")? Any other check (like checking available storage)?
- We should write the options to a `bitcoin.conf`, and not just startup `bitcoind` with those. This is to avoid any surprising behaviour to the user, and also in case we update those options.
- What if there is an existing data directory but we can't find any `bitcoind` binary?
- What if there is an existing data directory, we can find a `bitcoind` binary, but not a bitcoin.conf?
- The installer should also download and start bitcoind instead of just checking the connection.
- Where should the bitcoin.conf live? In general it's cleaner if we keep the files we write into our own data directory, but then if we don't write the bitcoin.conf within bitcoind's data directory it kills the purpose of having a bitcoin.conf in the first place. (For instance, we download bitcoind, start it up pruned, it syncs, and later the user starts bitcoind again with no knowledge of the bitcoin.conf we used and it starts IBD again because they didn't specify `-prune`.)
- Should we go for a more radical option and have the bitcoind data directory live inside our own data directory? So that it would prevent most footguns, but could be a waste of disk space if for some reason the user starts using a bitcoind with a regular data directory.
- Where should we store the `bitcoind` binary we download?

darosior:
Meta point: the development could be split in two PRs: 1) find and startup bitcoind 2) download bitcoind if we can't find it.

darosior:
Related: https://github.com/wizardsardine/liana/issues/101.

pythcoiner:
> * What warning should appear about those in the GUI (like "around 20G of space will be used")? Any other check (like checking available storage)?

do we have the ability to check if the space available is on SSD rather than HDD? (I guess users will not wait several weeks for IBD)

pythcoiner:
> * What if there is an existing data directory but we can't find any `bitcoind` binary?

At least ask the user to confirm that we can use data from this directory?

pythcoiner:
> * What if there is an existing data directory, we can find a `bitcoind` binary, but not a bitcoin.conf?

[...]

PR #586, opened 2023-07-31
bitcoind: create watchonly wallet with `load_on_startup = true`

darosior:
Upon startup bitcoind will update loaded wallets before pruning historical chain data.

This is a trivial fix for people using an aggressively pruned `bitcoind` to not be forced to `-reindex` if they haven't started their Liana wallet for longer than their prune target allows for.

See https://github.com/wizardsardine/liana/issues/570#issuecomment-1658108967 for a discussion about this.


darosior:
self-ACK 8ae597fec11c2fdd7873c9bd011a75a5fbbd9301

PR #591, opened 2023-08-05
add script for build AppImage

pythcoiner:
I built AppImage at commit 9f0fe9b from Ubuntu 22.10 and tested successfully against (all were fresh installs on Proxmox VM):

- Ubuntu 23.04
- Ubuntu 22.10
- Ubuntu 22.04
- Last [Archbang](https://archbang.org/) Live ISO ( I had trouble installing Arch on my Proxmox VM)
- Debian 12
- Fedora 38

Note: Running AppImages on Ubuntu needs [FUSE](https://github.com/AppImage/AppImageKit/wiki/FUSE), can be install with:

Ubuntu (>= 22.04)
```shell
sudo add-apt-repository universe
sudo apt install libfuse2
```

Tested against Debian 11.7 but got the `GLIBC` version issue.

### DIFF
--- a/contrib/BUILD_APPIMAGE.md
+++ b/contrib/BUILD_APPIMAGE.md
@@ -1,0 +1,21 @@
+## Tooling
+For build an AppImage package appimagetool is needed, you can download it at 
+[https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage](https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage)
+
+Make it executable:
+```shell
+chmod a+x appimagetool-x86_64.AppImage
+```
+
+Store it in /usr/local/bin:
+```shell
+sudo mv appimagetool-x86_64.AppImage /usr/local/bin/appimagetool
+```
+
+## Build
+
+Just run:
+```shell
+ARCH=x86_64 appimagetool Liana.AppDir
+```
+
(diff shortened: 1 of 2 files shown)


darosior:
> Note: Running AppImages on Ubuntu needs [FUSE](https://github.com/AppImage/AppImageKit/wiki/FUSE)

Hmmm bummer. The whole point of using an `AppImage` was that it was a simple click away, if it has dependencies too we might as well use the `.deb`..

pythcoiner:
> > Note: Running AppImages on Ubuntu needs [FUSE](https://github.com/AppImage/AppImageKit/wiki/FUSE)
> 
> Hmmm bummer. The whole point of using an `AppImage` was that it was a simple click away, if it has dependencies too we might as well use the `.deb`..

I need to check, but i guess my Ubuntu has been installed using the 'minimal' option, so some dependencies can be missing.
What about keeping both .deb(Debian/Ubuntu) and AppImage(non-Debian)?

pythcoiner:
> * Last [Archbang](https://archbang.org/) Live ISO ( I had trouble installing Arch on my Proxmox VM)

Just had a fresh arch installed on one of my laptops, AppImage will not run until FUSE installed also:

```shell
sudo pacman -S fuse
```

darosior:
I guess what makes the most sense is to simply distribute the binaries (instead of an AppImage) along with the packages. Thanks for looking into this, even if it doesn't end up getting merged it will have been a useful exploration.

pythcoiner:
just try a fresh Ubuntu22.04 with a normal (not minimal) install, still need to install `fuse` by hand.....

PR #592, opened 2023-08-05
gui: configure and start bitcoind

jp1ac4:
This is to resolve part of #570 (configure and start an already-installed bitcoind).

I'm opening this draft PR so that you can provide feedback and check if any changes in approach are required.

I've added optional steps to the installer for the user to configure and start an "internal" bitcoind that uses `~/.liana/bitcoind_datadir` as its data directory.

The main things missing are:
- [x] Make it work on Windows.
- [x] Stop internal bitcoind when Liana is not running.
- [x] Start this internal bitcoind, if applicable, when returning to Liana after installation. One option for this would be to check in the Liana config file whether the bitcoind `.cookie` file is within the `~/.liana/bitcoind_datadir` folder, which would indicate that the internal bitcoind should be used, and then to start it (we might need to store the executable path if it's not in PATH).
- [x] Tests!


pythcoiner:
> 2\. Stop internal bitcoind when Liana is not running.

it's only in the case Liana had started bitcoins? or even if bitcoind already running on Liana's startup?

pythcoiner:
Here I think it's important that pruning is opt-in, maybe have a (default uncheck) checkbox?
image

jp1ac4:
> it's only in the case Liana had started bitcoind? or even if bitcoind already running on Liana's startup?

Hmm, I think it's still to be decided. In general, this internal bitcoind will be managed by Liana and so probably only running when Liana is open. But we could add options/prompts where the user can choose to leave it running.

> Here I think it's important that pruning is opt-in, maybe have a (default uncheck) checkbox?

I think setting `0` will disable pruning, but then I would need to change the validation to allow that. I expect there will be other additions/changes required to this page as part of this PR, and this could be one of them. It also depends how much of this configuration we want to hide from the user, as this internal bitcoind is currently aimed at less technical users.

darosior:
> it's only in the case Liana had started bitcoind? or even if bitcoind already running on Liana's startup?

As discussed in #570 i believe an automatically managed `bitcoind` is only ever useful for those users that don't already have a `bitcoind` running on this machine. For them i believe it makes sense for us to take care of the `bitcoind` process from cradle to grave. On the other hand, if the user already has an external-to-Liana `bitcoind` installation we shouldn't mess with it at all. For those users a simple "yo start your bitcoind" error would be enough anyways.

Therefore i don't think we should ever stop an externally started `bitcoind`. Regarding our own internally started `bitcoind`, i'm on the fence. Let's consider it:
- Pros
  - Any Liana-related process is contained to Liana's lifetime. It could confuse the user for instance if bitcoind continues taking a significant share of CPU although Liana is closed.
- Cons
  - During IBD, we probably want to let bitcoind sync even though they close Liana.
  - Outside of IBD, starting bitcoind at Liana's startup always incurs a delay to let bitcoind catch up to the current tip.
  - It's useful to let bitcoind run in the background so it can provide us with fee estimates.

> It also depends how much of this configuration we want to hide from the user, as this internal bitcoind is currently aimed at less technical users.

At this point we should consider a user that wants to fiddle with bitcoind's configuration options as technical enough and/or willing to run their own bitcoind. I don't think we should expose any configuration of our internally managed bitcoind.

gui/src/installer/step/mod.rs
```diff
@@ -229,6 +638,71 @@ impl From<DefineBitcoind> for Box<dyn Step> {
...
+                message::StartInternalBitcoindMsg::ExePathEdited(exe_path) => {
+                    self.exe_path.value = exe_path;
+                    self.exe_path.valid = PathBuf::from(self.exe_path.value.clone()).exists();
+                }
+                message::StartInternalBitcoindMsg::Start => {
+                    let args = vec![
+                        format!("-chain={}", self.network.to_core_arg()),
+                        format!("-conf={}", self.conf_path.to_string_lossy()),
+                    ];
+                    // TODO: How to run in Windows as daemon-like process?
```
darosior, line 677: I don't think you can. You'd have to run it in its own thread. Pretty much how the GUI is running the Liana daemon right now if i'm not mistaken @edouardparis?

gui/src/installer/step/mod.rs
```diff
@@ -229,6 +638,71 @@ impl From<DefineBitcoind> for Box<dyn Step> {
...
+        Command::none()
+    }
+
+    fn apply(&mut self, _ctx: &mut Context) -> bool {
+        if let Some(res) = self.started.as_ref() {
+            // This only tells us the command was executed successfully.
+            // Connection is checked in the following `DefineBitcoind` step.
+            return res.is_ok();
+        }
+        false
```
darosior, line 694: By using `-daemonwait` i think we could get rid of the "set up connection to the Bitcoin node" step that follows:
image

It would be a good thing because it's either the user is managing their bitcoind and they need to give us the details to be able to connect, or we do and they shouldn't be bothered (they wouldn't even know the details).

gui/src/installer/view.rs
```diff
@@ -884,6 +886,236 @@ pub fn define_bitcoin<'a>(
...
+            form::Form::new("/usr/local/bin/bitcoind", exe_path, |msg| {
+                Message::StartInternalBitcoind(message::StartInternalBitcoindMsg::ExePathEdited(
+                    msg,
+                ))
+            })
+            .warning("Please enter correct executable path")
+            .size(20)
+            .padding(10),
+        )
+        .spacing(10);
```
darosior, line 1051: I don't think we should ask them about the executable path. We should have a set of paths to be looking at (as you do above using `which`) and fail if there is none. Soon we'll have the option of downloading Bitcoin Core if the executable cannot be found.

gui/src/installer/step/mod.rs
```diff
@@ -229,6 +638,71 @@ impl From<DefineBitcoind> for Box<dyn Step> {
...
+impl From<StartInternalBitcoindStep> for Box<dyn Step> {
+    fn from(s: StartInternalBitcoindStep) -> Box<dyn Step> {
+        Box::new(s)
+    }
+}
+
+impl StartInternalBitcoindStep {
+    pub fn new(liana_datadir: &PathBuf) -> Self {
+        Self {
+            conf_path: internal_bitcoind_config_path(liana_datadir),
```
darosior, line 650: Somehow my Liana datadir gets catenated to my Bitcoin datadir
image

darosior (commented):
This is looking great. I also had @edouardparis skim through the installer code and he told me it's the right approach (kudos apparently the installer is a bit intricate).

I've left comment on specific things inline, but as a general note i think you are asking too much from the user. I think we should really focus on less technical users, where sane defaults and less configuration options is better.

- I think i would go as far as automating the `bitcoind` startup if we could locate a file. But we can always do that in the future and it's reasonable to let them choose here. It also nicely works around the guesses we'd have to make otherwise (do they have an existing bitcoin datadir already? etc..).
image

- I think we should get rid of this entirely. I don't expect the target audience for this feature to be able to reasonably choose their P2P port, RPC port or prune target. Let's go with sane defaults (P2P and RPC ports as anything available, prune at 15G) and if they ever want anything else they can always manage their own bitcoind. 
image

pythcoiner:
> For those users a simple "yo start your bitcoind" error would be enough anyways.

+1

gui/src/installer/step/mod.rs
edouardparis: GUI is running the daemon in a std::thread by calling directly the rust function.

jp1ac4:
Thanks for all the comments so far. I'm currently working through them and will update here when I'm done.

gui/src/installer/step/mod.rs
```diff
@@ -78,24 +269,26 @@ fn bitcoind_default_cookie_path(network: &Network) -> Option<String> {
...
+
+fn bitcoind_network_dir(network: &Network) -> String {
+    match network {
+        Network::Bitcoin => "",
+        Network::Testnet => "testnet3",
+        Network::Regtest => "regtest",
+        Network::Signet => "signet",
+        _ => "",
+    }
+    .to_string()
```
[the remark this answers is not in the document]
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/commit/0cc9d91a959d222f5fb33dbb74caae71dde74cf5.

gui/src/installer/step/mod.rs
jp1ac4: I've changed this slightly in https://github.com/wizardsardine/liana/commit/87509e9b4beeecbc49b618c72b3940c284ff345c so that it creates a child process and doesn't wait for the output. I could make some Unix-specific behaviour, but for now it's the same in both cases.

In order to check it's running fine, perhaps it would be worth proceeding to the `DefineBitcoind` step to check the connection, but with the cookie path and address forms hidden and only the "Check connection" button visible. Or I could try to reuse the `ping()` function in this same step.

gui/src/installer/step/mod.rs
jp1ac4: I've changed to `daemonwait` in https://github.com/wizardsardine/liana/commit/2a28dc72e063f867f6eba70dbb30b6265bccd7e8, 

If I'm no longer waiting for the output when starting bitcoind (as per https://github.com/wizardsardine/liana/pull/592#discussion_r1288317633), is using `daemon`/`daemonwait` now redundant?

gui/src/installer/view.rs
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/commit/09b6c27eabd428e194a97f1e5c2a48c5f543cef6.

gui/src/installer/step/mod.rs
jp1ac4: Hmm, I haven't encountered this error, but will try to see if I'm mixing directories somewhere.

jp1ac4:
I've now pushed several commits addressing the main comments from above. I haven't squashed any commits yet and will wait until later for that.

> Therefore i don't think we should ever stop an externally started `bitcoind`. Regarding our own internally started `bitcoind`, i'm on the fence.

Perhaps for an internally started `bitcoind`, there could be a prompt for the user when closing Liana that asks if they'd like to leave `bitcoind` running in the background and explains some benefits of doing so.

darosior:
Testing this i get an error (which seems to be an instance of https://github.com/wizardsardine/liana/pull/592#discussion_r1285619016):
> Error: Error reading configuration file: specified config file "/home/darosior/.bitcoin/aaa/bitcoind_datadir/bitcoin.conf" could not be opened.

But the GUI tells me everything's fine:
image

gui/src/installer/step/mod.rs
darosior: > I could make some Unix-specific behaviour, but for now it's the same in both cases.

In my opinion we should have bitcoind keep running as a daemon on Unix systems. But let's keep this for a follow-up. We can release Liana v2 with the bitcoind running in a thread that is stopped when Liana is stopped, get feedback and discuss in an issue letting it run as a daemon on Unix, and implement it for v3.

> [..] Checking the connection to bitcoind

We already have implemented those checks for the daemon. It could make sense to reuse the `BitcoinD` structure from there which performs sanity checks when instantiated?
https://github.com/wizardsardine/liana/blob/b205a25ed9eddcf4fea19efc790c27fb816dca02/src/bitcoin/d/mod.rs#L225-L230

I think this would require making the `bitcoin/d` module from the Liana daemon public, which is fine imo.

gui/src/installer/step/mod.rs
darosior: Yes. As suggested in https://github.com/wizardsardine/liana/pull/592#discussion_r1288536749 let's stick to running-a-bitcoind-in-foreground-in-a-thread for now. We can make it a daemon in a future release.

gui/src/installer/view.rs
```diff
@@ -884,6 +885,131 @@ pub fn define_bitcoin<'a>(
...
+                            .align_items(Alignment::Center),
+                    )
+                    .padding(20),
+                )
+                .push(
+                    Container::new(
+                        Column::new()
+                            .spacing(20)
+                            .align_items(Alignment::Center)
+                            .push(p1_regular("Let Liana manage my node").style(color::GREY_3))
```
darosior, line 921: nit: Could this text directly be in the button below instead of naming the button "Select"?

gui/src/installer/view.rs
```diff
@@ -884,6 +885,131 @@ pub fn define_bitcoin<'a>(
...
+                            Row::new()
+                                .spacing(10)
+                                .align_items(Alignment::Center)
+                                .push(icon::circle_cross_icon().style(color::RED))
+                                .push(
+                                    text(res.as_ref().err().unwrap().to_string()).style(color::RED),
+                                ),
+                        )
+                    }
+                })
```
darosior, line 1005: It would be nice if the result of this appeared right after "bitcoind already installed", not after "Next".

image

darosior (commented):
Cool, i tried it again and the flow is much smoother now! However we should also start `bitcoind` on startup, not only on first installation.

A couple more comments below.

gui/src/installer/view.rs
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/commit/b4451e8f6332507117963ad9628ddc9a28bfee5b.

gui/src/installer/view.rs
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/commit/b152f9f7f147e13bfe43e3aeb4f714752d23d417.

gui/src/installer/view.rs
```diff
@@ -884,6 +885,131 @@ pub fn define_bitcoin<'a>(
...
+        true,
+    )
+}
+
+pub fn start_internal_bitcoind<'a>(
+    progress: (usize, usize),
+    exe_path: Option<&PathBuf>,
+    started: Option<&Result<std::process::Child, std::io::Error>>,
+) -> Element<'a, Message> {
+    let start_button = button::primary(None, "Start bitcoind").width(Length::Fixed(200.0));
```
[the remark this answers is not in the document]
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/commit/830d6f482ef08e7bf110228786f20ace14f505dd.

jp1ac4:
> Running it with the default datadir works fine. I think we should simply make the given datadir path absolute.

I've updated to use the absolute path in https://github.com/wizardsardine/liana/commit/6ef5baef61a6d203977f8dafe497175d5232521d. I also needed to do the same for the cookie path in https://github.com/wizardsardine/liana/commit/a7ff519f8f9edb0036172e703402acff17179722 so that it can connect when restarting Liana.

> However we should also start `bitcoind` on startup, not only on first installation.

I'll hopefully get this working tomorrow, as well as stopping `bitcoind` when closing Liana.

gui/src/installer/step/mod.rs
jp1ac4: I've updated to use `BitcoinD` to check the connection in https://github.com/wizardsardine/liana/commit/7ced7475cd93ec5dc0ca4c2dcd880f33812fa6f2.

This is using my Liana branch https://github.com/jp1ac4/liana/tree/make-bitcoind-public, which is branched from the commit used by the GUI (https://github.com/wizardsardine/liana/blob/master/gui/Cargo.lock#L1922).

jp1ac4:
I've made a change in https://github.com/wizardsardine/liana/commit/2f31a5c00a63c6ee9ea920506d24ba05e2803ed2 to start/stop the internal `bitcoind` when opening/closing Liana. I used the GUI config file to determine whether an internal `bitcoind` is being used.

gui/src/app/mod.rs
```diff
@@ -119,6 +119,18 @@ impl App {
...
             }
         }
+        if let Some(exe_config) = &self.config.internal_bitcoind_exe_config {
+            let args = vec![
+                format!("-chain={}", &self.cache.network.to_core_arg()),
+                format!("-datadir={}", exe_config.data_dir.to_string_lossy()),
+                "stop".to_string(),
+            ];
+            info!("Stopping internal bitcoind");
+            let _res = std::process::Command::new("bitcoin-cli")
```
jp1ac4, line 129: I'm using `bitcoin-cli` here, but perhaps I should determine `bitcoin-cli`'s path based on `bitcoind`'s path, which is saved in the config file, or otherwise kill the `bitcoind` process directly.

gui/src/launcher.rs
```diff
@@ -177,6 +178,19 @@ async fn check_network_datadir(mut path: PathBuf, network: Network) -> Result<Ne
...
 
+    if let Some(exe_config) = cfg.internal_bitcoind_exe_config {
+        let args = vec![
+            format!("-chain={}", &network.to_core_arg()),
+            format!("-datadir={}", exe_config.data_dir.to_string_lossy()),
+        ];
+        info!("Starting internal bitcoind");
+        let _ = std::process::Command::new(exe_config.exe_path)
+            .args(&args)
+            .stdout(std::process::Stdio::null())
```
darosior, line 189: Note to (self and other) reviewers: it's fine to discard stdout, we still get `bitcoind`'s logs in the `debug.log`.

gui/src/launcher.rs
```diff
@@ -177,6 +178,19 @@ async fn check_network_datadir(mut path: PathBuf, network: Network) -> Result<Ne
...
+            .spawn();
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
darosior, line 191: This is brittle. If this is to wait for `bitcoind` to start up you can just use `BitcoinD::new()` which will do just that. In addition this allows you to detect a `bitcoind` startup error in which case you can print the `stderr`?

darosior (commented):
This is starting to look really good. The installer flow is great. I need to test this more in depth now, but leaving you a couple preliminary comments.

gui/src/app/mod.rs
darosior: > Better yet, since you start the process in a thread i presume you have some sort of process handle? Can't you just stop the started process from this?

I've looked into this and sending a SIGTERM isn't portable to Windows, and other portable ways of doing this require new dependencies. Simply calling `stop` on `BitcoinD` make more sense.

gui/src/launcher.rs
jp1ac4: Shall I add this as a comment in the code?

gui/src/launcher.rs
jp1ac4: Without the `sleep`, I seem to get an error that the cookie file is missing. This also happens during the installer.

gui/src/launcher.rs
darosior: Sure, good idea.

gui/src/launcher.rs
darosior: Good to know, then this is racy. A more robust way of doing this would be (in pseudocode):
```rust
while !cookie_file.exists() {
     sleep(1s);
}
```

One way it could be achieved without having to compute the cookie file path yourself would be to retry on `BitcoinD` error about the cookie file. Note however that we need to limit the number of retries, to avoid an infinite loop if the bitcoind process dies for whatever reason.

gui/src/app/mod.rs
jp1ac4: OK great thanks. My initial impression from a quick check was that keeping track of the child process would require bigger changes so I was tending towards the `stop()` approach.

I've added the `stop()` method in a separate branch:
https://github.com/wizardsardine/liana/compare/master...jp1ac4:liana:make-bitcoind-public

From some initial tests, it seems to work nicely and I'll commit those changes shortly.

gui/src/launcher.rs
darosior: Interesting edge case that came up when testing: if the error is that `bitcoind` is already running:
> Error: Cannot obtain a lock on data directory /home/darosior/projects/wizardsardine/liana/gui/aa/bitcoind_datadir. Bitcoin Core is probably already running.

We should continue, otherwise the user would be in an infinite loop. I don't think there is a specific return code for this error, so we would unfortunately have to match on the error string.. Something like `e.to_string().contains("is probably already running.")`.

darosior (commented):
Some more comments. I think we should be close to out of draft?

I'm starting a Signet sync as if i had a fresh wallet and no bitcoind. Looking good so far, it started bitcoind on a cusom port and started syncing.
image

gui/src/launcher.rs
jp1ac4: As part of https://github.com/wizardsardine/liana/pull/592/commits/6072da8837c4d8daac1c612efb41cbe9da671d22, I introduced a `start_internal_bitcoind()` function that includes a comment about `debug.log`.

gui/src/installer/step/mod.rs
```diff
@@ -64,7 +69,199 @@ pub struct DefineBitcoind {
...
+pub struct InternalBitcoindNetworkConfig {
+    rpc_port: u16,
+    p2p_port: u16,
+    prune: u32,
+}
+
+#[derive(Debug, Clone)]
+pub struct InternalBitcoindConfig {
+    #[cfg(unix)]
+    daemonwait: bool, // Used instead of `daemon` to ensure RPC server is running before exiting.
```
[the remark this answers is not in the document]
jp1ac4: Ah yes, that was a leftover from the original approach.

For now, I've removed it in https://github.com/wizardsardine/liana/pull/592/commits/62eca0567f3b03728cd580f31fb594c3b1b1fae7 and also added some other checks when reading the config file.

gui/src/installer/step/mod.rs
```diff
@@ -111,6 +314,120 @@ fn bitcoind_default_address(network: &Network) -> String {
...
+                    InternalBitcoindConfig::new()
+                };
+                // Insert entry for network if not present.
+                if conf.networks.get(&ctx.bitcoin_config.network).is_none() {
+                    let network_conf = match (get_available_port(), get_available_port()) {
+                        (Ok(rpc_port), Ok(p2p_port)) => {
+                            // In case ports are the same, user will need to click button again for another attempt.
+                            if rpc_port == p2p_port {
+                                return false;
+                            }
```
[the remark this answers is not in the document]
jp1ac4: Ah OK, I was thinking that after the function is called once, the port would still be available the next time the function is called, although even then it would probably be unlikely to return the same value. I've left it in for now just in case.

gui/src/installer/step/mod.rs
```diff
@@ -111,6 +314,120 @@ fn bitcoind_default_address(network: &Network) -> String {
...
+                            InternalBitcoindNetworkConfig {
+                                rpc_port,
+                                p2p_port,
+                                prune: PRUNE_DEFAULT,
+                            }
+                        }
+                        _ => {
+                            return false;
```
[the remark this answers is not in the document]
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/pull/592/commits/934f38ec972293d162a9b7c0158844cbdc02e1a5 so that any errors encountered are displayed to the user, although I'm not actually logging them yet.

gui/src/installer/step/mod.rs
```diff
@@ -229,6 +550,148 @@ impl From<DefineBitcoind> for Box<dyn Step> {
...
+                            self.started =
+                                Some(Err(StartInternalBitcoindError::CommandError(e.to_string())));
+                            return Command::none();
+                        }
+                        // Need to wait for cookie file to appear.
+                        std::thread::sleep(std::time::Duration::from_secs(1));
+                        let cookie_path = match internal_bitcoind_cookie_path(
+                            &self.bitcoind_datadir,
+                            &self.network,
+                        )
```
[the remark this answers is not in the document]
jp1ac4: This has now been updated (the commits are given in https://github.com/wizardsardine/liana/pull/592#discussion_r1293606783).

gui/src/installer/step/mod.rs
```diff
@@ -229,6 +550,148 @@ impl From<DefineBitcoind> for Box<dyn Step> {
...
+        if self.started.is_none() {
+            return Command::perform(async {}, |_| {
+                Message::StartInternalBitcoind(message::StartInternalBitcoindMsg::Start)
+            });
+        }
+        Command::none()
+    }
+
+    fn apply(&mut self, ctx: &mut Context) -> bool {
+        if let Some(Ok(_)) = self.started {
```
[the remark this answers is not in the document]
jp1ac4: As part of https://github.com/wizardsardine/liana/pull/592/commits/934f38ec972293d162a9b7c0158844cbdc02e1a5, I've added a comment here to say that any errors should have been handled as part of `message::StartInternalBitcoindMsg::Start` and so I think we should be fine here to only check for the `Ok` case.

jp1ac4:
I've made some updates to address comments.

Additionally, in https://github.com/wizardsardine/liana/pull/592/commits/b27d40fe272c12e53df94a9f2d5969fa01f8f847, I now stop the internal `bitcoind` if the installer is closed before completion. It's still possible to start an internal `bitcoind` and then go back and choose to use a self-managed node, in which case the internal `bitcoind` would be left running. I think I can added something for that case too.

jp1ac4:
I've added https://github.com/wizardsardine/liana/pull/592/commits/a0e3f7b60e0f7387c7b4a418edc95b089ee66a5d that now stops the internal `bitcoind` in case the user goes back and selects to use their own node. The main issue had been that the context was not updated when the user clicked "Previous", but I've added an "Apply" command when starting the internal `bitcoind` so that the context is updated without the user clicking "Next" as is usually required.

gui/src/installer/step/mod.rs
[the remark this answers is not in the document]
darosior: Showing them is good enough, it's just a matter of not silencing them.

darosior (commented):
I think this is ready to be undrafted and commits to be squashed? Then i'll ask @edouardparis to have a look.

Did you plan on adding anything else here? Maybe unit tests at least for the configuration file?

gui/src/app/mod.rs
```diff
@@ -119,6 +121,24 @@ impl App {
                 info!("Internal daemon stopped");
             }
         }
+        if self.config.internal_bitcoind_exe_config.is_some() {
+            if let Some(daemon_config) = self.daemon.config() {
+                if let Some(bitcoind_config) = &daemon_config.bitcoind_config {
+                    match liana::BitcoinD::new(
+                        bitcoind_config,
+                        "internal_bitcoind_stop".to_string(),
```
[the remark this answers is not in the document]
jp1ac4: Ah yes, I wasn't sure what was best to put there.

jp1ac4:
> Did you plan on adding anything else here? Maybe unit tests at least for the configuration file?

Yep, I think those unit tests at least should be added. I should probably also add some docstrings on new functions/structs. Once that's done, I can squash commits. Would it be OK as a single commit?

I also have this other branch with the changes to `BitcoinD` (https://github.com/jp1ac4/liana/tree/make-bitcoind-public) that will also need a PR.

darosior:
> Would it be OK as a single commit?

Yes, go for the simplest. In the future i find it easier to review a PR broken down by commits and it also makes the git history cleaner. (FYI see [this](https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention).) But no need to do this here.

> I also have this other branch

Yes this needs to be PR'd first.

jp1ac4:
> Yes, go for the simplest. In the future i find it easier to review a PR broken down by commits and it also makes the git history cleaner. (FYI see [this](https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention).) But no need to do this here.

OK thanks, I'll do a single commit on this occasion and will try to do more atomic commits in future.

> Yes this needs to be PR'd first.

Sure, will do. Should the PR be based on master or one of the other branches? As far as I can tell, the GUI gives some errors when using Liana from master, unless I'm doing something wrong.

pythcoiner:
Just one thought,( I do not yet take the time to look at the code) what append if a bitcoind instance running on mainnet but Liana started as testnet? do you check wich network is bitcoind running?

jp1ac4:
> Just one thought,( I do not yet take the time to look at the code) what append if a bitcoind instance running on mainnet but Liana started as testnet? do you check wich network is bitcoind running?

Yep, the check is based on the network. Once the user selects a network, Liana will check whether a managed bitcoind should be used for that network, and if so, it will start/stop it while Liana is running. Any other bitcoind instances that are already running shouldn't be affected.

gui/src/launcher.rs
```diff
@@ -206,6 +207,20 @@ async fn check_network_datadir(mut path: PathBuf, network: Network) -> Result<Ne
             )
         }
     })?;
+        if let Some(exe_config) = cfg.internal_bitcoind_exe_config {
+            if let Some(bitcoind_config) = daemon_config.bitcoind_config {
+                info!("Starting internal bitcoind");
+                if let Err(e) = app::start_internal_bitcoind(&network, exe_config) {
+                    return Err(format!("Failed to start internal bitcoind: {}", e));
+                }
+                utils::poll_for_file(&bitcoind_config.cookie_path, 200, 15);
```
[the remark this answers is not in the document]
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/pull/592/commits/5053384edafbf217df4ac415c73b9d5d26cfe28b.

gui/src/installer/step/mod.rs
```diff
@@ -229,6 +578,148 @@ impl From<DefineBitcoind> for Box<dyn Step> {
...
+                        };
+                        if let Err(e) = start_internal_bitcoind(&self.network, exe_config) {
+                            self.started =
+                                Some(Err(StartInternalBitcoindError::CommandError(e.to_string())));
+                            return Command::none();
+                        }
+                        // Need to wait for cookie file to appear.
+                        let cookie_path =
+                            internal_bitcoind_cookie_path(&self.bitcoind_datadir, &self.network);
+                        poll_for_file(&cookie_path, 200, 15);
```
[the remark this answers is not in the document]
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/pull/592/commits/5053384edafbf217df4ac415c73b9d5d26cfe28b.

gui/src/app/mod.rs
```diff
@@ -119,6 +121,24 @@ impl App {
...
+                    ) {
+                        Ok(bitcoind) => {
+                            info!("Stopping internal bitcoind...");
+                            bitcoind.stop();
+                        }
+                        _ => {
+                            warn!("Could not create interface to internal bitcoind");
```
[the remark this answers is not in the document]
jp1ac4: I've updated this in https://github.com/wizardsardine/liana/pull/592/commits/00e619073bfaad59d34c92f1f9c209f500d57f17. I also extracted the common code to a function.

darosior:
Hmm you're right. I was planning on merging it today anyways, testing it currently.

darosior:
#581 was merged. It created some conflicts but they appear to be trivial.

kloaec:
Hopping in, about bitcoind

There are 2 "options":
- only run bitcoind when Liana is open (best in term of resources, and data/bandwidth management)
- run bitcoind at startup (best to avoid sync when starting Liana. Could possibly enable notifications too.)

Keeping bitcoind alive after closing Liana makes no sense: either run it all the time so no wait time at startup of Liana, or only run it during Liana use. Starting it with Liana and keeping it alive after closing Liana doesn't bring anything valuable.

I suppose the default should be to only run bitcoind with Liana. If IBD, then when trying to close Liana just say "closing Liana will interrupt the Bitcoin network synchronization. Are you sure?" yes, close / no, keep synching

For users who may complain about initial sync, we could offer an option in Settings to make the bitcoind process start at OS startup. It would need to make the compromise clear: this will possibly affect data (for those without unlimited internet) and ram.


Other comment:
- when closing bitcoind, it requires a bit of time to close. Maybe we should make this obvious in Liana? "shutting down bitcoin core..." or something.
- are there any issue closing bitcoind "too fast" after broadcasting a tx? Would be nice to check if it's actually properly broadcast.

kloaec:
Also @darosior , any reason why 15000 now that we load at bitcoind startup? Maybe something lighter would be good. 2GB is a week.

darosior:
> Keeping bitcoind alive after closing Liana makes no sense

Absolute statements make no sense.

> Starting it with Liana and keeping it alive after closing Liana doesn't bring anything valuable.

It does if you consider people would use the wallet multiple times between restarts. In fact, most people don't shutdown their laptops (the other day a Mac user told me it was 2 years and a half since he last had shutdown his laptop). Considering this, letting `bitcoind` run in the background after closing Liana makes you pay the cost of starting `bitcoind` only once for the N times you'll start Liana again before you shut down your computer.

> I suppose the default should be to only run bitcoind with Liana.

This is what this PR implements. As i mentioned in https://github.com/wizardsardine/liana/pull/592#discussion_r1288536749 it would be preferable to have these high-level discussions in their own issue about the pros and cons of letting `bitcoind` run in the background on Unix systems in a future version of Liana.

> we could offer an option in Settings to make the bitcoind process start at OS startup

This sounds pretty complicated to do in a portable manner, and a bit outside the scope of what a wallet software should be doing.

> [..] data usage

I think considerations about data usage is completely out of scope for this project: we do IBD and chain synchronization anyways, any other bandwidth usage is dwarfed compared to that. Basically you shouldn't be using Liana (or bitcoind for that matter) on a metered connection.

> when closing bitcoind, it requires a bit of time to close. Maybe we should make this obvious in Liana? "shutting down bitcoin core..." or something.

Have you experienced this in testing? The `stop` RPC command used in this PR returns immediately, so it should not make Liana hang at all.

> are there any issue closing bitcoind "too fast" after broadcasting a tx? Would be nice to check if it's actually properly broadcast.

The RPC command we use on `bitcoind` to broadcast a transaction (`sendrawtransaction`) makes sure it was broadcasted to all our peers before returning. Therefore if we show a success message to the user, the transaction was successfully broadcast no matter how soon after that they close anything after that.

> any reason why 15000 now that we load at bitcoind startup?

Because:
- Storage is *cheap*
- Reducing the prune target in the future is *cheap* but increasing it is *very expensive*
- Setting it too low results in *expensive* (for the user) mistakes, as you already know

For these reasons and because we might have overlooked something, let's not be overconfident with how much we prune the chain. About 2 months of past blocks is already not much.

jp1ac4:
I've now added a simple test for the internal `bitcoind` config and squashed commits.

I think this PR is now ready for review so I'll remove its draft status.

darosior:
The unit test is failing on Windows because the carriage returns aren't encoded in a portable way. (`\n` vs `\r\n`.)

jp1ac4:
> The unit test is failing on Windows because the carriage returns aren't encoded in a portable way. (`\n` vs `\r\n`.)

I should have waited for CI to complete before writing my last message :blush: 

I'll try to fix that now.

gui/src/installer/step/mod.rs
```diff
@@ -111,6 +336,141 @@ fn bitcoind_default_address(network: &Network) -> String {
...
+                message::SelectBitcoindTypeMsg::UseInternal(selected) => {
+                    self.use_internal = selected;
+                }
+            };
+            return Command::perform(async {}, |_| Message::Next);
+        };
+        Command::none()
+    }
+
+    fn apply(&mut self, ctx: &mut Context) -> bool {
```
edouardparis, line 404: most of this code, should be;
in the StartBitcoindStep.load_context() where it populates StartBitcoindStep with information like the network.
then it should creates a command in  the step `load` method with a message returning the 
internal_bitcoind_config and internal_bitcoind_exe_config that the step will handle with the step `update` method`
the command will asynchonously take care of
- file reading like InternalBitcoindConfig::from_file
- `get_available_port` loop 
etc

For clarity you can wrap your step in an enum like:
```
pub enum InternalBitcoindStep {
   Saving{..} // state of the step while the command of the`load()` method is running
   Starting{..} // state of step starting the internal bitcoind step
   Started{..} // state of the step after the command of the start bitcoind is returned
}
```

gui/src/installer/step/mod.rs
```diff
@@ -111,6 +336,141 @@ fn bitcoind_default_address(network: &Network) -> String {
...
+                }
+                ctx.internal_bitcoind_config = Some(conf.clone());
+                if let Err(e) = conf.to_file(&internal_bitcoind_config_path(&ctx.data_dir)) {
+                    self.error = Some(e.to_string());
+                    return false;
+                };
+            }
+        } else {
+            // In case the internal option was selected initially and then user returned to this page, stop bitcoind if it was started.
+            if ctx.internal_bitcoind_config.is_some() {
```
edouardparis, line 456: It should be done in a `reload` method that you can add to the Step trait and that is triggered in the `previous` method of the installer,

edouardparis (commented):
```mermaid
stateDiagram-v2
    SelectBitcoindStep.choose  --> SelectBitcoindStep.type_chosen  : user click
    SelectBitcoindStep.type_chosen --> SelectBitcoindStep.ports_found : update
    SelectBitcoindStep.ports_found  --> context : apply
    context --> SelectBitcoindStep.choose : slct.load_context
    context --> StartBitcoindStep.starting : start.load_context
    StartBitcoindStep.starting  --> StartBitcoindStep.started  : update
    StartBitcoindStep.started --> context : apply
```

gui/src/installer/step/mod.rs
edouardparis: Or maybe even better would be to modify the view so when clicking on previous button, and other message is send to stop the bitcoind daemon then propagate the previous message.

edouardparis:
After reflexion, I would prefer that most of the code is being encapsulate in only one step.

SelectBitcoindTypeStep should only update context.bitcoind_external field for example and in the same way that is implemented correctly here, ExternalBitcoindStep (previously DefineBitcoindStep) and InternalBitcoindStep should skip according to its value in the context.

Then InternalBitcoindStep should look for available ports and file reading in an asynchronous command triggered by `step::load` method. Then it updates itself with result of the command, then try to trigger the start of the bitcoind daemon.
It will be way easier then to have a middle state to Download the binary.

The internal state of the state could be considered as an enum that will evolve:
```rust
 InternalBitcoindStatus {
   Loading{..} // waiting command that search for available ports and read files
   Downloading{..} // downloading the binary (not in this PR)
   Starting{..} // starting the daemon
   Started{..} // started with result
 }
```

The InternalBitcoindStep should have the previous button in its view that does not send immediately the Message::Previous but an other message that say to the step to erase everything and stop everything before sending the message::Previous so it goes back in the process.

pythcoiner:
some GUI glitch there:


also maybe the label `let Liana install and manage a pruned Bitcoin node` should be replaced by `let Liana manage a (pruned) Bitcoin node` until the install feature is implemented to avoid users expect Liana to install bitcoind?

gui/src/installer/view.rs
```diff
@@ -885,6 +888,128 @@ pub fn define_bitcoin<'a>(
...
+                                )
+                                .align_items(Alignment::Center),
+                        )
+                        .padding(20),
+                    ),
+            )
+            .push_maybe(error.map(|e| card::invalid(text(e)))),
+        true,
+    )
+}
```
pythcoiner, line 937: this looks better on my side
```suggestion
pub fn select_bitcoind_type(progress: (usize, usize), error: Option<&String>) -> Element<Message> {
    layout(
        progress,
        "Choose Bitcoin installation type",
        Column::new()
            .push(text(prompt::SELECT_BITCOIND_TYPE))
            .spacing(20)
            .push(
                Row::new()
                    .align_items(Alignment::End)
                    .spacing(20)
                    .push(
                        button::primary(None, "I want to manage my own node")
                            .width(Length::FillPortion(2))
                            .on_press(Message::SelectBitcoindType(
                                message::SelectBitcoindTypeMsg::UseInternal(false),
                            )),

                    )
                    .push(
                        button::primary(None, "Let Liana manage my node")
                            .width(Length::FillPortion(2))
                            .on_press(Message::SelectBitcoindType(
                                message::SelectBitcoindTypeMsg::UseInternal(true),
                            )),
                    ),
            )
            .push_maybe(error.map(|e| card::invalid(text(e)))),
        true,
    )
}
```

darosior:
> some GUI glitch there:

This is common to all screens, if it is to be fixed it should be generally and not specifically in this PR.

gui/src/installer/view.rs
jp1ac4: Thanks for this. There may be other changes required on this page, such as more detailed explanations of the different options, so I'll wait for those before making these changes.

jp1ac4:
> also maybe the label `let Liana install and manage a pruned Bitcoin node` should be replaced by `let Liana manage a pruned Bitcoin node` until the install feature is implemented to avoid users expect Liana to install bitcoind?

That's true, it's a bit misleading currently. I think @kloaec is currently preparing some updated wording for this page.

jp1ac4:
Hi @edouardparis, I've made some of the changes you suggested.

> SelectBitcoindTypeStep should only update context.bitcoind_external field for example and in the same way that is implemented correctly here, ExternalBitcoindStep (previously DefineBitcoindStep) and InternalBitcoindStep should skip according to its value in the context.
>
> Then InternalBitcoindStep should look for available ports and file reading in an asynchronous command triggered by step::load method. Then it updates itself with result of the command, then try to trigger the start of the bitcoind daemon.
It will be way easier then to have a middle state to Download the binary.

In https://github.com/wizardsardine/liana/pull/592/commits/5c362e68684d5d3adb7ff7c65caa678a661900c5 and https://github.com/wizardsardine/liana/pull/592/commits/f53732367bce7aaecd24766a5565c9d982718bec, I've put the configuration and starting of internal `bitcoind` in the same step, but I haven't renamed the steps yet ,which I'll do in a later commit.


> The internal state of the state could be considered as an enum that will evolve:
> 
> ```rust
>  InternalBitcoindStatus {
>    Loading{..} // waiting command that search for available ports and read files
>    Downloading{..} // downloading the binary (not in this PR)
>    Starting{..} // starting the daemon
>    Started{..} // started with result
>  }
> ```

I haven't done this yet. It might take me a little bit of time to do nicely, so depending on when the changes need to be ready by, would it be worth leaving this for the follow-up "download bitcoind" PR?

> The InternalBitcoindStep should have the previous button in its view that does not send immediately the Message::Previous but an other message that say to the step to erase everything and stop everything before sending the message::Previous so it goes back in the process.

I've done this in https://github.com/wizardsardine/liana/pull/592/commits/173a6a0ba73768ce2ed5b9f84816828f0f3352c5 and also in https://github.com/wizardsardine/liana/pull/592/commits/7ffd9c268e8ab8e2da80feef3bc7c78dfca9420a I removed the `Apply` message I previously added.

edouardparis:
Having the code more encapsulated its own step makes it clearer,  the general state of the PR is good enough for me and I think it will be easier then to modify the module in other PRs. Good work @jp1ac4

jp1ac4:
I've squashed and rebased my changes.

I also renamed some of the `StartInternalBitcoind...` objects to `InternalBitcoind...`. I haven't renamed `DefineBitcoind` to `ExternalBitcoind` yet so as not to introduce additional changes, but can do so if required.

I've also just added a commit (https://github.com/wizardsardine/liana/pull/592/commits/dd0fdb5d1dc0d88a9083b1b5d0c2c0947e6f3aae) to stop internal `bitcoind` while syncing is in progress, but it doesn't quite work and gives warnings:
```
WARN liana::bitcoin::d:348: Transient error when sending request to bitcoind: Bitcoind RPC server error: transport error: minreq: Connection refused (os error 111)
```
I'll see if I can resolve that.

Apart from that, there may still be some wording/layout changes required on the "select bitcoind type" page as per previous comments.

darosior:
It shouldn't be an issue to stop bitcoind while it's syncing. And the `BitcoinD` impl should be retrying the request if it's temporarily inaccessible.

darosior:
Also just noting @kloaec will give suggestions for improving the wording and/or presentation in the issue https://github.com/wizardsardine/liana/issues/570 so we can keep track of the bigger picture there while moving forward with the PRs. That is, don't consider a blocker for this PR.

jp1ac4:
OK sure, I'll leave further wording/presentation changes for a follow-up PR.

I'll squash my last commit into the others once any other changes that may be required have been identified.

darosior:
Found it. The issue isn't with how you stop `bitcoind`, which is fine. It's that you stop `bitcoind` *before* stopping `lianad`, which makes it crash!

jp1ac4:
I've added the command to stop `bitcoind` after the one for `lianad`, but I guess `lianad` is not fully stopped before the command to stop `bitcoind` is reached. If I add a one-second sleep in between the two commands, it works. Not sure if there's a better way of doing that :thinking:

darosior:
Yeah, in general never sleep with no check. This is racy and a recipe for troubles.

Unfortunately it seems like waiting for a process to terminate in a portable manner is not trivial to achieve in Rust...

darosior:
Ah, this brought up yet another edge case! We shouldn't let the GUI manage the bitcoind but not the lianad. We should check for this independently of this issue.

Assuming now that if there is any bitcoind to stop then lianad was managed by the GUI, we can assume in this case lianad was always started in a thread. So we can just wait for the thread to terminate before trying to stop bitcoind?

gui/src/launcher.rs
```diff
@@ -206,6 +207,25 @@ async fn check_network_datadir(mut path: PathBuf, network: Network) -> Result<Ne
             )
         }
     })?;
+        if let Some(exe_config) = cfg.internal_bitcoind_exe_config {
```
edouardparis, line 210: Launcher is an intermediary step to check if an install already exist for the existing network and if none it redirects user to the install process. I believe this piece of code should be done on the loader side.

jp1ac4: OK sure. Would `Loader::on_load()` be a suitable place, just before it tries to start the Liana daemon in one of the error `match` arms?

darosior:
> Ah, this brought up yet another edge case! We shouldn't let the GUI manage the bitcoind but not the lianad. We should check for this independently of this issue.

I looked into this, it seems it would be trivial since there is a `is_external()` method to `self.daemon`.

> Assuming now that if there is any bitcoind to stop then lianad was managed by the GUI, we can assume in this case lianad was always started in a thread. So we can just wait for the thread to terminate before trying to stop bitcoind?

I looked into this and that should already be the behaviour. Still, there is a thread (apparently the poller's syncing logic) that is still making requests to bitcoind after we've stopped it. I'll investigate.

darosior:
Ok, what a mess.

I've noticed an empty (!!) `else` branch here:
https://github.com/wizardsardine/liana/blob/bdeaf93ec8809e14b3533f6af4af848dca798ade/gui/src/loader.rs#L177-L178

Introducing a log there exposed that the line above it actually always returns `None`:
https://github.com/wizardsardine/liana/blob/bdeaf93ec8809e14b3533f6af4af848dca798ade/gui/src/loader.rs#L174

Therefore, this is an existing bug in the GUI. The Liana daemon was never cleanly shut down when closing the window during the synchronization progress step. Stopping the internal `bitcoind` after this line exposed this bug which was silently ignored (the empty `else` clause), by causing another bug noticed by @jp1ac4 above.

cc @edouardparis: it looks like fixing this would be non-trivial for me, could you have a look?

jp1ac4:
>> Ah, this brought up yet another edge case! We shouldn't let the GUI manage the bitcoind but not the lianad. We should check for this independently of this issue.
>
> it seems it would be trivial since there is a `is_external()` method to `self.daemon`.

I could add this check when deciding whether to stop the internal `bitcoind`. E.g. in case the user had started their internal `bitcoind` and `lianad` manually when returning to Liana after the initial installation.

gui/src/launcher.rs
edouardparis: I think it should be done in the 'start_daemon' method of the loader, that you can rename in 'start_bitcoind_and_daemon'
because if a daemon is already running it means that a bitcoind is also running.

jp1ac4: Thanks for that. I've made this change in https://github.com/wizardsardine/liana/pull/592/commits/783d93038d5575f880a97fc61e290b5bd96e8fe0.

darosior:
We fixed the issue in #623 and #624. Also #625 should help make the Liana daemon shutdown smoother. You should be able to rebase on this very soon.

darosior:
The last of the above PR was merged. This should fix the bug experienced here once this is rebased.

jp1ac4:
> The last of the above PR was merged. This should fix the bug experienced here once this is rebased.

Thanks for that. I'll rebase and also make sure to check the Liana daemon is not external before stopping `bitcoind`.

> And we should also forbid to have an internal bitcoind config set at startup if we connect to an external Liana daemon.

If an external Liana daemon is running, I could ignore any internal bitcoind config and not attempt to start `bitcoind`. Perhaps for now I'll leave the internal bitcoind config in `gui.toml` in case they want to revert to using it later.

darosior:
I'm not saying we should remove it from the `gui.toml`, but that we should error explicitly (instead of silently ignoring something that is invalid).

jp1ac4:
@edouardparis If the user closes the installer on the `InternalBitcoindStep` after starting `bitcoind`, the config is not yet written to the context so it's tricky to stop it in `Installer::stop()`. Would it make sense to add a `stop()` method to the `Step` trait that is empty by default, but for this particular step would stop the internal `bitcoind` if it's been started? Then we could call `current_step.stop()` at the start of `Installer::stop()`.

edouardparis:
> @edouardparis If the user closes the installer on the InternalBitcoindStep after starting bitcoind, the config is not yet written to the context so it's tricky to stop it in Installer::stop(). Would it make sense to add a stop() method to the Step trait that is empty by default, but for this particular step would stop the internal bitcoind if it's been started? Then we could call current_step.stop() at the start of Installer::stop().

Good for me !

edouardparis:
> @edouardparis Perhaps we could check this in Loader::on_load() in case res is Ok(daemon) (so external). We could proceed only if self.gui_config.internal_bitcoind_exe_config.is_none() and otherwise throw an error?

Looks like it is the good way to do it, but I disagree that is should be considered as an invalid behaviour.
I think it is cool to be able to launch an external daemon and connect liana-gui to it for a particular setup without having to modify the  gui config. Wdyt darosior ?

jp1ac4:
I've rebased on master and added two new commits:
- https://github.com/wizardsardine/liana/pull/592/commits/90931c2f7c8ddae6e4b3fbb29159040fa5b45e8e - stiop bitcoind only if lianad is external
- https://github.com/wizardsardine/liana/pull/592/commits/6c8a14e39505d765a691f2364f5196e1d4bad45d - make sure bitcoind is stopped if installer is closed on same step that bitcoind is started

I still sometimes get that transient error when closing Liana during syncing. Would it be worth adding an extra sleep before I stop bitcoind to make it less likely?

darosior:
> I think it is cool to be able to launch an external daemon and connect liana-gui to it for a particular setup without having to modify the gui config. Wdyt darosior ?

So, it would be about supporting someone (presumably non-technical) who used an internal bitcoind during the installation (presumbaly sync'd it and all) eventually starts Liana as a daemon, wants to attach their GUI to the running daemon. So that presumes this person would have sync'd another, external, bitcoind beforehand. All this while not requiring them to modify their GUI config file, because they may come back to use the internal bitcoind.

To be honest this sounds quite unrealistic. And compared to having the error be explicit if someone tries to connect to an externally-running `lianad` by mistake this seems inferior: both are unlikely, but the latter a little bit less?

In any case i don't have a strong opinion. @jp1ac4 do whatever you find suitable.

darosior:
> I still sometimes get that transient error when closing Liana during syncing.

Arf i could reproduce it, let me look into it...

jp1ac4:
> To be honest this sounds quite unrealistic. And compared to having the error be explicit if someone tries to connect to an externally-running `lianad` by mistake this seems inferior: both are unlikely, but the latter a little bit less?
>
> In any case i don't have a strong opinion. @jp1ac4 do whatever you find suitable.

I think both approaches have their merits and I'm happy with either. I suppose as long as Liana still works correctly, it should still be fine to let them continue with an external daemon. Perhaps as a middle option, I'll just add a `warn!` for now to say that the internal bitcoind config is being ignored in case they're using an external daemon.

jp1ac4:
Added https://github.com/wizardsardine/liana/pull/592/commits/0c8e8b7b99977365630564ba9a4eeb9173fa87d9 with the warning.

I think those were all the required changes. If all is well, I can then squash my commits again.

darosior:
Yes, please do.

edouardparis:
Sorry can you fetch master and rebase this PR on it one last time so I can check that the error during sync is gone ?

jp1ac4:
Have just rebased (will squash after you've reviewed the recent commits).

gui/src/app/mod.rs
```diff
@@ -206,3 +216,75 @@ impl App {
         self.state.view(&self.cache).map(Message::View)
     }
 }
+
+/// Possible errors when starting bitcoind.
+#[derive(PartialEq, Eq, Debug, Clone)]
```
edouardparis, line 221: please move all the code from 221 to the end to a new module in `src/bitcoind.rs`, part of the code is also used by loader and we want to keep loader and app code separated.

edouardparis (approved):
I just have a comment and after I think you can squash the commit.
Do you have this error while stopping during sync ?
`2023-08-23T13:56:36.472677Z  WARN liana_gui::app:287: Could not create interface to internal bitcoind: 'Reading bitcoind cookie file: No such file or directory (os error 2)'.
`
It is not very important as my bitcoind is stopped and darosior cannot reproduce it.

darosior:
@edouardparis does it stop the bitcoind though? Most likely not, right?

edouardparis:
I cannot find then the .cookie and no bitcoind running with htop, after I got the error and the app is closed.

jp1ac4:
I can't reproduce that error at the moment, but will keep an eye out in case it appears at some point.

I'll now move the code as per your comment.

gui/src/app/mod.rs
jp1ac4: I've done this in https://github.com/wizardsardine/liana/pull/592/commits/36446048939d4006650b2535fb86256a74603b09 and https://github.com/wizardsardine/liana/pull/592/commits/9f0b92149f609fdc58f4f5ed75461544cad627d5 (forgot to add the new file in first commit :blush:).

jp1ac4:
I did just notice that if I close the GUI quickly during syncing, the internal bitcoind is not stopped. I think this is because it's still on `Step::Started` instead of `Step::Syncing` and so the `Loader::stop()` method misses it. I could try fixing that or otherwise leave it for a follow-up PR.

gui/src/app/mod.rs
darosior: @jp1ac4 feel free to squash your commits

edouardparis:
Opened an issue https://github.com/wizardsardine/liana/issues/631 that will be fix by me in an other PR, so @jp1ac4 can work on the download.

edouardparis (approved):
ACK 698eff7059b1efb9ae92d4a3e3f78113f4c9f5b0
Some little arrangement can be done in other PRs.

jp1ac4:
Thanks to everyone for the reviews and feedback :pray:

PR #597, opened 2023-08-08
Iced 0.10

edouardparis:
wip


darosior:
> Iced 0.10 is enabling renderer fallback from wgpu to tiny-skia if wgpu cannot work.

How does this affect our runtime dependency requirements?

darosior:
Also, leaving a message here as a reminder: we need to check the Guix build and the MSRV here. Also the build documentation may have to be updated.

darosior:
Just leaving a note that we'll target the next release for this, as we are aiming for a v2 feature freeze at the beginning of next week.

edouardparis:
This PR sadly reintroduce the needs for X11 users to have the WINIT_X11_SCALE_FACTOR env var set to 1.
See https://docs.rs/winit/latest/winit/dpi/index.html#how-is-the-scale-factor-calculated

darosior:
What does it mean? Why? Could you be a bit clearer and give more context? It's hard to use our judgement otherwise.

edouardparis:
iced uses logical pixels. Sizes are scaled by your monitor's scale factor to target physical pixels during rendering.
Scale factor is determined by different ways: https://docs.rs/winit/0.28.7/winit/dpi/index.html
If you have X11, it may be possible that winit makes a wrong guess of the screen scale factor, and you can see in the logs:
```
2024-01-05T14:10:20.309331Z  INFO winit::platform_impl::platform::x11::window:156: Guessed window scale factor: 1.1666666666666667
``` 
which explain why the everything seems too big.
`winit` does not see [it as an issue](https://github.com/rust-windowing/winit/issues/2231)
The current way to force it to scale it properly at 1, is to set the env var `WINIT_X11_SCALE_FACTOR=1`

We had the same problem at previous liana releases, [see the troubleshooting section of the gui readme](https://github.com/wizardsardine/liana/tree/master/gui#troubleshooting) 
I am asking to iced devs why we do not have the problem with iced-0.9

This issue affects only linux users using X11.

darosior:
What's left to do here? Would be nice to get it in early so 1) you don't have to maintain such a large diff across conflicts 2) we can catch any bug early on.

edouardparis:
@kloaec found some problems:
with tiny-skia backend: same problem than https://github.com/iced-rs/iced/issues/2173
with wgpu, image are black rectangle:
- https://github.com/iced-rs/iced/discussions/1767
- https://discourse.iced.rs/t/images-arent-properly-loading/84

Next version of iced will reintroduce support of WebGL
https://github.com/iced-rs/iced/blob/master/Cargo.toml#L49
I hope it will allow us to catch up iced new features and have the current behavior with still the same requirements, while bugs are fixed on their side or downstream (wgpu etc)

darosior:
Another comment in case we forget about it: we do still rely on your patched `iced_futures` right?

edouardparis:
Right, I removed it for tests, but I think we still need it, there was no fix yet on iced side
https://github.com/wizardsardine/liana/pull/597/files#diff-6ec93b2f00bf881b5541dd9ebd0611a05738fcf970422f3ae38f13c3a5414650L51

edouardparis:
I managed to patch multiple iced crates in order to bump MSRV to 1.70 instead of 1.74
The bump of Iced made appear some layouts glitch but it does by doing correction of previous "bug" like font size
previously multiplied by 1.2 and `width(Length::Fill)` not interpreted. I corrected the code in https://github.com/wizardsardine/liana/pull/597/commits/ebeb69a9ef11003f609fc572875d7d5b187d65cb and https://github.com/wizardsardine/liana/pull/597/commits/bf5b4326017e9378bc12f58d3cab963e0b35c14e

Now the backend part:
Iced **opengl** feature has still a broken widget which is form input:
20240403_10h55m26s_grim


Iced **wgpu** feature does not have the problem. In general the wgpu renderer is the first class citizen of the library.
20240403_10h52m30s_grim


Iced **tiny-skia** (to test it run `ICED_BACKEND=tiny-skia cargo run`) is the fallback to **wgpu** in case it failed to find support for Vulkan, Metal and DX12. Sadly it has the same problem than the **opengl** feature.
20240403_10h53m56s_grim


So I introduced again the wgpu feature with **tiny-skia** fallback as I believe the tradeoff is ok and we can expect the problem can be fixed by the iced maintainers.

edouardparis:
OP points were about 0.10, but are still relevant

>Font size are now too big, I guess it is because of the new cosmic-text feature of iced, we need to update it.

I talked about it in https://github.com/wizardsardine/liana/pull/597#issuecomment-2033973712
It is fixed.

> Using tiny-skia only make everything very slow with cargo run but the speed is good with release mode.

It is a little bit slow in debug but fast in release. **wgpu** is always fast.

edouardparis:
iced master is also affected
https://github.com/iced-rs/iced/issues/2371

darosior:
This can be rebased now that #1080 was merged.

darosior:
The new branch you use in your Github repo also needs to be updated in the Guix config:
```diff
diff --git a/contrib/reproducible/guix/build.sh b/contrib/reproducible/guix/build.sh
index 1499723..482632a 100755
--- a/contrib/reproducible/guix/build.sh
+++ b/contrib/reproducible/guix/build.sh
@@ -20,7 +20,7 @@ replace-with = "vendored_sources"
 
 [source."https://github.com/edouardparis/iced"]
 git = "https://github.com/edouardparis/iced"
-branch = "fix-futures-recipe"
+branch = "patch-0.12.3"
 replace-with = "vendored_sources"
 EOF
 ```

**Comment by edouardparis** [comment]
I have no error running `cargo vendor` both with rust 1.70 and nightly

**Comment by darosior** [comment]
Ha! It's the version of Cargo used to vendor which is too old..

**Comment by darosior** [comment]
Please update the PR to include my patch from above https://github.com/wizardsardine/liana/pull/597#issuecomment-2082702333.

Here are the hashes i get on this PR with my patch by running
```
BUILD_ROOT=/tmp/tmp.Geqp4c0ail ./contrib/reproducible/guix/guix-build.sh && sha256sum /tmp/tmp.Geqp4c0ail/out/release/lianad && sha256sum /tmp/tmp.Geqp4c0ail/out/gui/release/liana-gui
```

```
Build successful. Output available at /tmp/tmp.Geqp4c0ail/out
779d1ae77ff7baf18df1dc9d1f48d4abb918491d0cd39a28b61d5ea0f8ede46d  /tmp/tmp.Geqp4c0ail/out/release/lianad
8914b89458d6af2983347964eeab66b8188dbb1d7dc90718130ee4f7ec324836  /tmp/tmp.Geqp4c0ail/out/gui/release/liana-gui
```

edouardparis:
sha256sum /tmp/tmp.5xnaVVlBAx/out/release/lianad && sha256sum /tmp/tmp.5xnaVVlBAx/out/gui/release/liana-gui
779d1ae77ff7baf18df1dc9d1f48d4abb918491d0cd39a28b61d5ea0f8ede46d  /tmp/tmp.5xnaVVlBAx/out/release/lianad
8914b89458d6af2983347964eeab66b8188dbb1d7dc90718130ee4f7ec324836  /tmp/tmp.5xnaVVlBAx/out/gui/release/liana-gui

darosior:
Can you update OP? It's quite confusing to have TODO items while this is going to be merged very soon.

edouardparis:
I updated the OP

darosior:
Needs rebase, the last commit is now in master

darosior:
ACK 52e32b6a69c751f68e461f4e858170e4ade6619d -- it's been tested a bunch, in particular by Kevin.

PR #599, opened 2023-08-09
Descriptors: allow the multipath step to be different from `<0;1>`

darosior:
Ledger and wallet policies disallow having more than 2 depth after the
placeholder, therefore we can't do `@1/0/<0;1>/*`, `@1/1/<0;1>/*`, ..
Instead we have to do `@1/<0;1>/*`, `@1/<2;3>/*`, ..
    
Why not? Salvatore also says the cost of deriving another depth is
non-trivial on a signing device.
    
Don't pick a fight with Salvatore, instead just let the GUI (or whatever
creates the desc) use different multipath steps for keys derived from
the same xpubs.

Based on #584. We need to make sure we don't make the assumptions of the multipath step always being `<0;1>` anywhere else the codebase.

### DIFF
--- a/src/descriptors/analysis.rs
+++ b/src/descriptors/analysis.rs
@@ -80,9 +80,9 @@ impl DescKeyChecker {
     /// We require the descriptor key to:
     ///  - Be deriveable (to contain a wildcard)
     ///  - Be multipath (to contain a step in the derivation path with multiple indexes)
-    ///  - The multipath step to only contain two indexes, 0 and 1.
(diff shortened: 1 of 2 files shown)


darosior:
Rebased.

darosior:
Rebased to get the MSRV fix from #584.

edouardparis (approved):
ACK 605c3f657ebfdb92e03937f02e966fa8f9c18557
tested with GUI

PR #601, opened 2023-08-10
`lianad`: print a more helpful error message on startup failure

### LINKED ISSUE #559: Example config files
When I run `lianad`:
```
$ ./liana-1.0-x86_64-linux-gnu/lianad 
Error parsing config: Could not locate the configuration file.
```

I try and figure out what flags may exist here, but that also fails.
```
$ ./liana-1.0-x86_64-linux-gnu/lianad -h
Unknown arguments '["./liana-1.0-x86_64-linux-gnu/lianad", "-h"]'.
Only '--conf <configuration file path>' is supported.

$ ~/Documents/projects/liana$ ./liana-1.0-x86_64-linux-gnu/lianad help
Unknown arguments '["./liana-1.0-x86_64-linux-gnu/lianad", "help"]'.
Only '--conf <configuration file path>' is supported.
```
Please show examples of available config options.

I like how LND has an example available on the top level of the repo:
https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf

darosior:
Specifically, link to a sample configuration file. How are users supposed to know how to write the config file otherwise?

Fixes #559.

### DIFF
--- a/src/bin/daemon.rs
+++ b/src/bin/daemon.rs
@@ -13,8 +13,9 @@ fn parse_args(args: Vec<String>) -> Option<PathBuf> {
     }
 
     if args.len() != 3 {
-        eprintln!("Unknown arguments '{:?}'.", args);
-        eprintln!("Only '--conf <configuration file path>' is supported.");
+        eprintln!("A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.");
+        eprintln!("A documented example of the configuration file is available at https://github.com/wizardsardine/liana/blob/607c0abddab2ff5761b371c6020b0793b1a1bc97/contrib/lianad_config_example.toml.");
+        eprintln!("The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.");
         process::exit(1);
     }
 
--- a/src/bin/daemon.rs
+++ b/src/bin/daemon.rs
@@ -9,7 +9,7 @@ use liana::{config::Config, DaemonHandle};
 
 fn print_help_exit() {
     eprintln!("A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.");
-    eprintln!("A documented sample is available at https://github.com/wizardsardine/liana/blob/607c0abddab2ff5761b371c6020b0793b1a1bc97/contrib/lianad_config_example.toml.");
+    eprintln!("A documented sample is available at 'contrib/lianad_config_example.toml' in the source tree (https://github.com/wizardsardine/liana/blob/v1.0/contrib/lianad_config_example.toml).");
     eprintln!("The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.");
     process::exit(1);
 }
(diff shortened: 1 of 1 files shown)


darosior:
```
$ ./target/debug/lianad 
Error parsing config: Could not locate the configuration file.
A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.
A documented sample is available at https://github.com/wizardsardine/liana/blob/607c0abddab2ff5761b371c6020b0793b1a1bc97/contrib/lianad_config_example.toml.
The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.
```

```
$ ./target/debug/lianad --help
A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.
A documented sample is available at https://github.com/wizardsardine/liana/blob/607c0abddab2ff5761b371c6020b0793b1a1bc97/contrib/lianad_config_example.toml.
The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.
```

```
$ ./target/debug/lianad --conf doesnt_exist.toml
Error parsing config: Could not locate the configuration file.
A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.
A documented sample is available at https://github.com/wizardsardine/liana/blob/607c0abddab2ff5761b371c6020b0793b1a1bc97/contrib/lianad_config_example.toml.
The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.
```

pythcoiner:
looks clear to me

kornpow:
The example config file is clear and helpful

kornpow:
I am not a huge fan of having a direct link to github in the help text, especially since it holds that big blob random large number. Potentially just say a sample is available at the source at `contrib/lianad_config_example.toml`

pythcoiner:
I agree on the blob (hash?) looks ugly, but feels more easy to user to just copy/paste the link, instead of finding the repo + file path.

kornpow:
Could just link to the master branch if we must have a link:
https://github.com/wizardsardine/liana/blob/master/contrib/lianad_config_example.toml

darosior:
Thanks a lot for the feedback here.

> I am not a huge fan of having a direct link to github in the help text [..] Could just link to the master branch if we must have a link

I'm happy to update the message to both point at the file in the source, but prefer to keep the link as well. I figure it would be impose unnecessary burden on the user to vaguely tell them how they could find the file instead of just linking to it.
Note in addition this is a permalink (with the commit hash) instead of a link to the master branch. This is because the master branch is subject to change, contrary to binaries once distributed. Using a permalink makes sure the resource is:
- Still accessible if one day we move the file in the source tree
- Still relevant for the matching software version. For instance if we add new necessary options in the config file in a future version, the link will still point to the config file sample without those.

pythcoiner:
> * Still relevant for the matching software version. For instance if we add new necessary options in the config file in a future version, the link will still point to the config file sample without those.

good point


> but does that mean that now you must keep this blob hash updated?

also wonder about this point



> It appears that tag URLs work as well.
> https://github.com/wizardsardine/liana/blob/v1.1/contrib/lianad_config_example.toml

I think it's much better than the blob hash

darosior:
I've pushed a version pointing to the file within the source tree and used a tag in the URL in place of the commit hash as suggested.

A couple responses to your questions.

> does that mean that now you must keep this blob hash updated?

Well if we update the config, we need to update the sample and point to the new sample yes.

> seemingly a fairly random commit?

It's simply a permalink to a specific point when the config sample matched the binary's. It doesn't matter the specific point (here a commit) as long as the config file matches.
I've updated it to use a tag though as you suggested..

darosior:
self-ACK d4da2bf6b451de0cd44630c077b83426ed581940 -- trivial.

PR #606, opened 2023-08-11
gui: installer: convey registering on a signing device isn't always required

### LINKED ISSUE #545: Installer: convey in "Register descriptor" step that registration is not necessary if they are not using a signing device
When you create a new wallet with only 'hot' keys  (local and remote), liana might NOT ask user to register descriptor on HW signing device, because customer don't have one...
image

darosior:
As discussed in #545, in some cases we nudge the user to register the descriptor on their signing device although they might not have one. Those cases only ever arise when importing a descriptor (either when recovering from backup or participating in the creation of a descriptor on another laptop), since when the descriptor is created beforehand we can simply detect whether a signing device was used and thereby needs to be registered (implemented since https://github.com/wizardsardine/liana/pull/470). 

Therefore, detect when the registration step arises as part of an import process and if so adjust the language to convey registration on a signing device may not be necessary.

Result:
image

Fixes #545.

### DIFF
--- a/gui/src/installer/mod.rs
+++ b/gui/src/installer/mod.rs
@@ -113,7 +113,7 @@ impl Installer {
                     DefineDescriptor::new(self.signer.clone()).into(),
                     BackupMnemonic::new(self.signer.clone()).into(),
                     BackupDescriptor::default().into(),
-                    RegisterDescriptor::default().into(),
+                    RegisterDescriptor::new_create_wallet().into(),
                     DefineBitcoind::new().into(),
                     Final::new(hot_signer_fingerprint).into(),
                 ];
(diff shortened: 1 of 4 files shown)


darosior:
In the future we could ask them beforehand whether they'll be using a signing device and only show them this step if so. But until we do that it's a minimal patch for the current misleading behaviour.

edouardparis:
When importing the descriptor, user needs to store again the ledger HMAC. I am ok to make the step more easy to skip for the import descriptor process, but user with ledger will have to do it.

darosior:
I don't understand your comment in the context of this PR?

edouardparis (approved):
ACK bba5cd63e81a491c28eddc14e4087ba179794a3c

PR #610, opened 2023-08-14
Release packaging improvements

darosior:
This cleans up the MacOS distribution and adds Debian packaging to the release script. This gives users of Debian-based distributions (Ubuntu, etc..) a simple way of installing our software.

Fixes https://github.com/wizardsardine/liana/issues/519.


darosior:
So i looked into having an Arch package. We could have a PKGBUILD that untars the `.deb` to distribute a `liana-bin` package, but that can't be included in our release script (since the PKGBUILD needs to be fetching the .deb from the internet).

Anyways, let's have a package for Debian-based system already. It must be 95% of the desktop distributions used. We can always add other package types (Pacman, RPM) in future releases.

darosior:
Let's not merge this until we are actually release 2.0, since the README would be incorrect for 1.0.

kloaec:
> So i looked into having an Arch package. We could have a PKGBUILD that untars the `.deb` to distribute a `liana-bin` package, but that can't be included in our release script (since the PKGBUILD needs to be fetching the .deb from the internet).
> 
> Anyways, let's have a package for Debian-based system already. It must be 95% of the desktop distributions used. We can always add other package types (Pacman, RPM) in future releases.

While I understand it's hard to track, 95% seems waaaay far off. Manjaro and EndeavourOS are amonst the most popular distros, and are both Arch based. Reality is probably close to 50% Arch based at this stage, especially for newbies (Manjaro especially). Steam survey of last month gives Arch-based more than 50%. https://itsfoss.com/content/images/2023/08/August-2023.png

darosior:
Fair enough. I'd be interested in the number of persons polled, but that seems fair (and surprising to me!) numbers.

I can have another look at not using `makepkg` along with a PKGBUILD but instead building the archive manually myself.
I'd welcome your input if you have any regarding:
- The format of the Pacman package we should distribute. (that's a `.pkg.tar.zst` right?)
- The method of creating a binary Pacman package *before* having released the other packages.

pythcoiner:
> Fair enough. I'd be interested in the number of persons polled, but that seems fair (and surprising to me!) numbers. I can have another look at not using `makepkg` along with a PKGBUILD but instead building the archive manually myself. I'd welcome your input if you have any regarding: - The format of the Pacman package we should distribute. (that's a `.pkg.tar.zst` right?) - The method of creating a binary Pacman package *before* having released the other packages.

can have a look about that if it can wait until this week end

darosior:
We discussed it offline with @kloaec. We'll just push a PKGBUILD to AUR after each release, and not release some sort of package on Github: instead we'll document Arch user can use `yum` or whatever other wrapper to get Liana from AUR.

darosior:
ACK 2888db425df61a5e64ce12f9425bbef7cb3452b6

PR #612, opened 2023-08-16
bitcoind: Make `BitcoinD` public and add `stop` method

jp1ac4:
This is needed as part of https://github.com/wizardsardine/liana/pull/592 for both checking connectivity to the internal `bitcoind` and stopping it when closing Liana.


darosior:
utACK 3372a8d461f1b06cff08d913b45328e3dcd75da7

PR #621, opened 2023-08-21
Gui labels

edouardparis:
This PR use the lianad update_labels and get_labels commands.
It also introduce new concepts from talks with Kevin and Antoine:
- User spending is creating a payment, when he does not add multiple recipients anymore, he is doing multiple payment in one bitcoin transaction.
- a transaction that have multiple outgoing outputs (multiple payment) is tagged as a 'Batch'.

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -2112,8 +2112,8 @@ dependencies = [
 
 [[package]]
 name = "liana"
-version = "1.0.0"
-source = "git+https://github.com/wizardsardine/liana?branch=master#85d470dd8dd67e6726118fe6dd86f9b4c8d3b0ef"
+version = "2.0.0"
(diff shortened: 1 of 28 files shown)


pythcoiner:
i've build + run, it seems that adding a label on outgoing tx not confirmed will not save the label.

pythcoiner:
> I've built + run, it seems that adding a label on outgoing tx not confirmed will not save the label. edit: only from the `transaction` menu

I dig more into that:
 - the label is well saved but the displayed label in `transaction` menu is not updated, i need to go to another menu and then come back to `transaction` menu to update the label.
 - I wonder why we got a `payment` label and a `transaction` label? can a payment contain several tx? it's not really clear to me...

edouardparis (commented):
Self-ACK 757b53ebab43c05a444ce22c8dd666fb367e6ee0

Issue #622, opened 2023-08-22
The internally-managed daemon isn't properly shut down when stopping the GUI


darosior:
See https://github.com/wizardsardine/liana/pull/592#issuecomment-1686485130, https://github.com/wizardsardine/liana/pull/592#issuecomment-1686621375 and finally https://github.com/wizardsardine/liana/pull/592#issuecomment-1687861737.

The trivial way to go about it would be to add a Mutex. This is a lot of churn, and a lot of added overhead, only to be able to stop the daemon. I'd prefer to avoid this.

A minimal and hacky way of fixing this would be to use some unsafe code:
```patch
--- a/gui/src/loader.rs
+++ b/gui/src/loader.rs
@@ -172,11 +172,15 @@ impl Loader {
         if let Step::Syncing { daemon, .. } = &mut self.step {
             if !daemon.is_external() {
                 info!("Stopping internal daemon...");
-                if let Some(d) = Arc::get_mut(daemon) {
-                    d.stop().expect("Daemon is internal");
-                    info!("Internal daemon stopped");
-                } else {
+                // NOTE: this is a hack. The application will be stopped immediately after exiting
+                // this function, so no other reference should be accessed anymore.
+                while Arc::strong_count(&daemon) > 1 {
+                    unsafe {
+                        Arc::decrement_strong_count(Arc::as_ptr(&daemon));
+                    }
                 }
+                Arc::get_mut(daemon).expect("Strong count is 1 by now").stop().expect("Daemon is internal");
+                info!("Internal daemon stopped");
             }
             if self.gui_config.internal_bitcoind_exe_config.is_some() {
                 if let Some(daemon_config) = daemon.config() {
```
Of course, i'd prefer to avoid this too.

I've also considered adding some internal mutability inside the `DaemonHandle`'s poller.

Another option could be to somehow leverage the `Drop` implementation of the various states of the GUI? We'd stop the internally managed daemon here and eventually stop the internally managed bitcoind.

Finally, i've also considered a poll-based interface on `DaemonHandle` instead of the current `shutdown` that waits for the poller thread to shutdown. This is because waiting for the thread immediately triggers this for the GUI:
![image](https://github.com/wizardsardine/liana/assets/22457751/5150fe4e-0bdd-4d83-b4e9-5ce3da58419a)
So we could instead have some sort of interface like:
1. Trigger shutdown
2. Is it shut down yet?

edouardparis:
[...]

PR #623, opened 2023-08-22
lib: non-blocking daemon shutdown

darosior:
This makes it possible to trigger the shutdown of the daemon through the DaemonHandle, without having to block while waiting for the poller thread to join.

Incidently, this allows to avoid having to move `self` which in turns allows to fix a GUI bug (see
https://github.com/wizardsardine/liana/issues/622).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ required-features = ["daemon"]
 [features]
 default = ["daemon"]
 daemon = ["libc"]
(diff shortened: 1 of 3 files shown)


darosior:
Gated it behind an optional Cargo feature.

edouardparis (approved):
ACK 2a6b775f607dc0b56be874807bd36427922ac681

PR #625, opened 2023-08-22
bitcoin: looper: avoid large sleeps when bitcoind is syncing

darosior:
Sleeping for 30 whole seconds impedes the shutdown check. Sleep only .5s but still only do poll bitcoind one every 30s when it's syncing.

It was reported to make the GUI hang when closing it while syncing.


edouardparis (approved):
utACK 0d5041ca4a972a946244643db738e6d35b0d114e

PR #626, opened 2023-08-22
tests: don't wait for lianad to log on shutdown

darosior:
It seems to be causing some races that i don't want to be investigating at the moment. It's redundant anyways as we are already checking for the return code of the process.

It's fixing the annoying "Process died while waiting for logs" when stopping the daemon. AKA:
> Yes it died i stopped it that's the point.


darosior:
Just waiting for CI to pass as a sanity check of my local testing.

darosior:
self-ACK 97abc4699c7343e4be6a0596a25481f9e07bb3bb -- trivial

PR #627, opened 2023-08-22
bitcoind: fix the detection of unconfirmed RBF spending txs

darosior:
As we walk through the spend transactions in the wallet, we may return the txid of the transaction that was replaced instead of the new one.

Note: this fixes the `test_conflicting_unconfirmed_spend_txs` flakiness, which exposed this bug.


darosior:
Rebased after #628.

darosior:
ACK 421c1af6cd90f1b1af77c303b2ba2e74760b8781 -- tested it locally by running the `test_conflicting_unconfirmed_spend_txs` functional test around a thousand times.

PR #634, opened 2023-08-25
fix gui loader: stop bitcoind if running after close request

### LINKED ISSUE #631: gui loader: stop() does not shutdown bitcoind for every enum state of the loader

edouardparis:
close #631

### DIFF
--- a/gui/src/loader.rs
+++ b/gui/src/loader.rs
@@ -181,13 +181,21 @@ impl Loader {
                 info!("Stopping internal daemon...");
                 daemon.stop();
                 info!("Internal daemon stopped");
-                if self.gui_config.internal_bitcoind_exe_config.is_some() {
-                    if let Some(daemon_config) = daemon.config() {
-                        if let Some(bitcoind_config) = &daemon_config.bitcoind_config {
-                            stop_internal_bitcoind(bitcoind_config);
-                        }
-                    }
+            }
+        }
+        if self.gui_config.internal_bitcoind_exe_config.is_some() {
+            if let Ok(daemon_config) =
+                Config::from_file(self.gui_config.daemon_config_path.as_ref().cloned())
+            {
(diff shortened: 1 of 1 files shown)


darosior:
Nice!

Just to make sure: there is no other state (aside from a crash say) in which

darosior:
we wouldn't check for an internal bitcoind to stop?

(Sorry about that, writing from my mobile)

gui/src/loader.rs
```diff
@@ -181,13 +181,21 @@ impl Loader {
...
                 info!("Internal daemon stopped");
-                if self.gui_config.internal_bitcoind_exe_config.is_some() {
-                    if let Some(daemon_config) = daemon.config() {
-                        if let Some(bitcoind_config) = &daemon_config.bitcoind_config {
-                            stop_internal_bitcoind(bitcoind_config);
-                        }
-                    }
+            }
+        }
+        if self.gui_config.internal_bitcoind_exe_config.is_some() {
```
jp1ac4, line 186: Do we need to check that the Liana daemon is not external?

edouardparis: Internal daemon and managed bitcoind are an invariant, you are right that adding a check would be good

jp1ac4: Yes, I think so too. If a user starts their managed bitcoind before opening Liana, then we'll stop it when they close Liana, but it's unlikely they'll do this and so not worth adding extra logic. In any case, I guess this could happen whether the Liana daemon is internal or external. Perhaps we would need a simple flag to indicate if we had started bitcoind.

darosior: In this case maybe remove the log below?
```rust
info!("Stopped liana managed bitcoind");
```

As this looks like a success message whereas it might have failed.

edouardparis: Good idea, addressed in eb10cdacd6ed42c7294f454da62381e3e91269dc

PR #636, opened 2023-08-28
doc: polish the guide on restoring a Liana wallet backup with Bitcoin Core

darosior:


darosior:
self-ACK b5b96b95937ef914aed237b7829a71ff3672c3f4 -- it's only doc

PR #645, opened 2023-08-29
bitcoind: use the node client to make node requests

darosior:
Props to Kevin for finding this rather significant bug. We should really be catching this type of bugs with code review..


darosior:
Ironical self-ACK 9255b5ed941f04e56de4f32251fbe9b3d692baaa

PR #651, opened 2023-08-29
lib: windows: don't unload the wo wallet if not loaded

darosior:


darosior:
self-ACK 34f3946fc814bf5c915be131dd3092b21f4f48e1

PR #654, opened 2023-08-30
Add bitcoind logs to loader

edouardparis:
based on #630


darosior:
Thanks. Now that we log them on failure, do you think it's still worth logging everything?

PR #657, opened 2023-08-30
cleanup bitcoind process management

edouardparis:
- Small rearrangements of the bitcoind module.
- Add a wrapper around the bitcoind child process and share it to the states:
  Installer -> loader -> app
- It makes it more easy to share the information if the gui started a bitcoind process, no need anymore to check multiple configuration files, just answer the question: do we have a child process started ?

### DIFF
--- a/gui/src/bitcoind.rs
+++ b/gui/src/bitcoind.rs
@@ -1,8 +1,7 @@
 use liana::{config::BitcoindConfig, miniscript::bitcoin};
+use std::path::Path;
 
 use tracing::{info, warn};
-
-use crate::app::config::InternalBitcoindExeConfig;
 
 #[cfg(target_os = "windows")]
 use std::os::windows::process::CommandExt;
@@ -51,10 +50,10 @@ impl std::fmt::Display for StartInternalBitcoindError {
 /// Start internal bitcoind for the given network.
 pub fn start_internal_bitcoind(
     network: &bitcoin::Network,
-    exe_config: InternalBitcoindExeConfig,
+    bitcoind_datadir: &Path,
+    exe_path: &Path,
 ) -> Result<std::process::Child, StartInternalBitcoindError> {
-    let datadir_path_str = exe_config
-        .data_dir
+    let datadir_path_str = bitcoind_datadir
         .canonicalize()
         .map_err(|e| StartInternalBitcoindError::CouldNotCanonicalizeDataDir(e.to_string()))?
         .to_str()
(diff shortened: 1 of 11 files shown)


gui/src/bitcoind.rs
```diff
@@ -85,9 +87,9 @@ impl Bitcoind {
         #[cfg(target_os = "windows")]
         let command = command.creation_flags(CREATE_NO_WINDOW);
 
-        let process = command
+        let mut process = command
             .args(&args)
-            .stdout(std::process::Stdio::null()) // We still get bitcoind's logs in debug.log.
+            .stdout(std::process::Stdio::piped()) // We still get bitcoind's logs in debug.log.
```
darosior, line 92: Drop the comment if you are removing this?

gui/src/bitcoind.rs
```diff
@@ -114,6 +116,7 @@ impl Bitcoind {
             .map_err(|e| StartInternalBitcoindError::BitcoinDError(e.to_string()))?;
 
         Ok(Self {
+            stdout: process.stdout.take().map(|s| Arc::new(Mutex::new(s))),
```
darosior, line 119: It seems `stderr` would be what we'd be looking for to debug. If we are going to only log stdout maybe we could do a trick like piping stderr to stdout. Or maybe it's fine to just read stderr when we actually detect an error as we do currently.

darosior: Ignore this, i hadn't seen you actually displayed them on the screen.

darosior (commented):
ACK 6d8c9e437f6cc36d479f4be33779834b2180507d

gui/src/bitcoind.rs
darosior: We can do this in a followup

gui/src/loader.rs
```diff
@@ -125,12 +143,30 @@ impl Loader {
...
+        res: Result<(Arc<dyn Daemon + Sync + Send>, Option<Bitcoind>), Error>,
+    ) -> Command<Message> {
         match res {
-            Ok(daemon) => {
+            Ok((daemon, bitcoind)) => {
+                // bitcoind may have been already started and given to the loader
+                // We should not override with None the loader bitcoind field
+                if let Some(bitcoind) = bitcoind {
+                    self.internal_bitcoind = Some(bitcoind);
+                }
```
darosior, line 165: I'm not sure i can understand this comment. Shouldn't this be checking `self.internal_bitcoind` is `None` before overriding it, if the meaning of the comment is what i think it is?

edouardparis: The iced command that start_bitcoind_and_daemon, returns the daemon interface and an Option< Bitcoind> wrapper if it has started bitcoind, but If bitcoind is already started and stored in loader, the the result must be None and should not override the self.internal_bitcoind

PR #659, opened 2023-08-31
gui: installer: more robust poll for cookie file at bitcoind startup

darosior:
We would formerly wait a definite number of seconds for the cookie file to appear. But this approach presents a fundamental issue: how long is enough before we can be reasonably sure bitcoind will never start?

For instance on mainnet the checks performed at startup could take more than the 3 seconds we would previously wait for. If it does we would have incorrectly errored and let bitcoind run in the background.

Instead, assume bitcoind would exit if it errors. Until it does so, continue polling for the cookie file. This approach also presents drawbacks (for instance what if bitcoind is performing a very long operation before creating the cookie file?), but the former approach wouldn't be an acceptable solution in this case either. And the new one is preferable as its failure scenario seems much less probable.

### DIFF
--- a/gui/src/bitcoind.rs
+++ b/gui/src/bitcoind.rs
@@ -4,6 +4,8 @@ use liana::{
 };
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::thread;
+use std::time;
 use tokio::sync::Mutex;
 
 use tracing::{info, warn};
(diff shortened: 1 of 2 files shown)


edouardparis (commented):
ACK 76bb74371f7d33603ac0864b9ecea014a3cc0cda

PR #658, opened 2023-08-31
Change gui config and gui internal bitcoind directory architecture

edouardparis:
1. added the bitcoind parent directory
```
.liana/bitcoind
├── bitcoin-25.0
│   └── bin
└── datadir
    ├── anchors.dat
    ├── banlist.json
    ├── bitcoin.conf
    ├── blocks
    ├── chainstate
    ├── debug.log
    ├── fee_estimates.dat
    ├── mempool.dat
    ├── peers.dat
    └── settings.json
```

2. change gui config to have `start_internal_bitcoind` : `bool`

PR #661, opened 2023-08-31
TMP: DONOTMERGE: fix a race when starting bitcoind

darosior:
This is a quick-and-dirty demonstration of how to fix the bug @jp1ac4 caught.

### DIFF
--- a/gui/src/loader.rs
+++ b/gui/src/loader.rs
@@ -70,7 +70,7 @@ pub enum Message {
             Error,
         >,
     ),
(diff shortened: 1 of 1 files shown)


darosior:
The issue was that the stop of bitcoind was recently (#657, #658) made conditional on the data structure having `self.internal_bitcoind` set. But since bitcoind was started asynchronously and only thereafter stored into `self.internal_bitcoind` it was possible (and indeed quite likely since we waited on the Liana daemon to start before setting `self.internal_bitcoind`) that bitcoind be started *before* `self.internal_bitcoind` would be set.

PR #662, opened 2023-08-31
loader: try to stop bitcoind if connected

edouardparis:


gui/src/loader.rs
```diff
@@ -225,6 +225,15 @@ impl Loader {
 
         if let Some(bitcoind) = &self.internal_bitcoind {
             bitcoind.stop();
+        } else if self.gui_config.start_internal_bitcoind {
+            if let Ok(config) = Config::from_file(self.gui_config.daemon_config_path.clone()) {
+                if let Some(bitcoind_config) = &config.bitcoind_config {
+                    let mut stopped = false;
+                    while !stopped {
+                        stopped = stop_bitcoind(&bitcoind_config);
+                    }
```
darosior, line 234: Sleep for half a second?

gui/src/loader.rs
```diff
@@ -225,6 +225,15 @@ impl Loader {
 
         if let Some(bitcoind) = &self.internal_bitcoind {
             bitcoind.stop();
+        } else if self.gui_config.start_internal_bitcoind {
```
darosior, line 228: What if it gets called while we are performing the `connect`? This would be an infinite loop. I think we should set a flag right before calling `start_bitcoind_and_daemon`.

gui/src/loader.rs
```diff
@@ -225,6 +225,15 @@ impl Loader {
...
+            if let Ok(config) = Config::from_file(self.gui_config.daemon_config_path.clone()) {
+                if let Some(bitcoind_config) = &config.bitcoind_config {
+                    let mut stopped = false;
+                    while !stopped {
+                        stopped = stop_bitcoind(&bitcoind_config);
+                    }
+                }
+            }
```
darosior, line 236: Maybe log in the `else` clauses here.

gui/src/loader.rs
darosior: And limit the number of tries. Like 10 tries, sleeping .5s between each?

darosior (commented):
I think that's the right approach but as implemented it could create new issues

gui/src/loader.rs
```diff
@@ -225,12 +229,13 @@ impl Loader {
...
         if let Some(bitcoind) = &self.internal_bitcoind {
             bitcoind.stop();
-        } else if self.gui_config.start_internal_bitcoind {
+        } else if self.waiting_daemon_bitcoind && self.gui_config.start_internal_bitcoind {
             if let Ok(config) = Config::from_file(self.gui_config.daemon_config_path.clone()) {
                 if let Some(bitcoind_config) = &config.bitcoind_config {
                     let mut stopped = false;
                     while !stopped {
-                        stopped = stop_bitcoind(&bitcoind_config);
+                        stopped = stop_bitcoind(bitcoind_config);
```
darosior, line 237: nit: this could just be
```rust
while !stop_bitcoind() {
}
```

gui/src/loader.rs
```diff
@@ -225,12 +229,13 @@ impl Loader {
...
+                        std::thread::sleep(std::time::Duration::from_secs(1))
```
darosior, line 238: I think 1 second is too long and would trigger Gnome's "application isn't responding". Can you go for half a sec?

darosior (commented):
I think we need to limit how many times we try and fail to connect to bitcoind to stop it.

gui/src/loader.rs
```diff
@@ -160,6 +163,7 @@ impl Loader {
     ) -> Command<Message> {
         match res {
             Ok((daemon, bitcoind)) => {
+                self.waiting_daemon_bitcoind = false;
```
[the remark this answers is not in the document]
edouardparis: If I put it under the condition below, I should rename it for waiting_bitcoind

darosior:
Tested locally, this fixes the issue we've encountered.

darosior:
ACK 0a432d081dbdfce72e11751e080a941e24dd04ff

Issue #291, opened 2023-01-19
Allow users to set the account for the derivation path in the installer


darosior:

edouardparis:
#289 directly select accounts for users in the create wallet panel and let user generate multiple key with incremented indexes in the participate panel.

darosior:
Yeah, let's keep tracking this for power users that want to set the account themselves because, say, they are using this signing device at this derivation path elsewhere already.

kloaec:
NACK for the general case, ACK for an advanced option/display.
I would use this feature personally.

PR #682, opened 2023-09-01
GUI: installer: don't generate multiple times the same xpub in "participate"

darosior:
This is a quick and dirty fix to be backported. Doing something like #291 could make sense in the future.


darosior:
Thanks to @kloaec for spotting this bug.

edouardparis (approved):
utACK 9d2309ea5e7ee91eb0c06777153df59249d88926

PR #685, opened 2023-09-02
gui: use patch for iced_futures

### LINKED ISSUE #684: GUI: after running a random amount of time we got lot of TrySendError full errors in the logs and a freeze
While listening for close request in the application subscription, it seems that the subscription tracker is clogged.

edouardparis:
close #684

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -1750,8 +1750,7 @@ dependencies = [
 [[package]]
 name = "iced_futures"
 version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "215d51fa4f70dbb63775d7141243c4d98d4d525d8949695601f8fbac7dcbc04e"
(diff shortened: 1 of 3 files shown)


darosior (approved):
From my limiting reading and understanding of Iced and Winit, your patch makes sense to me. We definitely need to keep the discussion in https://discourse.iced.rs/t/a-problem-and-a-possible-solution-for-clogged-mpsc-channel-in-iced-futures/50 going though, as we may be missing something.

I need to look into adapting the GUIX build to your patched Iced, that shouldn't be too complicated.

darosior:
This also needs testing on Windows, @kloaec will take care of this in about a hour.

darosior:
ACK 3019608

darosior:
You probably want to PR this to master as well?

darosior:
If you do, maybe also include #687.

edouardparis:
Yes, doing it right now

PR #697, opened 2023-09-07
contrib: update Arch PKGBUILD to v2.0

darosior:

PR #699, opened 2023-09-07
Backport #697

darosior:


darosior:
Could also add 34afbe415bf475c75427b74fcfbe1feb9e219b7c here.

darosior:
I don't think that makes a lot of sense since we won't do a 2.1 and we are doing 3.0 now. Closing for now.

PR #700, opened 2023-09-07
commands: optionally filter `listcoins` by status and outpoint

### LINKED ISSUE #676: Add filters to the `listcoins` command
Currently the [`listcoins` command](https://github.com/wizardsardine/liana/blob/master/doc/API.md#listcoins) is bulky. It would return all coins, even historical spent ones. It would be nice to be able to filter by status.

Also, it would be nice to be able to retrieve coins by their outpoints.

Therefore we could add two optional parameters to filter the result of `listcoins`:
- A list of status (which ones? I'm thinking of `unconfirmed`, `confirmed`, `spending` and `spent`.)
- A list of outpoints

jp1ac4:
This is to resolve #676.

I've renamed `CoinType` to `CoinStatus` and updated its values.

The `listcoins` command and related functions can be optionally filtered by coin status and/or outpoint.

Opening this as draft to check that I'm on the right track.

Remaining tasks:
- [x] Update tests
- [x] Update API doc
- [x] Possibly use the updated functions elsewhere, e.g. `list_spending_coins()`

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -6,11 +6,11 @@ mod utils;
 
 use crate::{
     bitcoin::BitcoinInterface,
-    database::{Coin, CoinType, DatabaseInterface},
+    database::{Coin, DatabaseInterface},
     descriptors, DaemonControl, VERSION,
 };
 
-pub use crate::database::LabelItem;
+pub use crate::database::{CoinStatus, LabelItem};
 
 use utils::{
     deser_addr_assume_checked, deser_amount_from_sats, deser_fromstr, deser_hex, ser_amount,
@@ -289,11 +289,15 @@ impl DaemonControl {
         GetAddressResult::new(address)
     }
 
-    /// Get a list of all known coins.
-    pub fn list_coins(&self) -> ListCoinsResult {
+    /// Get a list of all known coins, optionally by status and/or outpoint.
+    pub fn list_coins(
+        &self,
+        statuses: &[CoinStatus],
+        outpoints: &[bitcoin::OutPoint],
+    ) -> ListCoinsResult {
         let mut db_conn = self.db.connection();
         let coins: Vec<ListCoinsEntry> = db_conn
-            .coins(CoinType::All)
+            .coins(statuses, outpoints)
             .into_values()
             .map(|coin| {
                 let Coin {
@@ -747,12 +751,15 @@ impl DaemonControl {
         let timelock =
             timelock.unwrap_or_else(|| self.config.main_descriptor.first_timelock_value());
         let height_delta: i32 = timelock.try_into().expect("Must fit, it's a u16");
-        let sweepable_coins = db_conn.coins(CoinType::Unspent).into_values().filter(|c| {
-            // We are interested in coins available at the *next* block
-            c.block_info
-                .map(|b| current_height + 1 >= b.height + height_delta)
-                .unwrap_or(false)
-        });
+        let sweepable_coins = db_conn
+            .coins(&[CoinStatus::Unconfirmed, CoinStatus::Confirmed], &[])
+            .into_values()
+            .filter(|c| {
+                // We are interested in coins available at the *next* block
+                c.block_info
+                    .map(|b| current_height + 1 >= b.height + height_delta)
+                    .unwrap_or(false)
+            });
 
         // Fill-in the transaction inputs and PSBT inputs information. Record the value
         // that is fed to the transaction while doing so, to compute the fees afterward.
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -85,7 +85,11 @@ pub trait DatabaseConnection {
     ) -> Option<(bip32::ChildNumber, bool)>;
 
     /// Get all our coins, past or present, spent or not.
-    fn coins(&mut self, coin_type: CoinType) -> HashMap<bitcoin::OutPoint, Coin>;
+    fn coins(
+        &mut self,
+        statuses: &[CoinStatus],
+        outpoints: &[bitcoin::OutPoint],
+    ) -> HashMap<bitcoin::OutPoint, Coin>;
 
     /// List coins that are being spent and whose spending transaction is still unconfirmed.
     fn list_spending_coins(&mut self) -> HashMap<bitcoin::OutPoint, Coin>;
@@ -192,8 +196,12 @@ impl DatabaseConnection for SqliteConn {
         self.complete_wallet_rescan()
     }
 
-    fn coins(&mut self, coin_type: CoinType) -> HashMap<bitcoin::OutPoint, Coin> {
-        self.coins(coin_type)
+    fn coins(
+        &mut self,
+        statuses: &[CoinStatus],
+        outpoints: &[bitcoin::OutPoint],
+    ) -> HashMap<bitcoin::OutPoint, Coin> {
+        self.coins(statuses, outpoints)
             .into_iter()
             .map(|db_coin| (db_coin.outpoint, db_coin.into()))
             .collect()
@@ -348,11 +356,29 @@ impl Coin {
     }
 }
 
+/// Possible (mutually exclusive) status of a coin.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub enum CoinType {
-    All,
-    Unspent,
+pub enum CoinStatus {
+    /// Has not yet been included in a block and has no spend transaction.
+    Unconfirmed,
+    /// Has been included in a block and has no spend transaction.
+    Confirmed,
+    /// Has an unconfirmed spend transaction, but coin itself may not yet have been included in a block.
+    Spending,
+    /// Has a confirmed spend transaction.
     Spent,
+}
+
+impl CoinStatus {
+    pub fn from_arg(s: &str) -> Option<CoinStatus> {
+        match s {
+            "unconfirmed" => Some(CoinStatus::Unconfirmed),
+            "confirmed" => Some(CoinStatus::Confirmed),
+            "spending" => Some(CoinStatus::Spending),
+            "spent" => Some(CoinStatus::Spent),
+            _ => None,
+        }
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
--- a/src/database/sqlite/mod.rs
+++ b/src/database/sqlite/mod.rs
@@ -487,7 +487,7 @@ impl SqliteConn {
         .expect("Database must be available")
     }
 
-    /// Mark a set of coins as spent.
+    /// Mark a set of coins as spending.
     pub fn spend_coins<'a>(
         &mut self,
         outpoints: impl IntoIterator<Item = &'a (bitcoin::OutPoint, bitcoin::Txid)>,
--- a/src/database/sqlite/mod.rs
+++ b/src/database/sqlite/mod.rs
@@ -23,7 +23,7 @@ use crate::{
                 maybe_apply_migration, LOOK_AHEAD_LIMIT,
             },
         },
-        Coin, CoinType, LabelItem,
+        Coin, CoinStatus, LabelItem,
     },
     descriptors::LianaDescriptor,
 };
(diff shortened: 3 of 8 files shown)


src/commands/mod.rs
```diff
@@ -747,12 +751,15 @@ impl DaemonControl {
...
             timelock.unwrap_or_else(|| self.config.main_descriptor.first_timelock_value());
         let height_delta: i32 = timelock.try_into().expect("Must fit, it's a u16");
-        let sweepable_coins = db_conn.coins(CoinType::Unspent).into_values().filter(|c| {
-            // We are interested in coins available at the *next* block
-            c.block_info
-                .map(|b| current_height + 1 >= b.height + height_delta)
-                .unwrap_or(false)
-        });
+        let sweepable_coins = db_conn
+            .coins(Some(&[CoinStatus::Confirmed]), None)
```
darosior, line 755: That's a behaviour change. Previously we'd query all unspent coins, including unconfirmed ones. I think just including `CoinStatus::Confirmed` too should fix this.

src/database/mod.rs
```diff
@@ -85,7 +85,11 @@ pub trait DatabaseConnection {
     ) -> Option<(bip32::ChildNumber, bool)>;
 
     /// Get all our coins, past or present, spent or not.
-    fn coins(&mut self, coin_type: CoinType) -> HashMap<bitcoin::OutPoint, Coin>;
+    fn coins(
+        &mut self,
+        statuses: Option<&[CoinStatus]>,
+        outpoints: Option<&[bitcoin::OutPoint]>,
```
darosior, line 91: What is `Some([])` supposed to mean? If it's not necessary, then let's drop the `Option` and use the empty list as describing the absence of filters?

src/database/mod.rs
```diff
@@ -349,12 +357,25 @@ impl Coin {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub enum CoinType {
-    All,
-    Unspent,
+pub enum CoinStatus {
+    Unconfirmed,
+    Confirmed,
+    Spending,
```
darosior, line 363: Let's have a doc comment for each variant about how they are mutually exclusive?

src/database/sqlite/mod.rs
```diff
@@ -357,18 +357,75 @@ impl SqliteConn {
...
+        } else {
+            String::new()
+        };
+        // SELECT * FROM coins WHERE (txid, vout) IN ((txidA, voutA), (txidB, voutB));
+        let op_condition = if let Some(outpoints) = outpoints {
+            if !outpoints.is_empty() {
+                let mut cond = "(txid, vout) IN (VALUES ".to_string();
+                for (i, outpoint) in outpoints.iter().enumerate() {
+                    // NOTE: SQLite doesn't know Satoshi decided txids would be displayed as little-endian
+                    // hex.
```
darosior, line 397: :)

darosior (commented):
Approach ACK, i think you are on the right track. A couple early comments below from a quick skim. You probably already intend to do that but just to be sure: please add a unit test exercising the various combinations for the `coins` method of the SQLite wrapper.

> Possibly use the updated functions elsewhere, e.g. list_spending_coins()

Oh, yes, that would be neat!

src/commands/mod.rs
jp1ac4: Ah yes, I'll include both confirmed and unconfirmed coins here.

src/database/mod.rs
jp1ac4: Oh yes I agree, we can remove the `Option` as the empty list achieves the same thing. I'll change that.

For the `listcoins` CLI args, should we still allow the user to omit the args completely or would we require them to pass an empty list?

If we allow to omit args, I don't think it's possible currently to pass only the second optional arg without the first, e.g.
```
listcoins --outpoints "[\"abcd:1\"]"
```
We would need to do
```
listcoins "[]" "[\"abcd:1\"]"
```

src/commands/mod.rs
jp1ac4: I've now updated this.

src/database/mod.rs
jp1ac4: I've now added a comment for each variant.

jp1ac4:
I've now made the following changes:
- added unit test for `coins()` method of SQLite wrapper
- updated functional test for `listcoins`
- removed `Option` from parameters (so empty list means no filter)
- consider both confirmed and unconfirmed coins for sweeping
- use `coins()` method in `list_spending_coins()`
- updated docstring of `spend_coins` to refer to "spending" instead of "spent"
- updated `listcoins` entry in API doc

The only doubt I have is whether it should be possible to omit the command-line args or if an empty list should be passed explicitly.

darosior:
Any reason for this to still be marked as WIP / draft?

jp1ac4:
> Any reason for this to still be marked as WIP / draft?

Hmm, not really :smile: I have one question about optional command-line args, but otherwise it's ready for review. Will change that now.

darosior (commented):
Ok from skimming through the diff here it looks good. Excellent test coverage. I'll play with it a bit locally now and investigate wtf is wrong again with our CI.

src/database/sqlite/mod.rs
```diff
@@ -889,6 +912,315 @@ CREATE TABLE spend_transactions (
...
+                conn.coins(&[], &[outpoint_b]),
+                conn.coins(&[], &[outpoint_c]),
+                conn.coins(&[], &[outpoint_d]),
+            ]
+            .iter()
+            .map(|c| c.len())
+            .all(|length| length == 1));
+
+            fs::remove_dir_all(tmp_dir).unwrap();
+        }
```
darosior, line 1221: This ought to be outside the scope of the connection object.

```suggestion
        }
    fs::remove_dir_all(tmp_dir).unwrap();
```

That'll fix the Windows CI.

darosior:
> The only doubt I have is whether it should be possible to omit the command-line args or if an empty list should be passed explicitly.

It should be possible to omit them, i think. And passing no argument or an empty list should have the same behaviour.

src/database/sqlite/mod.rs
jp1ac4: Thanks for this! The failing Windows test is now passing :relieved:

src/database/sqlite/mod.rs
```diff
@@ -889,6 +912,315 @@ CREATE TABLE spend_transactions (
...
+                conn.coins(&[CoinStatus::Unconfirmed], &[outpoint_b, outpoint_c]),
+                conn.coins(&[], &[outpoint_b, outpoint_c]),
+                conn.db_coins(&[outpoint_b, outpoint_c]),
+            ]
+            .iter()
+            .all(|coin| coin.len() == 2
+                && coin[0].outpoint == coin_b.outpoint
+                && coin[1].outpoint == coin_c.outpoint));
+
+            // Now spend second coin, even though it is still unconfirmed.
```
darosior, line 1169: Nice.

src/jsonrpc/api.rs
```diff
@@ -87,6 +87,41 @@ fn broadcast_spend(control: &DaemonControl, params: Params) -> Result<serde_json
     Ok(serde_json::json!({}))
 }
 
+fn list_coins(control: &DaemonControl, params: Option<Params>) -> Result<serde_json::Value, Error> {
+    let statuses_arg = params
+        .clone()
+        .and_then(|p| p.get(0, "statuses").cloned())
+        .and_then(|statuses| statuses.as_array().cloned());
```
darosior, line 94: That's a lot of `clone()`s. You don't need to allocate here since you just go through it.

```diff
diff --git a/src/jsonrpc/api.rs b/src/jsonrpc/api.rs
index 1432428..f2b2f4b 100644
--- a/src/jsonrpc/api.rs
+++ b/src/jsonrpc/api.rs
@@ -89,9 +89,9 @@ fn broadcast_spend(control: &DaemonControl, params: Params) -> Result<serde_json
 
 fn list_coins(control: &DaemonControl, params: Option<Params>) -> Result<serde_json::Value, Error> {
     let statuses_arg = params
-        .clone()
-        .and_then(|p| p.get(0, "statuses").cloned())
-        .and_then(|statuses| statuses.as_array().cloned());
+        .as_ref()
+        .and_then(|p| p.get(0, "statuses"))
+        .and_then(|statuses| statuses.as_array());
     let mut statuses: Vec<CoinStatus> = vec![];
     if let Some(statuses_arg) = statuses_arg {
         for status_arg in statuses_arg.iter() {
@@ -104,8 +104,9 @@ fn list_coins(control: &DaemonControl, params: Option<Params>) -> Result<serde_j
         }
     }
     let outpoints = params
-        .and_then(|p| p.get(1, "outpoints").cloned())
-        .and_then(|op| op.as_array().cloned())
+        .as_ref()
+        .and_then(|p| p.get(1, "outpoints"))
+        .and_then(|op| op.as_array())
         .and_then(|arr| {
             arr.iter()
                 .map(|entry| {
```

src/jsonrpc/api.rs
```diff
@@ -87,6 +87,41 @@ fn broadcast_spend(control: &DaemonControl, params: Params) -> Result<serde_json
...
+        .and_then(|p| p.get(1, "outpoints").cloned())
+        .and_then(|op| op.as_array().cloned())
+        .and_then(|arr| {
+            arr.iter()
+                .map(|entry| {
+                    entry
+                        .as_str()
+                        .and_then(|e| bitcoin::OutPoint::from_str(e).ok())
+                })
+                .collect::<Option<Vec<bitcoin::OutPoint>>>()
```
darosior, line 116: We shouldn't be silently ignoring these errors.

tests/test_rpc.py
```diff
@@ -73,6 +101,190 @@ def test_listcoins(lianad, bitcoind):
...
+            )["coins"],
+            key=lambda c: c["outpoint"],
+        )
+        == sorted(
+            lianad.rpc.listcoins(
+                ["spending", "spent"], [outpoint_a, outpoint_b, outpoint_c, outpoint_d]
+            )["coins"],
+            key=lambda c: c["outpoint"],
+        )
+    )
```
darosior, line 287: Great functional test too. A possible good addition is to sanity check the errors returned in case of invalid inputs. For instance make sure we catch an invalid status or an invalid outpoint. This would have caught the fact we actually ignore the error on malformed outpoints as i pointed above.

Example of such a sanity check:
https://github.com/wizardsardine/liana/blob/38d0fafac2b3cbdba57070173dc0136e20fd7019/tests/test_rpc.py#L610-L612

darosior (commented):
Looks great. Again, very nice test coverage. The only thing i really care about here is to not silently ignore malformed outpoints in the JSONRPC request.

src/jsonrpc/api.rs
jp1ac4: Thanks for this :pray: Have updated it now.

src/jsonrpc/api.rs
jp1ac4: Ah yes, I've updated this now.

tests/test_rpc.py
jp1ac4: That's a good idea thanks. I've added some error checks to the test.

src/jsonrpc/api.rs
```diff
@@ -87,6 +87,58 @@ fn broadcast_spend(control: &DaemonControl, params: Params) -> Result<serde_json
...
+                        bitcoin::OutPoint::from_str(op).map_err(|_| {
+                            Error::invalid_params(format!(
+                                "Invalid value {} in 'outpoints' parameter.",
+                                op_arg
+                            ))
+                        })
+                    })??,
+            );
+        }
+    }
```
jp1ac4, line 137: There might be a better way of writing this, but this was the best I could get working :blush:

jp1ac4:
Thanks for your feedback. I've now made the suggested changes.

src/jsonrpc/api.rs
jp1ac4: I think I've found a better way :nerd_face: Will test it and then commit shortly.

jp1ac4: I've now updated this.

jp1ac4:
I've just committed an improved way of getting the arguments from command line.

I also fixed a typo in the unit test name: `db_coins_()` -> `db_coins()`.

I don't have any further changes planned.

darosior:
ACK 36e04edc8207df2b3b3c9fac78f11700514ee2bd

Issue #611, opened 2023-08-15
"Mandatory signatory" in multisigs


darosior:
Since #53, we've implemented support (only) for simple k-of-N multisigs. We decided to not introduce too much complexity by throwing more complex multisigs into the installer, since people would still be used to legacy multisigs for a while.

As it turns out, we've been requested multiple times already if it was possible to have a mandatory signatory in a multisig. We should add support for it, but i'm reluctant to clutter the UI (at least until we've got some sort of "advanced view").

There are different paths we can take:
- Implement the possibility for having any kind of multisigs Miniscript permits (thresholds of thresholds, basically)
- Implementing a very restricted "mandatory signer" feature (like select only one or maybe several mandatory signers in the multisig)

Thoughts?

pythcoiner:
I'm in favor of the first option, keeping UI simple but letting advanced user import more complex descriptors.

Rob1Ham:
We call this feature a "key hierarchy" - which under the hood  takes the selected key(s) which are required as their own N of N multisig plus an and statement for the remaining keys. So a 2 of 3 with a "required key" is a single sig, plus a 1 of 2 multisig. If you were to just make this threshold of thresholds, you'd be able to achieve that functionality without having to add extra complications to the UX.

PR #706, opened 2023-09-08
Add bitbox support

edouardparis:
- Add bitbox02 as possible signer to liana-gui
- Introduce global_settings.json in datadir to keep the bitbox noise config.


darosior:
Just noting as a reminder i want to have a look at the spend info analysis on the daemon side and why it could crash, before we merge this.

edouardparis:
## The problem (see then solution done at the end)

I am gonna illustrate with an example.
Here a decarying multisig, with bitbox and a ledger, primary path is a 3 of 3 with ledger and two times the same bitbox.
```
wsh(or_i(and_v(v:pkh([75897fd4/48'/1'/0'/2']tpubDEtwaiemyuvA8UXisdZFR4j1y6aUhxQ8wf9Tf2wYWw1cooP1PHdA7uAUK6H53vZkBvF84ibChB7Wpgd1tkPuuZMVJpRp1iLx9h3AyqefUGh/<6;7>/*),older(3)),or_d(multi(3,[ffd63c8d/48'/1'/0'/2']tpubDExA3EC3iAsPxPhFn4j6gMiVup6V2eH3qKyk69RcTc9TTNRfFYVPad8bJD5FCHVQxyBT4izKsvr7Btd2R4xmQ1hZkvsqGBaeE82J71uTK4N/<0;1>/*,[75897fd4/48'/1'/0'/2']tpubDEtwaiemyuvA8UXisdZFR4j1y6aUhxQ8wf9Tf2wYWw1cooP1PHdA7uAUK6H53vZkBvF84ibChB7Wpgd1tkPuuZMVJpRp1iLx9h3AyqefUGh/<0;1>/*,[75897fd4/48'/1'/0'/2']tpubDEtwaiemyuvA8UXisdZFR4j1y6aUhxQ8wf9Tf2wYWw1cooP1PHdA7uAUK6H53vZkBvF84ibChB7Wpgd1tkPuuZMVJpRp1iLx9h3AyqefUGh/<2;3>/*),and_v(v:thresh(2,pkh([ffd63c8d/48'/1'/0'/2']tpubDExA3EC3iAsPxPhFn4j6gMiVup6V2eH3qKyk69RcTc9TTNRfFYVPad8bJD5FCHVQxyBT4izKsvr7Btd2R4xmQ1hZkvsqGBaeE82J71uTK4N/<2;3>/*),a:pkh([75897fd4/48'/1'/0'/2']tpubDEtwaiemyuvA8UXisdZFR4j1y6aUhxQ8wf9Tf2wYWw1cooP1PHdA7uAUK6H53vZkBvF84ibChB7Wpgd1tkPuuZMVJpRp1iLx9h3AyqefUGh/<4;5>/*)),older(2)))))#a4gxxd5y
```

```
wsh(or_i(
  and_v(v:pkh([75897fd4/48'/1'/0'/2']bitbox/<6;7>/*),older(3)),
  or_d(multi(3,[ffd63c8d/48'/1'/0'/2']ledger/<0;1>/*,[75897fd4/48'/1'/0'/2']bitbox/<0;1>/*,[75897fd4/48'/1'/0'/2']bitbox/<2;3>/*),
  and_v(v:thresh(2,pkh([ffd63c8d/48'/1'/0'/2']ledger/<2;3>/*),a:pkh([75897fd4/48'/1'/0'/2']bitbox/<4;5>/*)),older(2)))))
```
By creating a recovery for two coins using the last unlocked recovery path, gui provides us three choices:

* 3 more signatures from `Bitbox` and `Ledger`
* 2 more signatures from `Bitbox` and `Ledger`
* 1 more signature from `Bitbox`

(I choose the recovery example because with the multiple path it shows the confusion)

The bip32_derivations (BTreeMap field in the rust-bitcoin PSBT struct) looks like:

**Input  0**
index | pubkey | fingerprint | path | usage |
| ----- | --------- | ---------- | ----- | --- |
| `#0` | 028a1b3260 ... | 75897fd4 | m/48'/1'/0'/2'/0/1 | primary path |
| `#1` | 0293e5f0e65 ... | 75897fd4 | m/48'/1'/0'/2'/4/1 | recovery path older(2) |
| `#2` | 02f4e7c25e5f ...| ffd63c8d | m/48'/1'/0'/2'/0/1 | primary path |
| `#3` |0300b8f8792 ... | 75897fd4 | m/48'/1'/0'/2'/2/1 | primary path |
|`#4` | 036c044cc10a ...  | ffd63c8d | m/48'/1'/0'/2'/2/1 | recovery path older(2) |
| `#5` |03aca8fce1a ... | 75897fd4 | m/48'/1'/0'/2'/6/1 |  recovery path older(3)

**Input  1**
| index | pubkey | fingerprint | path | usage |
| ---- | --------- | ---------- | ----- | ---- |
| `#0` | 02348438ade2b ... | 75897fd4 | m/48'/1'/0'/2'/2/0 | primary path |
| `#1` | 023eeb02ffd936 ... | 75897fd4 | m/48'/1'/0'/2'/0/0 | primary path |
| `#2` | 02f165c3d2976 ... | ffd63c8d | m/48'/1'/0'/2'/0/0 | primary path |
| `#3` | 02f8f64e4f177e ... | ffd63c8d | m/48'/1'/0'/2'/2/0 | recovery path older(2) |
| `#4` | 03317868e0ff02 ... | 75897fd4 | m/48'/1'/0'/2'/6/0 | recovery path older(3) |
| `#5` | 03ef54bdf4dcc5 ... | 75897fd4 | m/48'/1'/0'/2'/4/0 | recovery path older(2) |

**Bitbox will not sign for all bip32_derivation per inputs but the first one that match its master fingerprint** 
It sadly means that without modification on our side, bitbox signs on first round:

input | bip32_derivation index | pubkey | fingerprint | path | usage |
|------| ----- | --------- | ---------- | ----- | --- |
| `#0` | `#0` | 028a1b3260 ... | 75897fd4 | m/48'/1'/0'/2'/0/1 | primary path |
| `#1` | `#0` | 02348438ade2b ... | 75897fd4 | m/48'/1'/0'/2'/2/0 | primary path |

Then (by removing the bip32_derivation that have a partial_sigs that match the fingerprint, otherwise bitbox02 will sign again the same derivation), bitbox signs on second round

input | bip32_derivation index | pubkey | fingerprint | path | usage |
|------| ----- | --------- | ---------- | ----- | --- |
| `#0` | `#1` | 0293e5f0e65 ... | 75897fd4 | m/48'/1'/0'/2'/4/1 | recovery path older(2) |
| `#1` | `#1 `| 023eeb02ffd936 ... | 75897fd4 | m/48'/1'/0'/2'/0/0 | primary path |

input#0 is missing a sig for the primary path (instead, it received a sig for a recovery path) but input#1 has the two signatures required.

:warning: [Lianad does not like that](https://github.com/wizardsardine/liana/blob/3a59790530b81d8ec50719793e31d9d229827dbe/src/descriptors/mod.rs#L291), it expects that when an input is signed for a particular path, all the other inputs are also signed for the same path.

## The solution/What was done in async-hwi and this PR:

1. remove before signing all bip32_derivation that has pubkey that match the partial_sigs pubkeys (aka bip32_derivation already signed) and put them back after the signing process is done.
2. Order derivation paths the bip32_derivations of each inputs and remove before signing any bip32_derivations that have the same master_fingerprints and that is not first in the order, then put them back once the device signed.

The bip32_derivations become:

**Input  0**
previous index | pubkey | fingerprint | path | usage |
| ----- | --------- | ---------- | ----- | --- |
|`#4` | 036c044cc10a ...  | ffd63c8d | m/48'/1'/0'/2'/2/1 | recovery path older(2) |
| `#5` |03aca8fce1a ... | 75897fd4 | m/48'/1'/0'/2'/6/1 |  recovery path older(3)

**Input  1**
| previous index | pubkey | fingerprint | path | usage |
| ---- | --------- | ---------- | ----- | ---- |
| `#3` | 02f8f64e4f177e ... | ffd63c8d | m/48'/1'/0'/2'/2/0 | recovery path older(2) |
| `#4` | 03317868e0ff02 ... | 75897fd4 | m/48'/1'/0'/2'/6/0 | recovery path older(3) |

Bitbox will sign only for the recovery path older(3) for both inputs.
Ledger will only sign for the last path the device as a key in which means recovery path older(2).

So previously, for the recovery, bitbox was only able to sign first for the primary path (first in the bip32_derivations order),
Now it signs for the same path  for each input and by order desc of derivation it will mostly sign for the last recovery path available, **if and only if** the following hypothesis works:

**Every inputs that are from the liana setup have all the same sets of bip32_derivation with only the input last derivation index as difference.**
(which is not sure, user can open bip174.org, remove bip32_derivation and give it to an other user that will import it and liana will not complete the missing bip32_derivations. Then some inputs may have only bip32_derivation of recovery path and others only primary path, so it breaks again the wallet)

darosior:
Ok, i think i finally understand the ordering issue. Indeed this is quite convoluted. I'll try to sum this up, let me know if i'm missing anything.

Contrary to the Ledger Nano series, the Bitbox won't fill up a PSBT with signatures for all its pubkeys but only for the first one as ordered in the [`PSBT_IN_BIP32_DERIVATION` BIP174 field](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#user-content-Specification). On our end, when creating a transaction, we fill the PSBT with as much information from the descriptor as possible (as [advised in BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#user-content-Updater)). When in turn passing this PSBT to the Bitbox for signing this creates two issues:
1. If multiple keys from the Bitbox are used in the same spending path, signing a transaction will require multiple signing rounds.
2. If multiple keys from the Bitbox are used at all (i.e. even across spending paths), the Bitbox may return incompatible signatures across inputs (i.e. for different spending paths, whichever key appears first in the `PSBT_IN_BIP32_DERIVATION` mapping).

The first issue is concerning but arguably only reachable for pathological descriptors. Concerning because it means a user would be shown twice "Confirm sending `X` BTC" on their device's screen, which would mean they are sending `2 * X` BTC but actually they aren't. Normalizing this practice would train users to ignore what their signing device is telling them, which is pretty bad. In addition i don't think there is any usecase for repeating a key in a single spending path (aside from https://github.com/wizardsardine/liana/issues/611#issuecomment-1680876629 but that's due to a self-imposed limitation). Therefore if it opens up issues like this one, we should just prevent users from creating such descriptors in the first place.

The second issue is however reachable for perfectly reasonable descriptors. A decaying or expanding multisig is one such example. Here the Bitbox is arguably not respecting the PSBT flow by artificially limiting how many signatures it provides (see for instance how the [BIP174 "Simple Signer Algorithm"](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#user-content-Simple_Signer_Algorithm) signs for all keys it can). By doing so it forces us to withhold some information from the PSBT to direct it at signing what we're interested in. This:
- introduces unnecessary complexity on our side;
- is confusing in itself to only partially approve a transaction when a user approves it entirely on their device (such as they may have to approve it entirely more than once to effectively only approve it once);
- may rule out some (edgy) use cases: when you sign for a latter-available spending path it's useful to sign for all previously-available spending paths. (The opposite isn't possible since the `nSequence` / `nLockTime` is committed to by the signature.)

darosior:
So instead of creating PSBTs with all the information and then having the GUI massage it to remove keys for other spending paths, how about we simply create the PSBT with only the information for a given spending path in the first place?

I think this would solve all the issues in a much less error-prone manner. Would it break any existing use though?

pythcoiner:
> So instead of creating PSBTs with all the information and then having the GUI massage it to remove keys for other spending paths, how about we simply create the PSBT with only the information for a given spending path in the first place?
> 
> I think this would solve all the issues in a much less error-prone manner. Would it break any existing use though?

looks great to me (until bitbox fix this issue), @edouardparis did you open an issue about that on the bitbox repo?

darosior:
I've also heard Coldcard has the same behaviour so this would preemptively patch it for supporting them as well.

darosior:
Creating PSBTs for single spending paths: https://github.com/wizardsardine/liana/pull/722.

darosior:
The other part is here: https://github.com/wizardsardine/liana/pull/723.

Once both #722 and #723 land and are integrated in the GUI, we can move forward with (a much simplified version of) this PR.

edouardparis:
rebased on master, for fixing multiple conflicts.

darosior (commented):
~~There is no need to filter the bip32 derivs (31ec3c9f71460c38628c7a39ee22544a70c91f11) anymore?~~

My bad, looks like it's reverted in a later commit.. 8bce12183f76b960da783307fdd26228d925a49a

edouardparis:
rebased and squashed all fix commits.

edouardparis:
updated liana version and rebased

edouardparis (commented):
Self-ACK 4be74ad496692d371adffcc1b3c868ef6a01cb52

PR #709, opened 2023-09-09
Implement listadresses

pythcoiner:
address #681 

todo:
- [x]  implement tests
- [x] update docs

edit: i'm really new to rust, don't hesitate to kick my ass when i write stupid code

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -7,9 +7,10 @@ Commands must be sent as valid JSONRPC 2.0 requests, ending with a `\n`.
 
 | Command                                                     | Description                                                   |
 | ----------------------------------------------------------- | ----------------------------------------------------          |
-| [`stop`](#stop)                                             | Stops liana daemon                                     |
+| [`stop`](#stop)                                             | Stops liana daemon                                            |
 | [`getinfo`](#getinfo)                                       | Get general information about the daemon                      |
 | [`getnewaddress`](#getnewaddress)                           | Get a new receiving address                                   |
+| [`listaddresses`](#listaddresses)                           | List addresses given start_index and count                     |
 | [`listcoins`](#listcoins)                                   | List all wallet transaction outputs.                          |
 | [`createspend`](#createspend)                               | Create a new Spend transaction                                |
 | [`updatespend`](#updatespend)                               | Store a created Spend transaction                             |
@@ -77,6 +78,28 @@ This command does not take any parameter for now.
 | Field         | Type   | Description        |
 | ------------- | ------ | ------------------ |
 | `address`     | string | A Bitcoin address  |
+
+
+### `listaddresses`
+
+List receive and change addresses given start_index and count. Both arguments are optional.
+Default value for `start_index` is 0.
+If no value is passed for `count` the maximum generated index between receive and change is selected.
+
+#### Request
+
+| Field         | Type              | Description                                                 |
+| ------------- | ----------------- | ----------------------------------------------------------- |
+| `start_index` | integer(optional) | Index of the first address to list                          |
+| `count`       | integer(optional) | Number of addresses to list                                 |
+
+#### Response
+
+| Field         | Type              | Description                                                 |
+| ------------- | ----------------- | ----------------------------------------------------------- |
+| `index`       | integer           | Derivation index                                            |
+| `receive`     | string            | Receive address                                             |
+| `change`      | string            | Change address                                              |
 
 
 ### `listcoins`
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -25,7 +25,7 @@ use std::{
 
 use miniscript::{
     bitcoin::{
-        self, address,
+        self, address, bip32,
         locktime::absolute,
         psbt::{Input as PsbtIn, Output as PsbtOut, PartiallySignedTransaction as Psbt},
     },
@@ -72,6 +72,8 @@ pub enum CommandError {
     /// An error that might occur in the racy rescan triggering logic.
     RescanTrigger(String),
     RecoveryNotAvailable,
+    InvalidAddressCount,
+    InvalidAddressIndex,
 }
 
 impl fmt::Display for CommandError {
@@ -136,6 +138,8 @@ impl fmt::Display for CommandError {
                 f,
                 "No coin currently spendable through this timelocked recovery path."
            ),
+            Self::InvalidAddressCount => write!(f, "Invalid address count, should be under 2^31-1"),
+            Self::InvalidAddressIndex => write!(f, "Invalid address index, should be under 2^31-1"),
         }
     }
 }
@@ -287,6 +291,69 @@ impl DaemonControl {
             .derive(index, &self.secp)
             .address(self.config.bitcoin_config.network);
         GetAddressResult::new(address)
+    }
+
+    /// list addresses
+    pub fn list_addresses(
+        &self,
+        start_index: Option<u32>,
+        count: Option<u32>,
+    ) -> Result<ListAddressesResult, CommandError> {
+        let mut db_conn = self.db.connection();
+        let receive_index: u32 = db_conn.receive_index().into();
+        let change_index: u32 = db_conn.change_index().into();
+
+        let start_index = start_index.unwrap_or(0);
+
+        if start_index > (2u32.pow(31) - 1) {
+            return Err(CommandError::InvalidAddressIndex);
+        }
+
+        let count = count.unwrap_or_else(|| receive_index.max(change_index) - start_index);
+
+        if count == 0 {
+            let out: Vec<AddressInfo> = Vec::new();
+            return Ok(ListAddressesResult::new(out));
+        }
+
+        let index = start_index
+            .checked_add(count)
+            .and_then(|index| index.checked_sub(1))
+            .and_then(|index| {
+                if index > (2u32.pow(31) - 1) {
+                    None
+                } else {
+                    Some(index)
+                }
+            })
+            .ok_or(CommandError::InvalidAddressCount)?;
+
+        let addresses: Vec<AddressInfo> = (start_index..=index)
+            .map(|index| {
+                let child = bip32::ChildNumber::from_normal_idx(index).expect("Cannot fail here");
+
+                let receive = self
+                    .config
+                    .main_descriptor
+                    .receive_descriptor()
+                    .derive(child, &self.secp)
+                    .address(self.config.bitcoin_config.network);
+
+                let change = self
+                    .config
+                    .main_descriptor
+                    .change_descriptor()
+                    .derive(child, &self.secp)
+                    .address(self.config.bitcoin_config.network);
+
+                AddressInfo {
+                    index,
+                    receive,
+                    change,
+                }
+            })
+            .collect();
+        Ok(ListAddressesResult::new(addresses))
     }
 
     /// Get a list of all known coins, optionally by status and/or outpoint.
@@ -860,6 +927,24 @@ impl GetAddressResult {
     }
 }
 
+#[derive(Debug, Clone, Serialize)]
+pub struct AddressInfo {
+    index: u32,
+    receive: bitcoin::Address,
+    change: bitcoin::Address,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct ListAddressesResult {
+    addresses: Vec<AddressInfo>,
+}
+
+impl ListAddressesResult {
+    pub fn new(addresses: Vec<AddressInfo>) -> Self {
+        ListAddressesResult { addresses }
+    }
+}
+
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct LCSpendInfo {
     pub txid: bitcoin::Txid,
@@ -968,6 +1053,55 @@ mod tests {
         // We won't get the same twice.
         let addr2 = control.get_new_address().address;
         assert_ne!(addr, addr2);
+
+        ms.shutdown();
+    }
+
+    #[test]
+    fn listaddresses() {
+        let ms = DummyLiana::new(DummyBitcoind::new(), DummyDatabase::new());
+
+        let control = &ms.handle.control;
+
+        let list = control.list_addresses(Some(2), Some(5)).unwrap();
+
+        assert_eq!(list.addresses[0].index, 2);
+        assert_eq!(list.addresses.last().unwrap().index, 6);
+
+        let addr0 = control.get_new_address().address;
+        let addr1 = control.get_new_address().address;
+        let _addr2 = control.get_new_address().address;
+        let addr3 = control.get_new_address().address;
+        let addr4 = control.get_new_address().address;
+
+        let list = control.list_addresses(Some(0), None).unwrap();
+
+        assert_eq!(list.addresses[0].index, 0);
+        assert_eq!(list.addresses[0].receive, addr0);
+        assert_eq!(list.addresses.last().unwrap().index, 4);
+        assert_eq!(list.addresses.last().unwrap().receive, addr4);
+
+        let list = control.list_addresses(None, None).unwrap();
+
+        assert_eq!(list.addresses[0].index, 0);
+        assert_eq!(list.addresses[0].receive, addr0);
+        assert_eq!(list.addresses.last().unwrap().index, 4);
+        assert_eq!(list.addresses.last().unwrap().receive, addr4);
+
+        let list = control.list_addresses(Some(1), Some(3)).unwrap();
+
+        assert_eq!(list.addresses[0].index, 1);
+        assert_eq!(list.addresses[0].receive, addr1);
+        assert_eq!(list.addresses.last().unwrap().index, 3);
+        assert_eq!(list.addresses.last().unwrap().receive, addr3);
+
+        let addr5 = control.get_new_address().address;
+        let list = control.list_addresses(Some(5), None).unwrap();
+
+        assert_eq!(list.addresses[0].index, 5);
+        assert_eq!(list.addresses[0].receive, addr5);
+        assert_eq!(list.addresses.last().unwrap().index, 5);
+        assert_eq!(list.addresses.last().unwrap().receive, addr5);
 
         ms.shutdown();
     }
--- a/src/jsonrpc/api.rs
+++ b/src/jsonrpc/api.rs
@@ -136,6 +136,26 @@ fn list_coins(control: &DaemonControl, params: Option<Params>) -> Result<serde_j
     Ok(serde_json::json!(&res))
 }
 
+fn list_addresses(
+    control: &DaemonControl,
+    params: Option<Params>,
+) -> Result<serde_json::Value, Error> {
+    let start_index: Option<u32> = params
+        .as_ref()
+        .and_then(|p| p.get(0, "start_index"))
+        .and_then(|i| i.as_u64())
+        .and_then(|i| i.try_into().ok());
+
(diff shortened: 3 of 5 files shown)


src/commands/mod.rs
```diff
@@ -289,6 +291,48 @@ impl DaemonControl {
         GetAddressResult::new(address)
     }
 
+    /// list addresses
+    pub fn list_addresses(&self, start_index: u32, count: u32) -> GetAddressesResult {
```
pythcoiner, line 295: not sure if start_index should be u32 or ChildNumber type

src/commands/mod.rs
```diff
@@ -853,6 +897,24 @@ impl GetAddressResult {
     }
 }
 
+#[derive(Debug, Clone, Serialize)]
+pub struct AddressInfo {
```
pythcoiner, line 901: not sure about wording of 'AddressInfo'

tests/test_rpc.py
```diff
@@ -36,6 +36,16 @@ def test_getaddress(lianad):
     assert res["address"] != lianad.rpc.getnewaddress()["address"]
 
 
+def test_listadresses(lianad):
+    list = lianad.rpc.listaddresses(2, 5)
```
pythcoiner, line 40: something weird here, if i call `lianad.rpc.listaddresses(0, 0)` the test will be stuck a this point (waiting indefinitely for a lianad answer):

```
tests/test_rpc.py::test_listadresses 
--------------------------------------------------------------------------------------------------------- live log setup ----------------------------------------------------------------------------------------------------------
DEBUG    root:utils.py:271 Starting 'bitcoind -datadir=/tmp/lianad-tests-bzxdc1cw/test_listadresses_1/bitcoind -printtoconsole -server -debug=1 -debugexclude=libevent -debugexclude=tor'
DEBUG    root:utils.py:357 Waiting for ['Done loading'] in the logs
DEBUG    root:utils.py:381 Found 're.compile('Done loading')' in logs
INFO     root:bitcoind.py:87 Bitcoind started
DEBUG    root:utils.py:271 Starting '/home/pyth/rust/liana_pythcoiner/tests/test_framework/../../target/debug/lianad --conf /tmp/lianad-tests-bzxdc1cw/test_listadresses_1/lianad/config.toml'
DEBUG    root:utils.py:357 Waiting for ['Database initialized and checked', 'JSONRPC server started.'] in the logs
DEBUG    root:utils.py:381 Found 're.compile('Database initialized and checked')' in logs
DEBUG    root:utils.py:381 Found 're.compile('JSONRPC server started.')' in logs
---------------------------------------------------------------------------------------------------------- live log call ----------------------------------------------------------------------------------------------------------
DEBUG    root:utils.py:208 Calling listaddresses with params (0, 0)
```
the issue seems to be `count=0`

i also try to run a lianad daemon + send `listaddresses 0 0 ` with liana-cli or [#2](https://github.com/wizardsardine/liana-rpc/pull/2) of [liana-rpc](https://github.com/wizardsardine/liana-rpc) python bindings, everything goes well, got the right answer and no panic on lianad side

src/commands/mod.rs
```diff
@@ -33,6 +33,8 @@ use miniscript::{
 };
 use serde::{Deserialize, Serialize};
 
+use bitcoin::bip32;
```
darosior, line 36: nit: could you include it as part of the brackets above?

src/commands/mod.rs
```diff
@@ -289,6 +291,48 @@ impl DaemonControl {
         GetAddressResult::new(address)
     }
 
+    /// list addresses
+    pub fn list_addresses(&self, start_index: u32, count: u32) -> GetAddressesResult {
+        let mut db_conn = self.db.connection();
+        let receive_index: u32 = db_conn.receive_index().into();
+        let change_index: u32 = db_conn.receive_index().into();
```
darosior, line 298: This should call `change_index()`?

src/commands/mod.rs
darosior: Since `count` needs to be `u32` i think it's better to keep `start_index` as a `u32` too.

src/commands/mod.rs
darosior: Looks ok.

src/commands/mod.rs
```diff
@@ -853,6 +897,24 @@ impl GetAddressResult {
...
 
+#[derive(Debug, Clone, Serialize)]
+pub struct AddressInfo {
+    index: u32,
+    receive: bitcoin::Address,
+    change: bitcoin::Address,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct GetAddressesResult {
```
darosior, line 908: nit: `ListAddressesResult`

src/commands/mod.rs
darosior, line 295: To make those parameters optional it's better to have them as `Option`s rather than using a magic "0" value.

src/commands/mod.rs
```diff
@@ -301,7 +301,7 @@ impl DaemonControl {
         let index = if count != 0 {
             start_index + count - 1
         } else {
-            max_index
+            max_index - 1
```
darosior, line 304: What if `max_index` is 0? :)

tests/test_rpc.py
darosior: I think that's because you trigger an underflow with `max_index - 1` (see my other comment above).

darosior (commented):
Approach ACK. Needs a bit more testing (and an entry in `doc/API.md`) but it's in good shape!

src/commands/mod.rs
pythcoiner: absolutely!

src/commands/mod.rs
pythcoiner: do you also want `count` to be optionnal in rpc call?

darosior: I think both should be. `start_height` should default to `0` and `count` to the max of the change and receive indexes.

src/commands/mod.rs
```diff
@@ -289,6 +289,70 @@ impl DaemonControl {
...
+                };
+            }
+            Some(c) => {
+                index = if (start_index + c) > 0 {
+                    start_index + c - 1
+                } else {
+                    0
+                }
+            }
+        }
```
pythcoiner, line 325: not sure if a match is the better way to handle this ?

jp1ac4: Perhaps as you are only interested in `Some(c)` or `None` you could use `if let Some(c) = count {} else {}`.

Nit: for the subtraction logic, you could try [saturating_sub](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub) instead of the `if/else`.

pythcoiner: thx a lot!

darosior:
Let me know when this is ready for review by moving this out of draft / WIP. No rush!

pythcoiner:
> Let me know when this is ready for review by moving this out of draft / WIP. No rush!

i think it is now ;)

doc/API.md
```diff
@@ -7,9 +7,10 @@ Commands must be sent as valid JSONRPC 2.0 requests, ending with a `\n`.
 
 | Command                                                     | Description                                                   |
 | ----------------------------------------------------------- | ----------------------------------------------------          |
-| [`stop`](#stop)                                             | Stops liana daemon                                     |
+| [`stop`](#stop)                                             | Stops liana daemon                                            |
 | [`getinfo`](#getinfo)                                       | Get general information about the daemon                      |
 | [`getnewaddress`](#getnewaddress)                           | Get a new receiving address                                   |
+| [`listaddresses`](#listaddresses)                           | List adresses given start_index and count                     |
```
jp1ac4, line 13: Nit: "adresses" -> "addresses"

doc/API.md
```diff
@@ -79,6 +80,28 @@ This command does not take any parameter for now.
 | `address`     | string | A Bitcoin address  |
 
 
+### `listaddresses`
+
+List receive and change adresses given start_index and count. Both arguments are optional.
```
jp1ac4, line 85: Nit: "adresses" -> "addresses"

doc/API.md
```diff
@@ -79,6 +80,28 @@ This command does not take any parameter for now.
...
+Default value for `start_index` is 0.
+If no value is pass for `count` the maximum generated index between receive and change is selected.
```
jp1ac4, line 87: Nit: "pass" -> "passed"

doc/API.md
```diff
@@ -79,6 +80,28 @@ This command does not take any parameter for now.
...
+
+#### Request
+
+| Field         | Type              | Description                                                 |
+| ------------- | ----------------- | ----------------------------------------------------------- |
+| `start_index` | integer(optional) | Index of the first address to list                          |
+| `count`       | integer(optional) | The number of adresses you want to list                     |
```
jp1ac4, line 94: Nit: "adresses" -> "addresses"

pythcoiner:
thx @jp1ac4  for your review, comments have been addressed in 4e0bc0a677ed17bdb83150aabb0b24da5e8278ee

tests/test_rpc.py
```diff
@@ -36,6 +36,24 @@ def test_getaddress(lianad):
     assert res["address"] != lianad.rpc.getnewaddress()["address"]
 
 
+def test_listadresses(lianad):
```
jp1ac4, line 39: I think you still need to apply [black](https://github.com/psf/black) formatting to this test.

src/jsonrpc/api.rs
```diff
@@ -258,6 +282,14 @@ pub fn handle_request(control: &DaemonControl, req: Request) -> Result<Response,
         }
         "getinfo" => serde_json::json!(&control.get_info()),
         "getnewaddress" => serde_json::json!(&control.get_new_address()),
+        "listaddresses" => {
+            let params = req.params.ok_or_else(|| {
```
jp1ac4, line 286: As the parameters are optional, I'm wondering if you could keep `params` as `Option<Params>` and pass that to `list_addresses`.

I'm able to run `listaddresses` without any parameters so I guess `req.params` is `Some` also in that case and it works fine, but not sure if there might be a situation where it is `None`.

@darosior may be able to help.

pythcoiner: you seems right, i wonder why `Error::invalid_params` is not triggered when call listaddresses w/o args? shouldn't it in that case?

tests/test_rpc.py
pythcoiner: yep, done ;)

darosior:
Could you rebase and [squash your commits into a couple ones](https://github.com/wizardsardine/liana/blob/master/CONTRIBUTING.md)? I'll review then.

pythcoiner:
I rebased and squash to 3 commits implement/test/fixes, feel free to tell me if you prefer I squash to a single one.

darosior:
Having tests in a separate commit is fine. However having fixups in a separate commit than the one where issues were introduced creates needless churn and confusion.

That's fine for this time, but next time please squash the fixup commit(s) in the appropriate original commit(s).

src/commands/mod.rs
```diff
@@ -289,6 +289,58 @@ impl DaemonControl {
...
+                start_index
+            } else {
+                max_index.saturating_sub(1)
+            }
+        };
+
+        let addresses: Vec<AddressInfo> = (start_index..=index)
+            .map(|index| {
+                let child = bip32::ChildNumber::from_normal_idx(index)
+                    .expect("Failed to convert index to ChildNumber");
```
darosior, line 318: Could we error instead of crashing if a value greater than `2**31 - 1` is passed for count? It's an insane argument to give to the command, but it's nicer to never crash on insane inputs.

Let's add an `InvalidAddressCount` variant to the `CommandError` enum and use it here and above?

src/commands/mod.rs
```diff
@@ -289,6 +289,58 @@ impl DaemonControl {
...
+        count: Option<u32>,
+    ) -> ListAddressesResult {
+        let mut db_conn = self.db.connection();
+        let receive_index: u32 = db_conn.receive_index().into();
+        let change_index: u32 = db_conn.change_index().into();
+
+        let start_index = start_index.unwrap_or(0);
+
+        let index = if let Some(c) = count {
+            (start_index + c).saturating_sub(1)
```
darosior, line 305: No. We shouldn't silently roll over. We should explicitly error if the start_index and/or count are so large that this goes above `u32::MAX`.

I would use checked_[add/sub] and return an error if it returns None. Something like:
```suggestion
            start_index.checked_add(c).and_then(|index| index.checked_sub(1)).flatten()
```

src/commands/mod.rs
```diff
@@ -289,6 +289,58 @@ impl DaemonControl {
...
+        } else {
+            let max_index = receive_index.max(change_index);
+            if start_index >= max_index {
+                start_index
+            } else {
+                max_index.saturating_sub(1)
```
darosior, line 311: Same here. Use `checked_sub` and error if it returns None instead of rolling over.

src/jsonrpc/api.rs
darosior: That's most likely you tested with the CLI which always includes a `params` entry in the JSONRPC request. But that's not mandated by the [JSONRPC2 specs](https://www.jsonrpc.org/specification#request_object).

I agree with @jp1ac4 this shouldn't error if params is None. You can take inspiration from how `list_coins` does this.

src/commands/mod.rs
darosior: Actually why do you substract `1` at all here? This is already the max index, not a count.

I think this can simply be `std::max(start_index, max_index)`.

tests/test_rpc.py
```diff
@@ -37,6 +37,24 @@ def test_getaddress(lianad):
...
+    assert "addresses" in list
+    addr = list["addresses"]
+    assert addr[0]["index"] == 2
+    assert addr[-1]["index"] == 6
+
+    list3 = lianad.rpc.listaddresses()
+    _ = lianad.rpc.getnewaddress()
+    _ = lianad.rpc.getnewaddress()
+    list4 = lianad.rpc.listaddresses()
+    assert len(list4["addresses"]) == len(list3["addresses"]) + 1
```
darosior, line 53: May be worth adding some comments here to make it easier to future readers (who may well be ourselves!).
```suggestion
    list3 = lianad.rpc.listaddresses()  # start_index = 0, receive_index = 0
    _ = lianad.rpc.getnewaddress()  # start_index = 0, receive_index = 1
    _ = lianad.rpc.getnewaddress()  # start_index = 0, receive_index = 2
    list4 = lianad.rpc.listaddresses()
    assert len(list4["addresses"]) == len(list3["addresses"]) + 1
```

Coupled with asserting the size of the list this would have made the off by one error from above evident:
```suggestion
    list3 = lianad.rpc.listaddresses()
    _ = lianad.rpc.getnewaddress()  # start_index = 0, receive_index = 1
    _ = lianad.rpc.getnewaddress()  # start_index = 0, receive_index = 2
    list4 = lianad.rpc.listaddresses()
    assert len(list4["addresses"]) == len(list3["addresses"]) + 2 == 3
```

darosior (commented):
Almost there, the main things are:
- There is an off by one when `count` is not provided
- Let's not crash on insane inputs
- Let's not error if a `params` entry isn't given in the JSONRPC request, since all fields are optional

pythcoiner:
comment have been adressed

darosior:
This comment wasn't addressed? https://github.com/wizardsardine/liana/pull/709#discussion_r1324600856

pythcoiner:
> This comment wasn't addressed? [#709 (comment)](https://github.com/wizardsardine/liana/pull/709#discussion_r1324600856)

right, it's done now

src/commands/mod.rs
```diff
@@ -136,6 +138,8 @@ impl fmt::Display for CommandError {
                 f,
                 "No coin currently spendable through this timelocked recovery path."
            ),
+            Self::InvalidAddressCount => write!(f, "Invalid address count, should be under 2^31-1"),
+            Self::InvalidAddressIndex => write!(f, "Invalid address index, should be under 2^31-1"),
```
darosior, line 142: Would be better to introduce a single enum variant, a general "invalid derivation index" which could be reused in the future.

src/commands/mod.rs
```diff
@@ -289,6 +293,69 @@ impl DaemonControl {
...
+    ) -> Result<ListAddressesResult, CommandError> {
+        let mut db_conn = self.db.connection();
+        let receive_index: u32 = db_conn.receive_index().into();
+        let change_index: u32 = db_conn.change_index().into();
+
+        let start_index = start_index.unwrap_or(0);
+
+        if start_index > (2u32.pow(31) - 1) {
+            return Err(CommandError::InvalidAddressIndex);
+        }
```
darosior, line 310: (Here and below) Ok, this works. But it would be much clearer to use `ChildNumber::from_normal_idx`'s error or at the very least add a comment.

src/commands/mod.rs
```diff
@@ -289,6 +293,69 @@ impl DaemonControl {
...
+
+        let count = count.unwrap_or_else(|| receive_index.max(change_index) - start_index);
```
darosior, line 312: What if `start_index` is larger than the maximum of the receive and change indexes? We need to really be careful about underflows in this code.

src/commands/mod.rs
```diff
@@ -289,6 +293,69 @@ impl DaemonControl {
...
+
+        if count == 0 {
+            let out: Vec<AddressInfo> = Vec::new();
+            return Ok(ListAddressesResult::new(out));
+        }
```
darosior, line 317: It's a bit ugly to special case 0 here to avoid the `checked_sub` below to return `None`. Could get rid of the necessity for a checked_sub altogether by using an exclusive range (`(x..y)`) below instead of an inclusive one (`(x..=y)`).

src/commands/mod.rs
```diff
@@ -860,6 +927,24 @@ impl GetAddressResult {
     }
 }
 
+#[derive(Debug, Clone, Serialize)]
```
darosior, line 930: Would be nice to make it PartialEq, Eq too.

src/commands/mod.rs
```diff
@@ -972,6 +1057,55 @@ mod tests {
         ms.shutdown();
     }
 
+    #[test]
+    fn listaddresses() {
```
darosior, line 1061: Would be good to check bounds here for hardened derivation index, overflows and underflows.

src/jsonrpc/api.rs
```diff
@@ -136,6 +136,26 @@ fn list_coins(control: &DaemonControl, params: Option<Params>) -> Result<serde_j
...
 
+fn list_addresses(
+    control: &DaemonControl,
+    params: Option<Params>,
+) -> Result<serde_json::Value, Error> {
+    let start_index: Option<u32> = params
+        .as_ref()
+        .and_then(|p| p.get(0, "start_index"))
+        .and_then(|i| i.as_u64())
+        .and_then(|i| i.try_into().ok());
```
darosior, line 147: We shouldn't silently ignore invalid parameters to `listaddresses`.

darosior:
I've addressed my comments here: https://github.com/darosior/liana/tree/get_all_addresses. Apparently i can't push them to your PR, so i'll open a follow-up..

darosior:
See https://github.com/wizardsardine/liana/pull/808.

darosior:
ACK 2660b77487d63218019413a4ca33b3a9629fbfc8 -- my requests are addressed in followup #808.

PR #715, opened 2023-09-14
installer: Finalize installation and start automatically on final step

### LINKED ISSUE #679: Installer: get rid of the final steps
It seems like it's superfluous to have a summary step. In addition you need to click twice to get to the wallet. We should probably get rid of it, start the wallet immediately when clicking next after the bitcoind step, and do something like #620 to be able to inspect your setup.

Thoughts?

jp1ac4:
This is to resolve #679.

The following changes have been made:
- Change title of final step to "Finalize installation"
- Install and load wallet automatically when reaching final step
- Remove fields from `Final` step that are no longer used

This is how the page looks when installing:
image

The following is only shown for ~1 sec once wallet has been installed and before loading wallet:
image

### DIFF
--- a/gui/src/installer/step/mod.rs
+++ b/gui/src/installer/step/mod.rs
@@ -20,7 +20,10 @@ use liana::miniscript::bitcoin::bip32::Fingerprint;
 
 use liana_ui::widget::*;
 
-use crate::installer::{context::Context, message::Message, view};
+use crate::{
+    bitcoind::Bitcoind,
+    installer::{context::Context, message::Message, view},
+};
 
 pub trait Step {
     fn update(&mut self, _message: Message) -> Command<Message> {
@@ -60,7 +63,7 @@ impl From<Welcome> for Box<dyn Step> {
 
 pub struct Final {
     generating: bool,
-    context: Option<Context>,
+    internal_bitcoind: Option<Bitcoind>,
     warning: Option<String>,
     config_path: Option<PathBuf>,
     hot_signer_fingerprint: Fingerprint,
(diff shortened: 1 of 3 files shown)


gui/src/installer/step/mod.rs
```diff
@@ -63,39 +62,34 @@ pub struct Final {
...
 
+impl Default for Final {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl Step for Final {
     fn load_context(&mut self, ctx: &Context) {
         self.context = Some(ctx.clone());
```
edouardparis, line 86: Maybe no need to clone the whole context anymore, instead keep only the internal_bitcoind used line 106

jp1ac4: I've changed this now.

edouardparis:
Sorry last comment:
Can you forbid user to click on `previous` while generating = true ?
User should be able to click on previous only if an error is returned.

jp1ac4:
In https://github.com/wizardsardine/liana/pull/715/commits/f771e674faccf150a5210b3abe829d443619c3e6, I've updated the `layout()` function so that if the `previous_message` parameter is `None`, the "Previous" button cannot be clicked. Before, passing `None` meant to use the default `Message::Previous`. Now, I pass `Message::Previous` explicitly instead of `None` each time `layout()` is called (except in `start_internal_bitcoind()` where a different message is passed). For the final step, I only allow to click "Previous" in case `generating` is `false` and an error has been raised.

Here's an example with an error. The previous button can be clicked still:
image

Otherwise, it's greyed out:
image

edouardparis (approved):
ACK f771e674faccf150a5210b3abe829d443619c3e6

PR #717, opened 2023-09-18
installer: show message when starting already-installed bitcoind

### LINKED ISSUE #714: Installer: No message shown when starting already-installed Liana-managed bitcoind
During installation, if the user chooses a Liana-managed Bitcoin node and a Liana-managed bitcoind is already installed (e.g. from having created a wallet on a different network), no message is shown while bitcoind is starting:

image

In the case of a newly installed bitcoind, "Starting..." is shown:

image

jp1ac4:
This is to resolve #714.

The "Starting..." message will now be shown also in case bitcoind was already installed on a previous occasion:

image

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -1064,15 +1064,17 @@ pub fn start_internal_bitcoind<'a>(
                         )
                     }
                 })
-            } else if let Some(InstallState::Finished) = install_state {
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
ACK 7079c6f82b89bdd968b21a6b8bce37cff7e51d89

PR #720, opened 2023-10-03
gui: Trim text inputs

### LINKED ISSUE #323: GUI: trim whitespaces in inputs
For instance it'd be nice for when importing a descriptor or a PSBT.

jp1ac4:
This is to resolve #323.

Following suggestions in #323, I've added a new `TrimmedString` struct that takes a string and applies `trim()`. This type is then used as `form::Value<TrimmedString>` to ensure text inputs are trimmed.

I'm creating this PR as draft to check if this is the right approach before applying the change to other inputs.

I've also removed a view file that seems to have been created accidentally.


edouardparis:
Sorry, I may just realized that we may not need to create a particular type. It is possible to do:
```diff
diff --git a/gui/ui/src/component/form.rs b/gui/ui/src/component/form.rs
index d268d2f..b24e501 100644
--- a/gui/ui/src/component/form.rs
+++ b/gui/ui/src/component/form.rs
@@ -44,6 +44,18 @@ where
         }
     }
 
+    pub fn new_trimmed<F>(placeholder: &str, value: &Value<String>, on_change: F) -> Self
+    where
+        F: 'static + Fn(String) -> Message,
+    {
+        Self {
+            input: text_input::TextInput::new(placeholder, &value.value)
+                .on_input(move |s| on_change(s.trim().to_string())),
+            warning: None,
+            valid: value.valid,
+        }
+    }
+
```
And only replace the form::new() in the concerned views.

jp1ac4:
Thanks for the suggestion. I've now applied that change to all forms except those concerning (fingerprint) labels.

edouardparis (approved):
Works well !
ACK 9895b493f1d79ae8fc5a1591ef455ccda2d81ccd

PR #723, opened 2023-10-07
descriptors: prevent using a signer more than once in a single path

darosior:
This is necessary to support signers which sign for a single key at once. It also doesn't make any sense to reuse a signer within the same spending path, so rule it out before it creates any new edge cases.

For more about the Bitbox signer support, which motivated this change, see
https://github.com/wizardsardine/liana/pull/706#issuecomment-1744705808.


edouardparis (approved):
ACK 730409eb5299724cdbe1a4da06d272098a791621
Tested with gui to trigger the error.
GUI will have its own check in #731
The breaking change needs to be noted somewhere.

darosior:
Yeah, we'll have a "breaking changes" section in the release notes.

PR #724, opened 2023-10-08
bitcoind: don't set load_on_startup when loading watchonly wallet

### LINKED ISSUE #641: Post v2 don't set `load_on_startup` when loading the wallet
It'd unnecessarily try to update the flag over and over again on bitcoind's side, always with the same value.

darosior:
This was necessary to transition wallets that were created without this flag, but since v2 they are always created with it so this is redundant: it'd unnecessarily try to update the flag over and over again on bitcoind's side, always with the same value.

Fixes #641.

### DIFF
--- a/src/bitcoin/d/mod.rs
+++ b/src/bitcoin/d/mod.rs
@@ -652,10 +652,7 @@ impl BitcoinD {
         }
         let res = self.make_fallible_node_request(
             "loadwallet",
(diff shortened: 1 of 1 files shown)


edouardparis (commented):
utACK e072e896c7dc744147c38a7d36d9eedd413399fa

PR #725, opened 2023-10-09
gui: display (estimated) feerate

### LINKED ISSUE #558: GUI: display fee-rate in Transactions and PSBTs screens
Currently only the absolute fee is shown, also displaying (while hovering or next to the fee) the fee-rate in sats/vbytes would be useful.

I got annoyed by a tx that wasn't getting mined, and I would like to know how far off i'm in terms of fee-rate.

screenshot

jp1ac4:
This is to resolve #558.

For transactions that have been broadcast, the feerate is calculated using the transaction's actual size. For PSBTs, the feerate is calculated using the max satisfaction size for all inputs, whether they have been signed or not (see https://github.com/wizardsardine/liana/issues/558#issuecomment-1750128022).

In both cases, the calculated feerate currently has type `u64`.

This is an example of a transaction on the Home tab:
image

The same transaction on the Transactions tab:
image

On the PSBTs tab, the feerate is an estimate:
image

### DIFF
--- a/gui/src/daemon/model.rs
+++ b/gui/src/daemon/model.rs
@@ -30,6 +30,9 @@ pub struct SpendTx {
     pub change_indexes: Vec<usize>,
     pub spend_amount: Amount,
     pub fee_amount: Amount,
+    /// The maximum size difference (in virtual bytes) of
+    /// an input in this transaction before and after satisfaction.
+    pub max_sat_vbytes: usize,
     pub status: SpendStatus,
     pub sigs: PartialSpendInfo,
     pub updated_at: Option<u32>,
@@ -49,6 +52,7 @@ impl SpendTx {
         psbt: Psbt,
         coins: Vec<Coin>,
         sigs: PartialSpendInfo,
+        max_sat_vbytes: usize,
     ) -> Self {
         let mut change_indexes = Vec::new();
         let (change_amount, spend_amount) = psbt.unsigned_tx.output.iter().enumerate().fold(
@@ -87,6 +91,7 @@ impl SpendTx {
             change_indexes,
             spend_amount,
             fee_amount: inputs_amount - spend_amount - change_amount,
+            max_sat_vbytes,
             status,
             sigs,
         }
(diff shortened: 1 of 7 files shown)


darosior:
The screenshots look great to me.

gui/src/daemon/model.rs
```diff
@@ -122,6 +127,13 @@ impl SpendTx {
     pub fn is_self_send(&self) -> bool {
         !self.coins.is_empty() && self.spend_amount == Amount::from_sat(0)
     }
+
+    /// Feerate obtained if all transaction inputs have the maximum satisfaction size.
+    pub fn min_feerate_vb(&self) -> u64 {
+        let max_tx_vbytes =
+            self.psbt.unsigned_tx.vsize() + (self.max_sat_vbytes * self.psbt.inputs.len());
+        self.fee_amount.to_sat() / max_tx_vbytes as u64
+    }
```
darosior, line 136: Could you add a comment about how this assumes all inputs are internal for this computation? I think it's worth having for the future.

(Also that's an ok assumption in the context of this estimation.)

jp1ac4:
> From my personal taste, what do you thing to make the `(...v/byte)` information smaller than the total sats ?

That sounds good to me. Here's how it looks if I use an `H3_SIZE` space followed by feerate in `H4_SIZE`:

image
image
image

I think going to `H5_SIZE` might be too small:

image

edouardparis:
Seems good to me ! thanks !

gui/src/daemon/model.rs
jp1ac4: I've added a comment in https://github.com/wizardsardine/liana/pull/725/commits/846431ed53efd2b29aefabe84ed56d054c0d4b45.

jp1ac4:
I've now changed the text to `H4_SIZE` in https://github.com/wizardsardine/liana/pull/725/commits/846431ed53efd2b29aefabe84ed56d054c0d4b45.

edouardparis (approved):
ACK 846431ed53efd2b29aefabe84ed56d054c0d4b45
tested spend and recovery

darosior (commented):
Github ACK 846431ed53efd2b29aefabe84ed56d054c0d4b45 -- feerate estimate assumption seem fine for now.

I made sure we can't import a PSBT with an empty tx (which would have made all those divisions undefined).

PR #729, opened 2023-10-18
Make sure we're actuall caught up to tip before starting up

### LINKED ISSUE #726: [GUI] Payments sometimes don't show up at start.
Sometimes starting Liana, payments don't show up until Lianad refreshes.
In this screenshot (i hid the balance and expiration), there should be many payments listed, but it's empty.
Screenshot from 2023-10-16 10-52-22

darosior:
Fixes #726. We actually do update our internal state before we are synced but we weren't waiting for the chain to be entirely synced.

### DIFF
--- a/src/bitcoin/d/mod.rs
+++ b/src/bitcoin/d/mod.rs
@@ -751,12 +751,10 @@ impl BitcoinD {
     pub fn sync_progress(&self) -> SyncProgress {
         // TODO: don't harass lianad, be smarter like in revaultd.
         let chain_info = self.block_chain_info();
-        let percentage = roundup_progress(
-            chain_info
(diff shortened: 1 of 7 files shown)


darosior:
I think to entirely fix #726 we should avoid returning `1.0` in `getinfo` as sync progress until we are completely caught up. Alternatively the daemon could wait at startup to be entirely caught up.

darosior:
An alternative could be to not return a 100% sync progress in getinfo if our best in-DB tip isn't equal to bitcoind's best tip..

edouardparis (approved):
ACK 6dfa04f6c7f6e5649f44ccf1503d9ce6b75d7568
tested with gui on old unsynced testnet3

PR #731, opened 2023-10-19
installer: forbid keys with same master fg in same path

edouardparis:
2023-10-19T17:59:11,287163921+02:00
based on #723

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -1295,6 +1295,7 @@ pub fn edit_key_modal<'a>(
     form_xpub: &form::Value<String>,
     form_name: &'a form::Value<String>,
     edit_name: bool,
+    duplicate_master_fg: bool,
 ) -> Element<'a, Message> {
     Column::new()
         .push_maybe(error.map(|e| card::error("Failed to import xpub", e.to_string())))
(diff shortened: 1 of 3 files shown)


gui/src/installer/view.rs
```diff
@@ -1411,8 +1412,15 @@ pub fn edit_key_modal<'a>(
                         Column::new()
                     },
                 )
+                .push_maybe(
+                    if duplicate_master_fg {
+                        Some(text("This path already contains a key with the same master fingerprint").style(color::RED))
```
darosior, line 1417: image

Since we ask them to "select a signing device", i think we should use the same wording in the error message:
```suggestion
                        Some(text("A single signing device may not be used more than once per path. (It can still be used in other paths.)").style(color::RED))
```

I also explicitly they can still use the same signing device more than once in the whole descriptor, as it seems to have been a point of confusion in our Telegram.

darosior (commented):
Looks good to me, tested locally by updating to master. Just a suggestion regarding the wording of the error.

In the future it would be nice to grey out signing devices in the list that were already used for this path. Let's maybe open an issue for this?

edouardparis:
Fixed the wording and created the issue in #733

edouardparis (commented):
Self-ACK 70dea710ae4e4427402120cc8f54516d9f2cf9b4

PR #734, opened 2023-10-20
commands: add weights before converting to vbytes

jp1ac4:
The `sanity_check_psbt()` function checks, among other things, that the feerate in sats/vbytes is in the required range. When estimating the transaction size, converting each part to vbytes first can lead to a larger result due to rounding multiple times and therefore a lower feerate, which could fall below 1 and fail the sanity check.

This issue has arisen in #560 when creating a transaction using coin selection.

### DIFF
--- a/src/descriptors/mod.rs
+++ b/src/descriptors/mod.rs
@@ -17,7 +17,7 @@ pub use keys::*;
 pub mod analysis;
 pub use analysis::*;
 
-const WITNESS_FACTOR: usize = 4;
+pub const WITNESS_FACTOR: usize = 4;
 
 #[derive(Debug)]
 pub enum LianaDescError {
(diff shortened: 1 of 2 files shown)


src/descriptors/mod.rs
```diff
@@ -17,7 +17,7 @@ pub use keys::*;
 pub mod analysis;
 pub use analysis::*;
 
-const WITNESS_FACTOR: usize = 4;
+pub const WITNESS_FACTOR: usize = 4;
```
darosior, line 20: In the future we should avoid rolling our own constant and instead use rust-bitcoin's.

darosior (commented):
utACK 176859ed926febfa16055d9e8e8f57ddc72cb9ba -- great catch

PR #735, opened 2023-10-20
Delete empty labels when they are updated

edouardparis:
In order to delete a label, client sends in the
request a `null` as label value.

It simplifies on the client side the check for empty string in labels returned by the daemon.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -592,9 +592,24 @@ impl DaemonControl {
         Ok(())
     }
 
-    pub fn update_labels(&self, items: &HashMap<LabelItem, String>) {
+    pub fn update_labels(&self, items: &HashMap<LabelItem, Option<String>>) {
         let mut db_conn = self.db.connection();
-        db_conn.update_labels(items);
+        let mut update_labels = HashMap::<LabelItem, String>::new();
+        let mut delete_labels = Vec::<LabelItem>::new();
+        for (item, label) in items {
+            if let Some(label) = label {
+                update_labels.insert(item.clone(), label.clone());
+            } else {
+                delete_labels.push(item.clone());
+            }
+        }
+
+        if !update_labels.is_empty() {
+            db_conn.update_labels(&update_labels);
+        }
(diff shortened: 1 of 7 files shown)


darosior:
Do we want to differentiate between the empty string and a null value?

edouardparis:
Mmh You are right that `null` value makes more sense for a json rpc API. :+1:

darosior:
But Concept ACK, seems reasonable and if it helps using those in the GUI i'm down!

tests/test_rpc.py
```diff
@@ -952,3 +952,18 @@ def test_labels(lianad, bitcoind):
     assert res[inexistent_txid] == "inex_txid"
     assert res[inexistent_outpoint] == "inex_outpoint"
     assert res[random_address] == "bitcoind-addr"
+
+    lianad.rpc.updatelabels(
```
darosior, line 956: Nit
```suggestion
    # Delete 2 of the labels set above. They shouldn't be returned anymore.
    lianad.rpc.updatelabels(
```

src/commands/mod.rs
```diff
@@ -592,9 +592,24 @@ impl DaemonControl {
...
-        db_conn.update_labels(items);
+        let mut update_labels = HashMap::<LabelItem, String>::new();
+        let mut delete_labels = Vec::<LabelItem>::new();
+        for (item, label) in items {
+            if let Some(label) = label {
+                update_labels.insert(item.clone(), label.clone());
+            } else {
+                delete_labels.push(item.clone());
+            }
+        }
```
darosior, line 605: Err, we could do this by just passing the map through, it's unfortunate that we'd now have to allocate two new maps and clone every single item (twice) and label..

darosior (commented):
I think we can avoid all the new allocations by simply making the DB's `update_labels` take an `Option<String>`. I'll see if it can satisfy my allocation OCD and propose a change if it works.

edouardparis:
deprecated in favor oif #737

Issue #732, opened 2023-10-19
Labels improvements


darosior:
This issue is for discussing potential improvements to the labeling system, and tracking the modifications we'd like to make.

- [x] When updating the label(s) for a transaction:
  - [x] If the transaction has a single payment output, only have this label be editable at the top.
  - [x] If the transaction has more than one payment output, have a "Batch label" editable at the top, along with a label editable per payment output (inside the output itself, not at the top).
- [x] In the list of transactions and list of psbts. If the transaction has a single payment, no transaction label, and this single payment is labelled, display this single payment's label
- [x] In the list of coins. If a coin does not have a label, and is a change output for either a single-payment transaction with a payment label, or a multi-payments transaction with a batch label, display "Change of XXX" with XXX being the batch or single-payment label.
- [x] ?Always ever use/display tx labels in case of a batch, call them "batch labels"?
- [x] ?In transaction details, display the label in place of the title, with an "Edit" button. If no label display the current title (like "Outgoing payment") with a "set label" button?

darosior:
So @edouardparis raised the question in cases where we infer a label for an item that isn't labelled, should we:
- Store this label in DB
- Just display it in context

I would lean toward the former. Edouard do you see any reason to go with the latter instead? Would the former be too much complexity (need to have the context and all)?

edouardparis:
Well it will certainly be a mixed of both.
For example, the gui must display the list of coins. For a particular Coin that is a change, if it does not have a outpoint label, then gui needs to find the label of the transaction to display Change of XXX, It has the txid it its outpoint, so gui can fetch that, but what if it was a single payment transaction, then gui must find the outpoint of the payment to get the label...
I think:
* it is preferable that at the spend creation process the coin label is stored directly as "Change of XXXX" with the coin outpoint as a key, but then what if user change the payment label later for YYYYY. In database we have coin labelled as Change XXXX, but no payment or transaction with the same label... 
* then when user change the transaction, payment label, gui has to change in db the change coin label too.

darosior:
Sounds good to me to have a default label "Change of XXX" when creating the transaction. This seems more brittle to me for the GUI to have to update the label of the change under various conditions, but i understand how it would add some load to have to fetch the previous transactions. So, overall ACK what you said for the change.

Now what about the rest? I don't think for instance when modifying the payment label in a single-payment tx, we should store the same label as tx label? Or should we?

What do you think about the other points?

edouardparis:
[...]

PR #736, opened 2023-10-20
Enhance labelling

edouardparis:
tackle most issues of #732, may need another round of external tries


edouardparis (commented):
Self-ACK b88781e62828cc6fc56e49f3ad1ebdeb39788573

PR #737, opened 2023-10-21
Make it possible to delete labels

darosior:
This is an alternative to #735. I believe it's slightly better as it's shorter and does not introduce new allocations. I also addressed my review comments from #735.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -312,8 +312,9 @@ cover the requested feerate.
 
 ### `updatelabels`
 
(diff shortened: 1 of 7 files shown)


edouardparis (approved):
ACK 01ca960370b55042cfe364c14aa012f016e1fc2b
It simplifies the API.
Performance wise, multiple requests to database is worst than few allocations, but 99% of the usage with be update of one or two labels in the same jsonrpc request.

PR #742, opened 2023-10-24
commands: don't add derivation paths for keys from different path but same signer

darosior:
When creating a PSBT, we were checking whether the key was for this path by checking its origin. This check would return false positive for keys from other paths but same signer (which shares the same fingerprint).

Instead, check the entire origin for each key to make sure it's actually the one used in the path we are interested about.

Thanks to Edouard Paris for finding this bug.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -26,7 +26,7 @@ use std::{
 
 use miniscript::{
     bitcoin::{
-        self, address,
+        self, address, bip32,
         locktime::absolute,
         psbt::{Input as PsbtIn, Output as PsbtOut, PartiallySignedTransaction as Psbt},
     },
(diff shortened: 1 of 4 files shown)


edouardparis (approved):
ACK 24edaecbdcfc0b6b53ac3355ad22991a8b5de018
tested with GUI

darosior:
Added a functional which would fail to provide a signature on master. Reverting 24edaecbdcfc0b6b53ac3355ad22991a8b5de018 will make the test fail if we apply this diff (to avoid an unrelated failure):
```diff
diff --git a/tests/test_rpc.py b/tests/test_rpc.py
index 4e420ca..b9f89f9 100644
--- a/tests/test_rpc.py
+++ b/tests/test_rpc.py
@@ -374,7 +374,7 @@ def test_create_spend(lianad, bitcoind):
         )
 
     # We can sign it and broadcast it.
-    sign_and_broadcast(lianad, bitcoind, PSBT.from_base64(res["psbt"]))
+    # sign_and_broadcast(lianad, bitcoind, PSBT.from_base64(res["psbt"]))
 
     # Try creating a transaction that spends an immature coinbase deposit.
     addr = lianad.rpc.getnewaddress()["address"]
```

Had to fixup the existing test, hence the new push.

darosior:
self-ACK 2e1c54491e96fda79b3831dc0c079c59250fabe0 - only made the test more robust since Edouard's ACK

PR #743, opened 2023-10-24
gui: coinbase transaction output scripts may not represent valid addresses

### LINKED ISSUE #738: GUI crashes on selecting on a coinbase transaction
Got a report such as if you mine to a Liana address, the GUI will list the transactions (apparently as 0-value -not sure about the cause of this one), but selecting a transaction will make the GUI crash.

The GUI should handle the "immature" state of coinbase deposits correctly. It's treated by the daemon as unconfirmed and tested since #578. More generally, the GUI should never crash except if continuing to run the program would be harmful.

edouardparis:
close #738
liana-gui v2 also tried to display addresses from transaction outputs to user when clicking on transaction detail.

### DIFF
--- a/gui/src/app/view/home.rs
+++ b/gui/src/app/view/home.rs
@@ -158,17 +158,14 @@ fn event_list_view(i: usize, event: &HistoryTransaction) -> Column<'_, Message> 
                 .to_string(),
             ) {
                 Some(p1_regular(label))
+            } else if let Ok(addr) =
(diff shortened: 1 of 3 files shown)


darosior:
Just my OCD :)

edouardparis (commented):
Self-ACK 2f2975c2790e83c592cd0ff594d3e8ce8a7c75d9

Issue #708, opened 2023-09-09
[lianad] wrong message if try to launch with `--daemon` but `daemon = false` in daemon.toml 


pythcoiner:
try to run this:
``` shell
 lianad --daemon --conf ~/.liana/signet/daemon.toml
```

got this
```
A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.
A documented sample is available at 'contrib/lianad_config_example.toml' in the source tree (https://github.com/wizardsardine/liana/blob/v1.0/contrib/lianad_config_example.toml).
The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.
```

the message let user guess that the toml file is missing, but in fact it's just about the .toml contain this line:

```
daemon = false
```

 shouldn't we instead apply a priority on the arg pass to cli instead param in toml file?

darosior:
Your command is invalid, `lianad` takes only one argument. The `--conf`. All configuration parameters must be part of the file.

-------- Original Message --------

pythcoiner:
so in that case shouldn't we inform user that `--conf` is an invalid argument?

darosior:
`--conf` is not invalid.. This is the *only valid* argument. All others must be part of the config file pointed to by `--conf`.

-------- Original Message --------

pythcoiner:
i mispelled, i mean shouldn't we inform user that no other than `--conf` is valid argument?

darosior:
We could be clearer that `--conf` is the only accepted command line argument, yeah.

PR #744, opened 2023-10-24
lianad add msg if wrong arg

pythcoiner:
address #708  
not sure about wording

### DIFF
--- a/src/bin/daemon.rs
+++ b/src/bin/daemon.rs
@@ -19,6 +19,11 @@ fn parse_args(args: Vec<String>) -> Option<PathBuf> {
         return None;
     }
 
+    if args[1] != "--conf" {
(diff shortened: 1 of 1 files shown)


src/bin/daemon.rs
```diff
@@ -19,6 +19,10 @@ fn parse_args(args: Vec<String>) -> Option<PathBuf> {
         return None;
     }
 
+    if args[1] != "--conf" {
+        eprintln!("lianad take only one (mandatory) argument: --conf");
+    }
```
darosior, line 24: No, it's not mandatory. :)

```suggestion
    if args[1] != "--conf" {
        eprintln!("Only a single command line argument is supported: --conf. All other configuration parameters must be specified in the configuration file.");
        print_help_exit();
    }
```

darosior (commented):
utACK b824a6433c353e0c1d8c16a3db4bba5097874828

Thanks!

Issue #690, opened 2023-09-06
Managed Bitcoin Core updates


darosior:
Let's say Bitcoin Core 25.1 gets released. The following release of Liana should upgrade the 25.0 `bitcoind` to `25.1`.

Major versions upgrades should be performed more carefully, but should still be performed.

edouardparis:
I believe #746 has already tackle the issue. While experimenting, liana is downloading bitcoin 26.0 even if I have 25.0 as liana managed node for an other network in the <datadir>/bitcoind directory.

edouardparis:
from @jp1ac4 
> The installer will download and use the latest version if not installed, regardless of any earlier versions installed. The app itself will use the latest version that has been installed

Maybe the last thing we can do is to provide a way from wallet > settings > bitcoind, to update the node if it is managed by the gui.

darosior:
That sounds like a good idea indeed.

jp1ac4:
> Maybe the last thing we can do is to provide a way from wallet > settings > bitcoind, to update the node if it is managed by the gui.

Alternatively, we could download & install the latest version automatically when starting the GUI. In case of an error, we could fall back to using the existing version.

PR #746, opened 2023-10-26
gui: Bump managed bitcoind version for new installations to 25.1

jp1ac4:
This PR bumps the managed bitcoind version for new Liana installations to 25.1.

The GUI will look for the most recent managed bitcoind version available when starting bitcoind.

Once a new managed bitcoind version is installed, all wallets using managed bitcoind will use this new version, even if a wallet on a particular network was installed with a previous bitcoind version.

The approach taken here may change as part of #690.

### DIFF
--- a/gui/src/bitcoind.rs
+++ b/gui/src/bitcoind.rs
@@ -16,7 +16,11 @@ use std::os::windows::process::CommandExt;
 #[cfg(target_os = "windows")]
 const CREATE_NO_WINDOW: u32 = 0x08000000;
 
-pub const VERSION: &str = "25.0";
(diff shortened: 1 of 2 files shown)


darosior (approved):
ACK fbec5e9218b24190c8f0ae1e1a09baf808ac70fa -- neat PR.

Tested running this:
- on an existing datadir with an internal bitcoind 25.0 (started it seamlessly)
- on a newly created signet in a datadir where there was an existing bitcoind 25.0 (it uses the existing bitcoind)
- on a newly created signet in a new datadir (it downloads and uses 25.1)

PR #748, opened 2023-10-26
gui: display info-level logs when starting daemon

### LINKED ISSUE #475: GUI: Daemon "info" level logs when starting it
On mainnet it takes a few minutes for my `bitcoind` to load the watchonly wallet. Therefore the GUI stays stuck at the following screen with no apparent progress.
image

I think it would be nice to display the daemon info-level logs to let the user know there is actual progress.

jp1ac4:
This is to resolve #475.

I've tried to follow the approach suggested in https://github.com/wizardsardine/liana/issues/475#issuecomment-1778744189.

I was having trouble using a `mpsc::channel` in the `iced::Subscription` closure as it can't be shared between threads, so for now I've created a channel using `crossbeam_channel::unbounded` instead. Perhaps there's a different way of creating the subscription that would allow me to use `mpsc::channel`.

I'm using the same `Message` variant for both Liana and bitcoind logs, and distinguishing between them in `on_log()` based on the step. I could also use a different message for each case.

### DIFF
--- a/gui/src/logger.rs
+++ b/gui/src/logger.rs
@@ -10,8 +10,6 @@ use tracing_subscriber::{
     reload, Registry,
 };
 
-use crossbeam_channel::unbounded;
-
 const INSTALLER_LOG_FILE_NAME: &str = "installer.log";
 const GUI_LOG_FILE_NAME: &str = "liana-gui.log";
 
@@ -39,7 +37,6 @@ pub struct Logger {
         Registry,
     >,
     level_handle: reload::Handle<filter::LevelFilter, Registry>,
-    receiver: crossbeam_channel::Receiver<String>,
 }
 
 impl Logger {
@@ -51,8 +48,6 @@ impl Logger {
             .with_file(false);
         let (file_log, file_handle) = reload::Layer::new(file_log);
         let stdout_log = tracing_subscriber::fmt::layer().pretty().with_file(false);
-        let (sender, receiver) = unbounded::<String>();
-        let streamer_log = LogStream { sender }.with_filter(filter::LevelFilter::INFO);
         tracing_subscriber::registry()
             .with(
                 stdout_log
@@ -74,12 +69,10 @@ impl Logger {
                             && !metadata.target().starts_with("ledger_transport_hid")
                     })),
             )
-            .with(streamer_log)
             .init();
         Self {
             file_handle,
             level_handle,
-            receiver,
         }
     }
 
@@ -124,10 +117,6 @@ impl Logger {
             .modify(|layer| *layer.writer_mut() = BoxMakeWriter::new(Arc::new(file)))?;
         self.level_handle.modify(|filter| *filter = log_level)?;
         Ok(())
-    }
-
-    pub fn receiver(&self) -> crossbeam_channel::Receiver<String> {
-        self.receiver.clone()
     }
 }
 
--- a/gui/src/logger.rs
+++ b/gui/src/logger.rs
@@ -1,13 +1,16 @@
 use liana::miniscript::bitcoin::Network;
 use std::path::PathBuf;
 use std::{fs::File, sync::Arc};
-use tracing::error;
+use tracing::{error, Event, Subscriber};
 use tracing_subscriber::{
     filter,
     fmt::{format, writer::BoxMakeWriter, Layer},
+    layer,
     prelude::*,
     reload, Registry,
 };
+
+use crossbeam_channel::unbounded;
 
 const INSTALLER_LOG_FILE_NAME: &str = "installer.log";
 const GUI_LOG_FILE_NAME: &str = "liana-gui.log";
@@ -36,6 +39,7 @@ pub struct Logger {
         Registry,
     >,
     level_handle: reload::Handle<filter::LevelFilter, Registry>,
+    receiver: crossbeam_channel::Receiver<String>,
 }
 
 impl Logger {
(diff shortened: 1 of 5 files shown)


edouardparis:
Concept ACK

darosior:
I tested it on mainnet and it works fine. I'll let @edouardparis judge the approach (esp whether introducing this new dependency is necessary). I wanted to have this in v3, but i'm starting to lean toward having it in v4 instead given the bugs we've got to deal with already in v3.

darosior:
Is there anything left to do here? I don't know about the implementation, but i unloaded my lianad watchonly wallet from my local mainnet node for a couple days in order to test this and this worked like a charm. It displayed "Loading watchonly wallet." for around 10 seconds before starting up while before it would just appear to be freezing.

jp1ac4:
That's great to hear. I don't have any further changes planned for now. I'll change it to ready for review.

darosior:
Just a self reminder to re-assess the approach here in light of the bugs with the way we were connecting the logs from bitcoind: https://github.com/wizardsardine/liana/pull/773.

darosior:
> Just a self reminder to re-assess the approach here in light of the bugs with the way we were connecting the logs from bitcoind: #773.

Since this is using an unbounded channel it appears it would never create a deadlock of the kind we had with bitcoind's logs?

jp1ac4:
> Since this is using an unbounded channel it appears it would never create a deadlock of the kind we had with bitcoind's logs?

I suppose not, unless there's some external "global limit" imposed on its capacity, although we won't be logging as much as bitcoind's IBD so would be less likely to reach it in any case. But otherwise, as we won't be reading from the channel most of the time, is there a chance it might grow too large and consume lots of memory? Ideally there'd be a way to close the channel when we're done with it.

gui/src/logger.rs
```diff
@@ -117,4 +125,39 @@ impl Logger {
...
+/// Used to record log messages by sending them to the channel.
+struct LogStreamVisitor<'a> {
+    sender: &'a crossbeam_channel::Sender<String>,
+}
+
+impl<'a> tracing::field::Visit for LogStreamVisitor<'a> {
+    fn record_debug(&mut self, field: &tracing::field::Field, value: &dyn std::fmt::Debug) {
+        if field.name() == "message" {
+            let msg = format!("{:?}", value);
+            let _ = self.sender.send(msg);
```
edouardparis, line 160: We should check error indicating that the other side of the channel is disconnected in order to stop sending message in it.

gui/src/logger.rs
```diff
@@ -36,6 +39,7 @@ pub struct Logger {
         Registry,
     >,
     level_handle: reload::Handle<filter::LevelFilter, Registry>,
+    receiver: crossbeam_channel::Receiver<String>,
```
edouardparis, line 42: I believe it should be removed. 
the layer can be added in `set_running_mode` and the receiver returned as Result of the method

gui/src/logger.rs
```diff
@@ -68,10 +74,12 @@ impl Logger {
                             && !metadata.target().starts_with("ledger_transport_hid")
                     })),
             )
+            .with(streamer_log)
```
edouardparis, line 77: can be moved to `set_running_mode`

gui/src/logger.rs
edouardparis: If not possible, change it in Option<> and use take to pass it to loader.
The ideal would be that this Receiver is dropped once loader changed its state.

jp1ac4:
Thanks for the review. I'm trying to make those changes currently.

Just wondering whether it would also be worth considering a similar approach as for the bitcoind logs and reading the GUI log file?

darosior:
Reading the logs from the file we write them to certainly sounds more roundabout and brittle (as in there is more moving parts, more could go wrong). Another avenue i thought of but don't know whether it's possible is to "hook" the `log` messages sent by the daemon. Maybe the GUI's `tracing` can do that?

gui/src/logger.rs
jp1ac4: When adding it in `set_running_mode`, I'm not sure how to make it active without calling `init` as is done in `setup`. The subscriber returns a message that it's trying to read from a disconnected channel (and this message is shown in the GUI on the loader screen).

The second approach using `take` seems to work in terms of making the channel become disconnected once the loader has completed (or shortly thereafter).

gui/src/logger.rs
jp1ac4: I haven't found a way to use the error to tell this layer to stop being triggered. I was trying to set a flag of some kind, but the `on_event` method only takes `&self` so I didn't manage to.

gui/src/logger.rs
jp1ac4: We might be able to create an `mpsc` channel inside the iced subscription closure itself as in this example:
https://github.com/iced-rs/iced/blob/dd249a1d11c68b8fee1828d58bae158946ee2ebd/futures/src/subscription.rs#L286-L355
Then the channel should be dropped once the loader completes.

We would pass the sender back to the loader and would just need a way to subscribe to log events while the loader is running. Perhaps using the layer with https://docs.rs/tracing/latest/tracing/dispatcher/fn.set_default.html.

jp1ac4:
I've made a change so that only the function `start_bitcoind_and_daemon` is instrumented with the channel layer, which should mean that no log messages are being sent to the channel otherwise. Furthermore, the channel is created as part of `Loader::new()` and so should be dropped once loading completes and the app starts.

I didn't manage yet to wrap the channel layer around the global default subscriber, which means that while `start_bitcoind_and_daemon` is being executed, logs are not being written to any other layers and are only displayed in the Liana GUI. Once that function completes, the usual logging resumes.

gui/src/logger.rs
jp1ac4: With my latest changes, we may not need to do anything here as we only send messages to the channel while `start_bitcoind_and_daemon` is running.

jp1ac4:
I've set this back to draft for now as it needs further work to keep existing logging while the daemon is starting.

pythcoiner:
This PR is stale for 2+ years, I'm closing, feel free  to re-openif you re-gain some interest in it.

PR #722, opened 2023-10-06
Only include BIP32 derivations for a single spending path when creating PSBTs

darosior:

PR #749, opened 2023-10-27
Revert PRs #722 and #742: roll back to including all BIP32 derivations in the PSBT input

darosior:
#722 pruned the BIP32 derivation paths in PSBT inputs for other spending paths than the one this PSBT was created for. This was to make it possible to use them with the Bitbox (see #722 OP for more details).

These were merged without testing at all whether it would break signing with a Ledger Nano. Turns out it did. Our bad. Thanks to Pythcoiner for finding this bug.

A follow-up PR will add a helper to prune the derivation separately so that it can only be applied to some signers and not all of them.


edouardparis (approved):
utACK 17ca01322e7badd37fd4120e0aeec8d834d0d6df

PR #750, opened 2023-10-27
descriptors: introduce a helper to prune BIP32 derivation paths

darosior:
Based on #749, this provides clients of the library with a way to prune all the BIP32 derivation paths in a PSBT for keys which are not part of the given spending path.


edouardparis (approved):
utACK ea4c4462406f5a1860140ca92bb7fca34e1322e0
The prune interface is neat

PR #751, opened 2023-10-27
GUI: prune BIP32 derivations for other spending paths when signing with BitBox02

darosior:


pythcoiner:
ACK tested signing w/ liana hot wallet + Nano S+ (2.1.2) + specter (USB)

pythcoiner:
Note: using a wallet created with V2, i had this issue signing w/ nano s+:
image
to solve this i need re-register descriptor on ledger.

We should notice user that will upgrade to V3 they need to follow this step.

darosior:
Rebased after #751 and #750 landed in master. The issue pointed out by @pythcoiner above was fixed in #754, which was also merged.

edouardparis (approved):
ACK 99bbb76228de0762521cf9152fa35e267fca95b2

Issue #728, opened 2023-10-17
Crash at startup with a bitcoind itself recovering from a previous crash


darosior:
```
2023-10-17T10:16:45.596107Z  INFO liana_gui::daemon::client:44: getinfo

  2023-10-17T10:16:45.596327Z ERROR liana_gui::daemon::client:46: method getinfo failed: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })

  2023-10-17T10:16:45.629436Z  INFO liana::database::sqlite:157: Checking if the database needs upgrading.

  2023-10-17T10:16:45.630273Z  INFO liana::database::sqlite::utils:207: Database is up to date.

  2023-10-17T10:16:45.657570Z  INFO liana:185: Database initialized and checked.

  2023-10-17T10:16:45.657743Z  INFO liana::bitcoin::d:268: Checking the connection to bitcoind.

  2023-10-17T10:16:45.660795Z  INFO liana::bitcoin::d:270: Connection to bitcoind checked.

  2023-10-17T10:16:48.294269Z  INFO liana:306: Loading our watchonly wallet on bitcoind.

  2023-10-17T10:16:48.307436Z  INFO liana:309: Watchonly wallet loaded on bitcoind and sanity checked.

  2023-10-17T10:16:48.314616Z  INFO liana::bitcoin::poller::looper:360: Block chain synchronization progress: 100.00%

  2023-10-17T10:16:48.320769Z  INFO liana::bitcoin::poller::looper:185: Block chain reorganization detected. Looking for common ancestor.

  2023-10-17T10:16:50.249969Z ERROR liana:55: panic occurred at line 433 of file /home/darosior/projects/wizardsardine/liana/src/bitcoin/d/mod.rs: Some("We must not fail to make a request for more than a minute: Server(Rpc(RpcError { code: -5, message: \"Block not found\", data: None }))")
   0: liana::setup_panic_hook::{{closure}}
             at /home/darosior/projects/wizardsardine/liana/src/lib.rs:49:18
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/alloc/src/boxed.rs:1999:9
      std::panicking::rust_panic_with_hook
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:709:13
[...]

PR #752, opened 2023-10-27
bitcoind: don't crash on `getblockheader` failure

darosior:
Fixes https://github.com/wizardsardine/liana/issues/728.


darosior:
Cleaned this up a bit.

darosior:
ACK 663fbd2c287791c608a056825ac103c1366b4c33

I still need to look into the conditions under which bitcoind would return not found for a block it previously had, but this change is harmless in any case.

PR #753, opened 2023-10-27
contrib: correct minimum libc version in Debian package

darosior:
It was 2.34, it should be 2.33.


darosior:
self ACK d8d01df6070bdd470b46de90dfcae4a85f704812 -- trivial

PR #754, opened 2023-10-30
wallet: load name from settings

### LINKED ISSUE #745: GUI: breaking change in connection with Ledger
If i'm understanding a report we had correctly the name of the policy was also changed for the Ledger policies. This is a breaking change and we should still support policies registered as simply "Liana".

edouardparis:
If an wallet settings already exists, the wallet name attached to the hardware wallet credentials must be use, instead of the new generated one.

close #745

### DIFF
--- a/gui/src/app/wallet.rs
+++ b/gui/src/app/wallet.rs
@@ -48,6 +48,11 @@ impl Wallet {
         }
     }
 
+    pub fn with_name(mut self, name: String) -> Self {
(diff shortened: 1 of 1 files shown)


darosior:
ACK 1cbbb8c0befe3c168fbbd34129a2f4f2a8c8726f -- tested locally and discussed it with Edouard. I can't think of another place where we'd like to use the old wallet name.

PR #760, opened 2023-10-30
fix gui: update labels if user signed an unsaved tx

edouardparis:
while doing the spend process, a spend transaction and its labels are stored in the database in two case:
* the user clicked on save button
* the user clicked on sign and sign the psbt, triggering the update in database

During the latter, the labels were not updated, which is a bug that makes the user do the whole spending process quickly without storing the labels.


darosior:
Could you describe the bug and the bugfix?

edouardparis:
Added the description of the bug to the PR

PR #763, opened 2023-10-30
fix wallet name in settings during install

edouardparis:
DEFAULT_WALLET_NAME aka simple `Liana` was still used during creation of wallet settings during install
this should have been noticed if I did test the wallet creation process in detail and if I removed the public exposition of the const variable. SHAME
based on #760


darosior (commented):
ACK 2e59bfec8d30db112e654620748efdaad8f7c7ed -- this makes sense to me and it was tested by Kevin who found both bugs.

PR #765, opened 2023-10-30
fix gui: update labels if user signed an unsaved tx

edouardparis:
It is a fix from #760


edouardparis (commented):
Self-ACK 789f952433d7ccc0f09924359bd87b83b0bdcf55
change already present in 3.x

PR #766, opened 2023-10-30
fix wallet name in settings during install

edouardparis:
DEFAULT_WALLET_NAME aka simple Liana was still used during creation of wallet settings during install
this should have been noticed if I did test the wallet creation process in detail and if I removed the public exposition of the const variable. SHAME


edouardparis (commented):
Self-ACK 9a4b665fa1a471a7a6c7e0dea09065f9f01ec35d
commit is present in 3.x

PR #768, opened 2023-10-31
fix single payment output label

edouardparis:
When a transaction has only one payment,
then its txid label is attached to the payment label. While modifying the label of an output it did not change the whole transaction label, which is a bug.


edouardparis (commented):
Self-ACK b81f0bd409ad2244b5849fd04285f2df341763a3

PR #769, opened 2023-10-31
fix single payment output label

edouardparis:
When a transaction has only one payment,
then its txid label is attached to the payment label. While modifying the label of an output it did not change the whole transaction label, which is a bug.

backport of the fix for 3.x in #768


edouardparis (commented):
Self-ACK d56a9a8a40671ad0ab0f4d8ccb8c6fa861d40429

PR #770, opened 2023-11-01
loader: read bitcoind debug.log file

edouardparis:


gui/src/bitcoind.rs
```diff
@@ -96,6 +95,16 @@ pub fn internal_bitcoind_cookie_path(bitcoind_datadir: &Path, network: &Network)
...
 
+/// Path of the cookie file used by internal bitcoind on a given network.
+pub fn internal_bitcoind_debug_log_path(bitcoind_datadir: &Path, network: &Network) -> PathBuf {
+    let mut debug_log_path = bitcoind_datadir.to_path_buf();
+    if let Some(dir) = bitcoind_network_dir(network) {
+        debug_log_path.push(dir);
+    }
+    debug_log_path.push("debug.log");
+    debug_log_path
+}
```
darosior, line 106: nit: could be merged with the above.

gui/src/bitcoind.rs
```diff
@@ -206,8 +214,6 @@ impl Bitcoind {
 
         let mut process = command
             .args(&args)
-            .stdout(std::process::Stdio::piped())
-            .stderr(std::process::Stdio::piped())
```
darosior, line 210: Wait, no. It's needed below. Either we remove the error logging below as well, or we keep the pipes and somehow remove them later? If we can't do that better to get rid of the `process.wait_with_output` entirely then.

gui/src/loader.rs
```diff
@@ -300,6 +314,20 @@ impl Loader {
     }
 }
 
+fn read_end_bitcoind_log_file(
+    datadir_path: &PathBuf,
+    network: bitcoin::Network,
+) -> Result<File, std::io::Error> {
+    let path = internal_bitcoind_debug_log_path(&internal_bitcoind_datadir(datadir_path), &network);
+    tracing::info!("Reading bitcoind log file: {}", path.display());
```
darosior, line 335: That's going to pollute the logs isn't it? Make it TRACE level?

gui/src/loader.rs
```diff
@@ -300,6 +314,20 @@ impl Loader {
     }
 }
 
+fn read_end_bitcoind_log_file(
```
darosior, line 330: This function doesn't read anything? It opens the file and seeks to its end.

gui/src/loader.rs
```diff
@@ -275,20 +277,32 @@ impl Loader {
...
+                    );
+                    return Subscription::none();
+                }
+            };
+
+            iced::subscription::unfold(0, log_file, move |mut log_file| async {
+                let mut buffer = [0u8; 1];
+                let mut last_line = String::new();
+                // Read bytes from the file and append them to the last_line string until a newline character is found
+                while log_file.read_exact(&mut buffer).is_ok() {
```
darosior: Isn't this going to block until there is more content added to the file, since you seeked to the end?

darosior (changes requested):

gui/src/loader.rs
```diff
@@ -300,6 +327,20 @@ impl Loader {
...
+    datadir_path: &PathBuf,
+    network: bitcoin::Network,
+) -> Result<File, std::io::Error> {
+    let path = internal_bitcoind_debug_log_path(&internal_bitcoind_datadir(datadir_path), &network);
+    tracing::info!("Reading bitcoind log file: {}", path.display());
+    let mut file = File::open(&path)?;
+    // Get the file length
+    let file_len = file.metadata()?.len().saturating_sub(2);
+    // Seek to the end of the file
+    file.seek(SeekFrom::Start(file_len))?;
```
darosior, line 340: Might as well seek from the end?

gui/src/bitcoind.rs
jp1ac4: Could we use
```
.stdout(std::process::Stdio::null())
.stderr(std::process::Stdio::piped())
```
 as otherwise bitcoind's stdout will be written to Liana's logs. I think we can still pipe the stderr and use that to check for errors.

darosior: It would still risk to get filled up and eventually stuck bitcoind

gui/src/loader.rs
```diff
@@ -275,21 +282,36 @@ impl Loader {
...
+                    self.datadir_path.clone(),
+                    self.network,
+                    Option::<std::io::BufReader<File>>::None,
+                ),
+                move |(datadir_path, network, mut reader)| async move {
+                    let mut reader = if let Some(reader) = reader {
+                        reader
+                    } else {
+                        match read_end_bitcoind_log_file(&datadir_path, network) {
+                            Ok(file) => std::io::BufReader::new(file),
```
darosior, line 298: Wait using a BufReader to store the whole log file could use a ton of memory?..

darosior (approved):
ACK 55bc8c1eb73bccc4344b85280ada5a06b465fdf5 --  with the caveat that most of this is my code.

gui/src/loader.rs
```diff
@@ -270,26 +277,72 @@ impl Loader {
...
-                        Err(e) => Message::BitcoindLog(Some(e.to_string())),
-                        Ok(_) => Message::BitcoindLog(Some(buffer)),
+        if self.internal_bitcoind.is_some() {
+            let log_path = internal_bitcoind_debug_log_path(&self.datadir_path, self.network);
+            iced::subscription::unfold(
+                0,
+                (log_path, std::time::Instant::now()),
+                move |(log_path, last_poll)| async move {
+                    // Reduce the io load.
+                    if last_poll.elapsed() < Duration::from_millis(500) {
```
darosior, line 293: This can of course be adapted, i went for a minimal change from the pretty fast update we had before. But IMO even updating only every 5 seconds would be more than enough.

darosior (changes requested):
Actually, this still freezes.

darosior (approved):
ACK ac3063b78ed68f9e389f92837b8ce5713b927c82


No more freezes, didn't know i could sleep in a tokio async closure, thanks Ed!

darosior:
Fixed the clippy lint

darosior:
ACK 8c7aa0da01f84b2a0b6310e9042188fbffe3060d

PR #772, opened 2023-11-01
installer: do not reload last step on next message

edouardparis:
If a user clicked very fast a lot on the Next button to trigger multiple Next messages, then the last
step of the install reloads multiple time.
This commit introduce a check that ignore next message if the step is the last one.


darosior:
utACK a14f3afa2aab52f686c2bc0a445ca359282922ae -- code makes sense and the patch was tested by Kevin

PR #773, opened 2023-11-01
gui: don't read bitcoind's stdout to avoid a deadlock

darosior:
"Forward port" of #770.

It was possible to get into a state where we were waiting for lianad to start, which was waiting on a response from bitcoind which in turn was waiting on the GUI to make room in its piped stdout buffer. Deadlock.

Instead of getting its logs from stdout, read its debug.log file.


edouardparis (approved):
ACK d0065898f70fbe14be7d269a686110e3bbbe7af8

PR #777, opened 2023-11-01
installer: start new bitcoind download if not finished previously

jp1ac4:
In case the user clicks on "Previous" before the bitcoind download has finished, this PR will clear the incomplete download (or a download that completed with errors) from `InternalBitcoindStep`.

If the user then returns to this step, either a new download will start or the finished download will be available still. This fixes a bug following #746 for users that already have bitcoind v25.0 installed. If they click "Previous" while downloading 25.1, then upon returning to this step, the existing 25.0 bitcoind will be started.

### DIFF
--- a/gui/src/installer/step/bitcoind.rs
+++ b/gui/src/installer/step/bitcoind.rs
@@ -658,6 +658,13 @@ impl Step for InternalBitcoindStep {
                         bitcoind.stop();
                     }
                     self.internal_bitcoind = None;
+                    if let Some(download) = self.exe_download.as_ref() {
+                        // Clear exe_download if not Finished.
+                        if let DownloadState::Finished { .. } = download.state {
+                        } else {
+                            self.exe_download = None;
+                        }
+                    }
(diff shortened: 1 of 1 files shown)


gui/src/installer/step/bitcoind.rs
```diff
@@ -658,6 +658,13 @@ impl Step for InternalBitcoindStep {
                         bitcoind.stop();
                     }
                     self.internal_bitcoind = None;
+                    if let Some(download) = self.exe_download.as_ref() {
+                        // Clear exe_download if not Finished.
+                        if let DownloadState::Finished { .. } = download.state {
```
darosior, line 663: Note for the future: i think these types of conditions are more clearly expressed using the `matches!` macro

darosior:
ACK bdd16c73bcae64e17931bd62d70f4d85b9a6af9a -- code makes sense to me and Kevin tested it fixed (one of) the bug he reported.

PR #783, opened 2023-11-02
GUI: fix the display of missing signatures in PSBT analysis

darosior:
The analysis was wrongly displayed. For instance it could display a fingerprint as having both signed and not signed yet.

Clean up the code and fix the analysis.

### DIFF
--- a/gui/src/app/view/psbt.rs
+++ b/gui/src/app/view/psbt.rs
@@ -8,9 +8,8 @@ use iced::{
 use liana::{
     descriptors::{LianaPolicy, PathInfo, PathSpendInfo},
     miniscript::bitcoin::{
-        bip32::{DerivationPath, Fingerprint},
-        blockdata::transaction::TxOut,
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
utACK fc169dd8016c5f513fcce076af42fb601a61be84

gui/src/app/view/psbt.rs
```diff
@@ -471,66 +509,12 @@ pub fn path_view<'a>(
...
-                            Container::new(text(value.to_string()))
-                                .padding(10)
-                                .style(theme::Container::Pill(theme::Pill::Simple))
-                        })
-                    }),
-            ),
+            .push_maybe(row_unsigned)
+            .push_maybe(
+                (!sigs.signed_pubkeys.is_empty())
+                    .then_some(p1_regular(", already signed by ").style(color::GREY_3)),
```
edouardparis, line 515: neat, i should use that instead if a { Some } else { None }

PR #784, opened 2023-11-02
3.x: Backport #777:  "installer: start new bitcoind download if not finished"

darosior:
In case user clicks on Previous before bitcoind download has finished, this will clear the incomplete download from `InternalBitcoindStep` so that a new download will start if the user returns to this step.


darosior:
self-ACK f26c6c30175b0e228ec4ce147e0f6fa6ebc42fef -- was tested in #777

PR #807, opened 2023-11-10
gui: display headers sync progress too in loader

### LINKED ISSUE #780: GUI: loader: header sync progress in addition to block chain validation progress
In v2 we used to have all the logs, including the header sync, when starting up a fresh managed bitcoind. In #773 we fixed a bug in how we were reading the log and filtered the log lines to only display the `UpdateTip` ones. This means when bitcoind is syncing the headers the progress appears to be stuck at 0% on the genesis block (the first `UpdateTip`).

image

Not a big deal but maybe we'd like to also display the header sync log lines?

Relevant code: https://github.com/wizardsardine/liana/blob/fd484a95239857c4aa2db8286b0c49fa6ee1cb87/gui/src/loader.rs#L292-L326

darosior:
Maybe we should just get rid of the filtering since there is no stability guarantees wrt logging. On the other hand we'd probably need to increase the update frequency if we do that. In the meantime at least display also the blockheaders so the sync doesn't appear to be stuck for 5 minutes at startup.

Fixes #780.

### DIFF
--- a/gui/src/loader.rs
+++ b/gui/src/loader.rs
@@ -322,7 +322,11 @@ impl Loader {
                 let last_update_tip = reader
                     .lines()
                     .into_iter()
-                    .filter(|l| l.as_ref().map(|l| l.contains("UpdateTip")).unwrap_or(false))
+                    .filter(|l| {
+                        l.as_ref()
+                            .map(|l| l.contains("UpdateTip") || l.contains("blockheaders"))
+                            .unwrap_or(false)
+                    })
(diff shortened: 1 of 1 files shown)


jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/807/commits/88d00ae1a3c99b32a5bd0aa457b16bb1a8ad064f.

Tested with a new install on mainnet and both entries appear:
image
image

PR #813, opened 2023-11-15
db: check for change using address strings only

jp1ac4:
Comparing the address itself includes the network, but an unchecked signet address will have its network set as testnet by `assume_checked` and so change addresses will not be matched.

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -243,7 +243,10 @@ impl DatabaseConnection for SqliteConn {
         self.db_address(address).map(|db_addr| {
             (
                 db_addr.derivation_index,
-                address == &db_addr.change_address.assume_checked(),
+                // We only compare address strings in case `assume_checked()` uses a different network.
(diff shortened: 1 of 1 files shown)


src/database/mod.rs
```diff
@@ -243,7 +243,7 @@ impl DatabaseConnection for SqliteConn {
         self.db_address(address).map(|db_addr| {
             (
                 db_addr.derivation_index,
-                address == &db_addr.change_address.assume_checked(),
+                address.to_string() == db_addr.change_address.assume_checked().to_string(),
```
darosior, line 246: Can you add a small comment with the rationale please? It's sneaky and i don't want us to forget about it.

jp1ac4: I've now added a comment here.

darosior:
utACK 79c5f92d9c690c36308a9d3fa4b35a4c4f6e01d4

PR #815, opened 2023-11-15
Add derivation_index to GetAddressResult

edouardparis:
The derivation index is required for
for client to derive and verify the address
on hardware wallets.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -75,9 +75,10 @@ This command does not take any parameter for now.
 
 #### Response
 
-| Field         | Type   | Description        |
-| ------------- | ------ | ------------------ |
-| `address`     | string | A Bitcoin address  |
+| Field                  | Type    | Description                           |
+| ---------------------- | ------- | ------------------------------------- |
+| `address`              | string  | A Bitcoin address                     |
+| `derivation_index`     | integer | The derivation index for this address |
 
 
 ### `listaddresses`


src/commands/mod.rs
```diff
@@ -914,22 +914,22 @@ pub struct GetInfoResult {
...
 pub struct GetAddressResult {
     #[serde(deserialize_with = "deser_addr_assume_checked")]
-    address: bitcoin::Address,
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct GetLabelsResult {
-    pub labels: HashMap<String, String>,
+    pub address: bitcoin::Address,
+    pub derivation_index: u32,
```
darosior, line 918: Any reason for not making it a typesafe `ChildNumber`?

tests/test_rpc.py
```diff
@@ -35,6 +35,8 @@ def test_getaddress(lianad):
     assert "address" in res
     # We'll get a new one at every call
     assert res["address"] != lianad.rpc.getnewaddress()["address"]
+    # new address has derivation_index higher than the previous one
+    assert res["derivation_index"] < lianad.rpc.getnewaddress()["derivation_index"]
```
darosior, line 39: It should be equal to exactly `+2` right?

darosior (commented):
Concept ACK, just a question about type safety.

tests/test_rpc.py
edouardparis: the new one, yes I think

src/commands/mod.rs
edouardparis: Not particular, reason it is only exposure to json.

darosior:
It's missing a doc update. Will push that shortly.

darosior:
WTF

darosior:
My bad

Issue #236, opened 2022-12-15
Feature: Cancel TX, and other replacement transactions (RBF)


kloaec:
A cancel TX of an already broadcast (not mined) transaction would be a nice feature.
Automatically craft a transaction spending the inputs, select a higher fee(-rate), ask to sign and broadcast. Tracking of success/failure would also be very nice.

A generalization of this tx flow would also be very nice, allowing to EDIT an existing TX. It would allow changing amounts, potentially replacing inputs and outputs, and make a good fee estimation to actually follow RBF rules.
This is an advanced feature but would be pretty cool.

darosior:
I don't think it should be that much configurable. A simple "cancel" button à-la Electrum seems better.

I don't think going as far as an EDIT button is a good idea. It would be a very brittle feature with a very poor UX: it would sometimes fail for no reason apparent to an unsuspecting user.

A cancel button is a good middle ground.

darosior:
See also already opened #43 about this.

darosior:
This is being tackled, but is on hold. See #43 and the coin selection / RBF pull requests.

darosior:
Closing in favour of https://github.com/wizardsardine/liana/issues/43, where most of the discussion has been happening.

Issue #43, opened 2022-10-06
RBF support for Spend transactions


darosior:
The biggest part is probably going to be the GUI's. Should we have a "bump fee" interface or a whole "replace by another Spend" interface?

darosior:
I've made it part of the Minisafe v0 project and assigned the feature freeze milestone. But let's consider it a nice to have. If it slips to Minisafe v1 it's no big deal.

darosior:
We won't have time to do this for v0.

darosior:
I think this should take the form of:
1. A "bump fees" button
2. A "cancel" button

Also see the discussion in the duplicate #236.

darosior:
Can't assign it to them using Github's feature for some reason but @jp1ac4 is going to take care of this. :rocket:

jp1ac4:
Hi @darosior, as well as the "bump fees" and "cancel" buttons, would we also want an "enable RBF" toggle when creating a transaction? I see in https://github.com/wizardsardine/liana/blob/master/src/commands/mod.rs#L371 that RBF is currently always enabled.

darosior:
Let's not have the option for now, i don't see any reasonable scenario where a user would disable it.

jp1ac4:
OK yep that makes sense thanks.

Regarding bumping fees, do you think this should be done via the `bitcoin-cli bumpfee` command or should the replacement transaction be constructed manually as a new spend transaction?

jp1ac4:
Yep, I think calling `createspend` with a higher feerate should work to begin with and then we can add checks to handle any special cases, e.g. if additional inputs are needed for the increased fee.

I tested this out a bit by broadcasting two transactions that differed only in the feerate. The higher feerate replaced the lower feerate as expected and the replaced PSBT was labelled as deprecated. (Conversely, when I broadcast the lower feerate after the higher feerate, which wouldn't be possible with the "bump fee" button, I got an error that the replacement was rejected due to insufficient fees.)

darosior:
We also need to make sure the transaction has more fees (as in absolute fees, not feerate). See https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy for the replacement policy we need to make sure we respect.
-------- Original Message --------

jp1ac4:
[...]

PR #816, opened 2023-11-16
commands: add `rbfpsbt` command

jp1ac4:
This PR relates to #43 and #236.

It adds a `rbfpsbt` command that generates a PSBT to replace an existing transaction using RBF. This replacement can either preserve non-change outputs and simply bump fees or remove non-change outputs and effectively cancel the transaction. The inputs and change output may need to be updated in accordance with the higher fee.

I've also added a `getmempoolentry` call to the bitcoin interface that is used for checking information about descendant transactions.

To facilitate development, I've made some temporary changes in the GUI so that replacement PSBTs can be signed and broadcast, but these changes might not be part of this PR in the end.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -17,6 +17,7 @@ Commands must be sent as valid JSONRPC 2.0 requests, ending with a `\n`.
 | [`listspendtxs`](#listspendtxs)                             | List all stored Spend transactions                            |
 | [`delspendtx`](#delspendtx)                                 | Delete a stored Spend transaction                             |
 | [`broadcastspend`](#broadcastspend)                         | Finalize a stored Spend PSBT, and broadcast it                |
+| [`rbfpsbt`](#rbfpsbt)                                       | Create a new RBF Spend transaction                            |
 | [`startrescan`](#startrescan)                               | Start rescanning the block chain from a given date            |
 | [`listconfirmed`](#listconfirmed)                           | List of confirmed transactions of incoming and outgoing funds |
 | [`listtransactions`](#listtransactions)                     | List of transactions with the given txids                     |
@@ -256,6 +257,43 @@ This command does not return anything for now.
 
 | Field          | Type      | Description                                          |
 | -------------- | --------- | ---------------------------------------------------- |
+
+### `rbfpsbt`
+
+Create PSBT to replace the given transaction, which must point to a PSBT in our database, using RBF.
+
+This command can be used to either:
+- "cancel" the transaction: the replacement will include at least one input from the previous transaction and will have only
+a single output (change).
+- bump the fee: the replacement will include all inputs from the previous transaction and all non-change outputs
+will be kept the same, with only the change amount being modified as required.
+
+In both cases, the replacement transaction may include additional confirmed coins as inputs if required
+in order to pay the higher fee (this applies also when replacing a self-send).
+
+If the transaction includes a change output to one of our own change addresses,
+this same address will be used for change in the replacement transaction, if required.
+
+If the transaction pays to more than one of our change addresses, then the one receiving the highest value
+will be used as a change address in the replacement and the others will be treated as non-change outputs
+(i.e. removed for cancel or otherwise kept the same).
+
+If `feerate` is not passed to the command, the target feerate of the replacement will be set to the minimum value
+allowed in order to replace this transaction using RBF (see https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy for further details about this and other conditions that must be satisfied when using RBF).
+
+#### Request
+
+| Field       | Type              | Description                                                     |
+| ----------- | ----------------- | --------------------------------------------------------------- |
+| `txid`      | string            | Hex encoded txid of the Spend transaction to be replaced.       |
+| `is_cancel` | bool              | Whether to "cancel" the transaction or simply bump the fee.     |
+| `feerate`   | integer(optional) | Target feerate for the RBF transaction (in sat/vb).             |
+
+#### Response
+
+| Field          | Type      | Description                                          |
+| -------------- | --------- | ---------------------------------------------------- |
+| `psbt`         | string    | PSBT of the spending transaction, encoded as base64. |
 
 ### `startrescan`
 
--- a/gui/src/daemon/model.rs
+++ b/gui/src/daemon/model.rs
@@ -87,7 +87,12 @@ impl SpendTx {
                     } else {
                         status = SpendStatus::Broadcast
                     }
-                } else {
+                // The txid will be different if this PSBT is to replace another transaction
+                // that is currently spending the coin.
+                // The PSBT status should remain as Pending so that it can be signed and broadcast.
+                // Once the replacement transaction has been confirmed, the PSBT for the
+                // transaction currently spending this coin will be shown as Deprecated.
+                } else if info.height.is_some() {
                     status = SpendStatus::Deprecated
                 }
             }
--- a/src/bitcoin/d/mod.rs
+++ b/src/bitcoin/d/mod.rs
@@ -1120,18 +1120,47 @@ impl BitcoinD {
 
     /// Whether this transaction is in the mempool.
     pub fn is_in_mempool(&self, txid: &bitcoin::Txid) -> bool {
+        self.mempool_entry(txid).is_some()
+    }
+
+    /// Get mempool entry of the given transaction.
+    /// Returns `None` if it is not in the mempool.
+    pub fn mempool_entry(&self, txid: &bitcoin::Txid) -> Option<MempoolEntry> {
         match self
             .make_fallible_node_request("getmempoolentry", &params!(Json::String(txid.to_string())))
         {
-            Ok(_) => true,
+            Ok(json) => Some(MempoolEntry::from(json)),
             Err(BitcoindError::Server(jsonrpc::Error::Rpc(jsonrpc::error::RpcError {
                 code: -5,
                 ..
-            }))) => false,
+            }))) => None,
             Err(e) => {
                 panic!("Unexpected error returned by bitcoind {}", e);
             }
         }
+    }
+
+    /// Get the list of txids spending those outpoints in mempool.
+    pub fn mempool_txs_spending_prevouts(
+        &self,
+        outpoints: &[bitcoin::OutPoint],
+    ) -> Vec<bitcoin::Txid> {
+        let prevouts: Json = outpoints
+            .iter()
+            .map(|op| serde_json::json!({"txid": op.txid.to_string(), "vout": op.vout}))
+            .collect();
+        self.make_node_request("gettxspendingprevout", &params!(prevouts))
+            .as_array()
+            .expect("Always returns an array")
+            .iter()
+            .filter_map(|e| {
+                e.get("spendingtxid").map(|e| {
+                    e.as_str()
+                        .and_then(|s| bitcoin::Txid::from_str(s).ok())
+                        .expect("Must be a valid txid if present")
+                })
+            })
+            .collect()
     }
 
     /// Stop bitcoind.
@@ -1392,5 +1421,50 @@ impl<'a> CachedTxGetter<'a> {
         } else {
             None
         }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct MempoolEntry {
+    pub vsize: u64,
+    pub fees: MempoolEntryFees,
+}
+
+impl From<Json> for MempoolEntry {
+    fn from(json: Json) -> MempoolEntry {
+        let vsize = json
+            .get("vsize")
+            .and_then(Json::as_u64)
+            .expect("Must be present in bitcoind response");
+        let fees = json
+            .get("fees")
+            .as_ref()
+            .expect("Must be present in bitcoind response")
+            .into();
+
+        MempoolEntry { vsize, fees }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct MempoolEntryFees {
+    pub base: bitcoin::Amount,
+    pub descendant: bitcoin::Amount,
+}
+
+impl From<&&Json> for MempoolEntryFees {
+    fn from(json: &&Json) -> MempoolEntryFees {
+        let json = json.as_object().expect("fees must be an object");
+        let base = json
+            .get("base")
+            .and_then(Json::as_f64)
+            .and_then(|a| bitcoin::Amount::from_btc(a).ok())
+            .expect("Must be present and a valid amount");
+        let descendant = json
+            .get("descendant")
+            .and_then(Json::as_f64)
+            .and_then(|a| bitcoin::Amount::from_btc(a).ok())
+            .expect("Must be present and a valid amount");
+        MempoolEntryFees { base, descendant }
     }
 }
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -15,7 +15,7 @@ pub use crate::database::{CoinStatus, LabelItem};
 use bdk_coin_select::InsufficientFunds;
 use utils::{
     deser_addr_assume_checked, deser_amount_from_sats, deser_fromstr, deser_hex,
-    select_coins_for_spend, ser_amount, ser_hex, ser_to_string,
+    select_coins_for_spend, ser_amount, ser_hex, ser_to_string, unsigned_tx_max_vbytes,
 };
 
 use std::{
@@ -79,6 +79,7 @@ pub enum CommandError {
     /// Overflowing or unhardened derivation index.
     InvalidDerivationIndex,
     CoinSelectionError(InsufficientFunds),
+    RbfError(RbfErrorInfo),
 }
 
 impl fmt::Display for CommandError {
@@ -144,7 +145,8 @@ impl fmt::Display for CommandError {
                 "No coin currently spendable through this timelocked recovery path."
            ),
             Self::InvalidDerivationIndex => write!(f, "Unhardened or overflowing BIP32 derivation index."),
-           Self::CoinSelectionError(e) => write!(f, "Coin selection error: '{}'", e),
+            Self::CoinSelectionError(e) => write!(f, "Coin selection error: '{}'", e),
+            Self::RbfError(e) => write!(f, "RBF error: '{}'.", e)
         }
     }
 }
@@ -169,6 +171,29 @@ pub enum InsaneFeeInfo {
     InvalidFeerate,
     TooHighFee(u64),
     TooHighFeerate(u64),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RbfErrorInfo {
+    MissingFeerate,
+    SuperfluousFeerate,
+    TooLowFeerate(u64),
+    NotSignaling,
+}
+
+impl fmt::Display for RbfErrorInfo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            Self::MissingFeerate => {
+                write!(f, "A feerate must be provided if not creating a cancel.")
+            }
+            Self::SuperfluousFeerate => {
+                write!(f, "A feerate must not be provided if creating a cancel. We'll always use the smallest one which satisfies the RBF rules.")
+            }
+            Self::TooLowFeerate(r) => write!(f, "Feerate too low: {}.", r),
+            Self::NotSignaling => write!(f, "Replacement candidate does not signal for RBF."),
+        }
+    }
 }
 
 /// A candidate for coin selection when creating a transaction.
@@ -814,6 +839,251 @@ impl DaemonControl {
         self.bitcoin
             .broadcast_tx(&final_tx)
             .map_err(CommandError::TxBroadcast)
+    }
+
+    /// Create PSBT to replace the given transaction using RBF.
+    ///
+    /// `txid` must point to a PSBT in our database.
+    ///
+    /// `is_cancel` indicates whether to "cancel" the transaction by including only a single (change)
+    /// output in the replacement or otherwise to keep the same (non-change) outputs and simply
+    /// bump the fee.
+    /// If `true`, the only output of the RBF transaction will be change and the inputs will include
+    /// at least one of the inputs from the previous transaction. If `false`, all inputs from the previous
+    /// transaction will be used in the replacement.
+    /// In both cases:
+    /// - if the previous transaction includes a change output to one of our own change addresses,
+    /// this same address will be used for change in the RBF transaction, if required. If the previous
+    /// transaction pays to more than one of our change addresses, then the one receiving the highest
+    /// value will be used as a change address and the others will be treated as non-change outputs.
+    /// - the RBF transaction may include additional confirmed coins as inputs if required
+    /// in order to pay the higher fee (this applies also when replacing a self-send).
+    ///
+    /// `feerate_vb` is the target feerate for the RBF transaction (in sat/vb). If `None`, it will be set
+    /// to 1 sat/vb larger than the feerate of the previous transaction, which is the minimum value allowed
+    /// when using RBF.
+    pub fn rbf_psbt(
+        &self,
+        txid: &bitcoin::Txid,
+        is_cancel: bool,
+        feerate_vb: Option<u64>,
+    ) -> Result<CreateSpendResult, CommandError> {
+        let mut db_conn = self.db.connection();
+
+        if is_cancel && feerate_vb.is_some() {
+            return Err(CommandError::RbfError(RbfErrorInfo::SuperfluousFeerate));
+        }
+
+        let prev_psbt = db_conn
+            .spend_tx(txid)
+            .ok_or(CommandError::UnknownSpend(*txid))?;
+        if !prev_psbt.unsigned_tx.is_explicitly_rbf() {
+            return Err(CommandError::RbfError(RbfErrorInfo::NotSignaling));
+        }
+        let prev_outpoints: Vec<bitcoin::OutPoint> = prev_psbt
+            .unsigned_tx
+            .input
+            .iter()
+            .map(|txin| txin.previous_output)
+            .collect();
+        let prev_coins = db_conn.coins_by_outpoints(&prev_outpoints);
+        // Make sure all prev outpoints are coins in our DB.
+        if let Some(op) = prev_outpoints
+            .iter()
+            .find(|op| !prev_coins.contains_key(op))
+        {
+            return Err(CommandError::UnknownOutpoint(*op));
+        }
+        if let Some(op) = prev_coins.iter().find_map(|(_, coin)| {
+            if coin.spend_block.is_some() {
+                Some(coin.outpoint)
+            } else {
+                None
+            }
+        }) {
+            return Err(CommandError::AlreadySpent(op));
+        }
+        // Compute the minimal feerate and fee the replacement transaction must have to satisfy RBF
+        // rules #3, #4 and #6 (see
+        // https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md). By
+        // default (ie if the transaction we are replacing was dropped from the mempool) there is
+        // no minimum absolute fee and the minimum feerate is 1, the minimum relay feerate.
+        let (min_feerate_vb, descendant_fees) = self
+            .bitcoin
+            .mempool_spenders(&prev_outpoints)
+            .into_iter()
+            .fold(
+                (1, bitcoin::Amount::from_sat(0)),
+                |(min_feerate, descendant_fee), entry| {
+                    let entry_feerate = entry
+                        .fees
+                        .base
+                        .checked_div(entry.vsize)
+                        .expect("Can't have a null vsize or tx would be invalid")
+                        .to_sat()
+                        .checked_add(1)
+                        .expect("Can't overflow or tx would be invalid");
+                    (
+                        std::cmp::max(min_feerate, entry_feerate),
+                        descendant_fee + entry.fees.descendant,
+                    )
+                },
+            );
+        // Check replacement transaction's target feerate, if set, is high enough,
+        // and otherwise set it to the min feerate found above.
+        let feerate_vb = if is_cancel {
+            min_feerate_vb
+        } else {
+            feerate_vb.ok_or(CommandError::RbfError(RbfErrorInfo::MissingFeerate))?
+        };
+        if feerate_vb < min_feerate_vb {
+            return Err(CommandError::RbfError(RbfErrorInfo::TooLowFeerate(
+                feerate_vb,
+            )));
+        }
+        // Get info about prev outputs to determine replacement outputs.
+        let prev_derivs: Vec<_> = prev_psbt
+            .unsigned_tx
+            .output
+            .iter()
+            .map(|txo| {
+                let address = bitcoin::Address::from_script(
+                    &txo.script_pubkey,
+                    self.config.bitcoin_config.network,
+                )
+                .expect("address already used in finalized transaction");
+                (
+                    address.clone(),
+                    bitcoin::Amount::from_sat(txo.value),
+                    db_conn.derivation_index_by_address(&address),
+                )
+            })
+            .collect();
+        // Set the previous change address to that of the change output with the largest value
+        // and then largest index.
+        let prev_change_address = prev_derivs
+            .iter()
+            .filter_map(|(addr, amt, deriv)| {
+                if let Some((ind, true)) = &deriv {
+                    Some((addr, amt, ind))
+                } else {
+                    None
+                }
+            })
+            .max_by(|(_, amt_1, ind_1), (_, amt_2, ind_2)| amt_1.cmp(amt_2).then(ind_1.cmp(ind_2)))
+            .map(|(addr, _, _)| addr)
+            .cloned();
+        // If not cancel, use all previous outputs as destinations, except for
+        // the output corresponding to the change address we found above.
+        // If cancel, the replacement will not have any destinations, only a change output.
+        let destinations = if !is_cancel {
+            prev_derivs
+                .into_iter()
+                .filter_map(|(addr, amt, _)| {
+                    if prev_change_address.as_ref() != Some(&addr) {
+                        Some((addr, amt))
+                    } else {
+                        None
+                    }
+                })
+                .collect()
+        } else {
+            HashMap::new()
+        };
+
+        // If there was no previous change address, we set the change address for the replacement
+        // to our next change address. This way, we won't increment the change index with each attempt
+        // at creating the replacement PSBT below.
+        let change_address = prev_change_address.unwrap_or_else(|| {
+            let index = db_conn.change_index();
+            let desc = self
+                .config
+                .main_descriptor
+                .change_descriptor()
+                .derive(index, &self.secp);
+            desc.address(self.config.bitcoin_config.network)
+        });
+        // If `!is_cancel`, we take the previous coins as mandatory candidates and add confirmed coins as optional.
+        // Otherwise, we take the previous coins as optional candidates and let coin selection find the
+        // best solution that includes at least one of these. If there are insufficient funds to create the replacement
+        // transaction in this way, then we set candidates in the same way as for the `!is_cancel` case.
+        let mut candidate_coins: Vec<CandidateCoin> = prev_coins
+            .values()
+            .map(|c| CandidateCoin {
+                coin: *c,
+                must_select: !is_cancel,
+            })
+            .collect();
+        let confirmed_cands: Vec<CandidateCoin> = db_conn
+            .coins(&[CoinStatus::Confirmed], &[])
+            .into_values()
+            .filter_map(|c| {
+                // Make sure we don't have duplicate candidates in case any of the coins are not
+                // currently set as spending in the DB (and are therefore still confirmed).
+                if !prev_coins.contains_key(&c.outpoint) {
+                    Some(CandidateCoin {
+                        coin: c,
+                        must_select: false,
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect();
+        if !is_cancel {
+            candidate_coins.extend(&confirmed_cands);
+        }
+        let max_sat_weight: u64 = self
+            .config
+            .main_descriptor
+            .max_sat_weight()
+            .try_into()
+            .expect("it must fit");
+        // Try with increasing fee until fee paid by replacement transaction is high enough.
+        // Replacement fee must be at least:
+        // sum of fees paid by original transactions + incremental feerate * replacement size.
+        // Loop will continue until either we find a suitable replacement or we have insufficient funds.
+        let mut replacement_vsize = 0;
+        for incremental_feerate in 0.. {
+            let min_fee = descendant_fees.to_sat() + replacement_vsize * incremental_feerate;
+            let rbf_psbt = match self.create_spend_internal(
+                &destinations,
+                &candidate_coins,
+                feerate_vb,
+                min_fee,
+                Some(change_address.clone()),
+            ) {
+                Ok(psbt) => psbt,
+                // If we get a coin selection error due to insufficient funds and we want to cancel the
+                // transaction, then set all previous coins as mandatory and add confirmed coins as
+                // optional, unless we have already done this.
+                Err(CommandError::CoinSelectionError(_))
+                    if is_cancel && candidate_coins.iter().all(|c| !c.must_select) =>
+                {
+                    for cand in candidate_coins.iter_mut() {
+                        cand.must_select = true;
+                    }
+                    candidate_coins.extend(&confirmed_cands);
+                    continue;
+                }
+                Err(e) => {
+                    return Err(e);
+                }
+            };
+            replacement_vsize = unsigned_tx_max_vbytes(&rbf_psbt.psbt.unsigned_tx, max_sat_weight);
+
+            // Make sure it satisfies RBF rule 4.
+            if rbf_psbt
+                .psbt
+                .fee()
+                .expect("has already been sanity checked")
+                >= descendant_fees + bitcoin::Amount::from_sat(replacement_vsize)
+            {
+                return Ok(rbf_psbt);
+            }
+        }
+
+        unreachable!("We keep increasing the min fee until we run out of funds or satisfy rule 4.")
     }
 
     /// Trigger a rescan of the block chain for transactions involving our main descriptor between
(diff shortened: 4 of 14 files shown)


src/bitcoin/d/mod.rs
```diff
@@ -1394,3 +1411,85 @@ impl<'a> CachedTxGetter<'a> {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct GetMempoolEntryRes {
```
jp1ac4, line 1416: I haven't included all the fields in the `getmempoolentry` response yet.

src/commands/mod.rs
```diff
@@ -736,6 +774,370 @@ impl DaemonControl {
             .map_err(CommandError::TxBroadcast)
     }
 
+    // This is like `create_spend` with slightly different parameters.
+    fn create_rbf_spend(
```
jp1ac4, line 778: For now, this is almost the same as `create_spend` with some slightly different parameters. Perhaps a follow-up PR could refactor to reduce duplicate code.

src/bitcoin/d/mod.rs
```diff
@@ -1134,6 +1134,23 @@ impl BitcoinD {
         }
     }
 
+    /// Get mempool entry.
+    // TODO: Should existing calls to `is_in_mempool` use this function instead and check is_some()?
```
darosior, line 1138: Yes. :)

src/bitcoin/d/mod.rs
darosior: No need to include the fields we won't be using (see how we only include those in every other command result).

src/bitcoin/d/mod.rs
jp1ac4: That's done now. Do you think it should be called `get_mempool_entry()` or `mempool_entry()`, or perhaps a mixture of both depending on where the method is?

src/bitcoin/d/mod.rs
jp1ac4: Do you think it's worth removing those I've already added that are not currently being used?

darosior: Yes. Dead code is bad.

src/bitcoin/d/mod.rs
darosior: I generally follow this rule: longer but explicit / clear is better than shorter but implicit / less clear. If clear already, shorter is better. In this case i think `mempool_entry()` is a better fit.

src/commands/mod.rs
jp1ac4: After https://github.com/wizardsardine/liana/pull/821 with the addition of a change parameter, the only real difference between this method and `create_spend` will be:
- a `min_fee` parameter (I think we can avoid this parameter by just calling this function with an increasing feerate until it has the requried fee, which I'm already doing anyway given that the min fee depends on the size of the replacement transaction itself, see https://github.com/wizardsardine/liana/pull/816/files#diff-bf2f650cdc66abd13076c47b25e8c9f783f06ec335bbd30822fdd690547615fdR1100)
- whether to use a mix of mandatory and optional candidates or just those in `coins_outpoints` -> we could change the logic of `create_spend` so that optional candidates are used not when `coins_outpoints` is empty but when this extra Boolean parameter is `true`.

So I think by adding a Boolean parameter to `create_spend` in addition to #821, we could remove `create_rbf_spend`.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+                        must_select: false,
+                    }),
+            );
+        }
+        // Try with increasing feerate until fee paid by replacement transaction is high enough.
+        // Replacement fee must be at least:
+        // sum of fees paid by original transactions + incremental feerate * replacement size.
+        for i in 0..100_u64 {
+            println!("trying feerate of {} with i {}", feerate_vb + i, i);
+            let rbf_psbt = self.create_rbf_spend(
```
jp1ac4, line 1102: In case `change_index` is `None`, I can pass an arbitrary value here to avoid incrementing the change index while we try different feerates, and then once the necessary feerate has been found, we can do a final call that increments the change index if necessary.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+        // confirmed coins as optional.
+        // TODO: For `is_cancel`, shall we include just one previous input?
+        let mut candidate_coins: Vec<CandidateCoin> = prev_coins
+            .into_values()
+            .map(|c| CandidateCoin {
+                coin: c,
+                must_select: true,
+            })
+            .collect();
+        if !destinations.is_empty() {
```
darosior, line 1086: Why not including them if it's a self-spend?

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+                    .map(|c| CandidateCoin {
+                        coin: c,
+                        must_select: false,
+                    }),
+            );
+        }
+        // Try with increasing feerate until fee paid by replacement transaction is high enough.
+        // Replacement fee must be at least:
+        // sum of fees paid by original transactions + incremental feerate * replacement size.
+        for i in 0..100_u64 {
```
darosior, line 1100: Oof. No. You should simply compute the minimum fee of the replacement instead of blindly trying out increasing feerates.

src/commands/mod.rs
darosior: > a min_fee parameter (I think we can avoid this parameter by just calling this function with an increasing feerate until it has the requried fee

I don't think we should do that.

But we can still have a `create_spend_internal` function which takes more parameter and is called by both `create_spend` and `rbf_psbt`.

darosior (commented):
I'm a bit baffled by your iterating approach to get the minimum absolute fee.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+            let prev_outpoints: &Vec<bitcoin::OutPoint> = &prev_psbt
+                .unsigned_tx
+                .input
+                .iter()
+                .map(|txin| txin.previous_output)
+                .collect();
+            let prev_coins = db_conn.coins_by_outpoints(prev_outpoints);
+            assert_eq!(prev_outpoints.len(), prev_coins.len());
+            prev_coins
+        };
```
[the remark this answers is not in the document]
jp1ac4: Yep, I found it was easier this way to keep track of which variables are needed later on. But in case you'd prefer consistency with the rest of the code, I could also avoid doing that.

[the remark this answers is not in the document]
darosior: Interesting. No, you can keep doing it.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+                &txo.script_pubkey,
+                self.config.bitcoin_config.network,
+            )
+            .map_err(CommandError::Address)?;
+            match db_conn.derivation_index_by_address(&address) {
+                Some((ind, true)) => {
+                    // TODO: check if there is more than one change output?
+                    println!("found change at address {} and index {}", address, ind);
+                    change_index = Some(ind);
+                }
```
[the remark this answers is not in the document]
jp1ac4: My general assumption has been that change addresses are added automatically to the transaction as change output and not manually used as destinations by the user, but that's not guaranteed, even if a single change address is used. Perhaps in case there's more than one change address used, we could preserve all of them and not treat any as change output.

src/commands/mod.rs
jp1ac4: My thinking was that for a self-send, we only want to consider the same coins again, in the same way that we don't use coin selection for a new self-send. But perhaps for RBF, allowing to add an extra coin to cover the fee would make sense.

darosior: Yes i think it should be expected that if you are going to RBF, some new inputs may be added to the transaction. We should warn the user in the documentation. And maybe also in the GUI but i'm not sure how not to confuse the user who doesn't care (most users don't i think).

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+                    // Non-change outputs are handled in the same way whether they are ours or not.
+                    if !is_cancel {
+                        destinations.insert(address, bitcoin::Amount::from_sat(txo.value));
+                    }
+                }
+            };
+        }
+        // We take all previous inputs as mandatory candidates and, if not self-send, include
+        // confirmed coins as optional.
+        // TODO: For `is_cancel`, shall we include just one previous input?
```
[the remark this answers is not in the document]
jp1ac4: OK yep that makes sense. I think starting with the largest would be good.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+            .ok_or(CommandError::RbfError(RbfErrorInfo::TxNotYetBroadcast(
+                *txid,
+            )))?;
+        // Check sum of descendant counts of all directly conflicting transactions is not too high.
+        // We are assuming we only have one directly conflicting transaction.
+        if mempool_entry.descendant_count > RBF_MAX_DESCENDANT_COUNT {
+            return Err(CommandError::RbfError(RbfErrorInfo::TooManyDescendants(
+                mempool_entry.descendant_count,
+            )));
+        }
```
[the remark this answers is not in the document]
jp1ac4: Ah yes, I should also check the ancestor count. Perhaps also ancestor/descendant sizes.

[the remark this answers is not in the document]
darosior: Another option is to not try to exhaustively match the current RBF policy of Bitcoin Core, and just let it fail in `sendrawtransaction` below? I don't think there is much benefit to erroring early, and being exhaustive would mean investing some time into it. Probably not worth it.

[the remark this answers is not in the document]
darosior: Sounds good.

jp1ac4:
I rebased on master and updated the functional test for the new `rbfpsbt` command, which depends on the fix from #829.

Other changes relating to the comments above are still ongoing.

tests/test_rpc.py
```diff
@@ -16,6 +16,7 @@
     get_txid,
     spend_coins,
     sign_and_broadcast,
+    sign_and_broadcast_psbt,
```
darosior, line 19: Thanks for cleaning up my mess. :)

There is already a `sign_and_broadcast` as you can see so maybe we should avoid duplicating them. But not here. Let's not scope creep this PR.

gui/src/app/state/transactions.rs
```diff
@@ -87,7 +87,11 @@ impl State for TransactionsPanel {
                 Err(e) => self.warning = Some(e),
                 Ok(txs) => {
                     self.warning = None;
-                    for tx in txs {
+                    // FIXME: This is a quick hack to ensure order is consistent in order to quickly
+                    // check and switch between transactions.
+                    let mut txs_sorted = txs;
+                    txs_sorted.sort_by_key(|tx| tx.tx.txid());
+                    for tx in txs_sorted {
```
darosior, line 94: Seems to be worth on its own?

gui/src/daemon/model.rs
```diff
@@ -129,7 +133,14 @@ impl SpendTx {
             psbt,
             change_indexes,
             spend_amount,
-            fee_amount: inputs_amount - spend_amount - change_amount,
+            // FIXME: this is a quick hack to deal with descendant PSBTs whose ancestor has been replaced.
+            // At least one input coin no longer exists so fee would be negative.
+            // Perhaps such PSBTs should be deleted or hidden?
```
darosior, line 138: I think we should keep them but display a "Can't compute fees" placeholder. xref https://github.com/wizardsardine/liana/issues/820 and https://github.com/wizardsardine/liana/pull/825.

darosior (commented):
Looking good!

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: I've changed the `is_cancel` case so that in the first instance, it uses the previous coins as optional candidates and we let the coin selection decide which of them are best to use given the min fee and feerate. If there are insufficient funds using just the previous coins as candidates, it then uses the same approach as for `!is_cancel`, i.e. all previous coins are mandatory candidates and confirmed coins are optional.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+        let prev_feerate_vb = {
+            let prev_abs_fee = prev_psbt
+                .fee()
+                .expect("fee must be Ok as transaction is already in mempool");
+            let prev_tx_size: u64 = prev_psbt
+                .clone()
+                .extract_tx()
```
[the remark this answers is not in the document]
jp1ac4: I've updated to use the mempool entry for vsize and the fee. I'm using `base` from the result, but there's also `modified`. I'm not sure which of these is used by bitcoin core in the context of checking the directly conflicting transactions' feerates for RBF.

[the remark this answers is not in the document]
darosior: `base` is the one we should use. :+1:

gui/src/app/state/transactions.rs
jp1ac4: There might be a better way to order the transactions, and possibly other changes that would be helpful when viewing the transactions list, so shall we leave this for a separate PR?

I think the only GUI change required for RBF, apart from the negative fees issue for missing coins that will be addressed in #820 and #825, is to leave the replacement PSBT status as "Pending" instead of "Deprecated" so that it can be signed and broadcast. I've made that change, but still need to check if we can set the replaced PSBT to "Deprecated" once the replacement has been confirmed.

darosior: Sure.

jp1ac4: PSBTs with input coins that have been spent by another confirmed transaction will be shown as Deprecated.

I've made a further change so that PSBTs with only inputs that no longer exist are shown as Deprecated. This can happen if the PSBT was spending the outputs of a transaction that was replaced or otherwise no longer exists.

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: This approach for `is_cancel` using optional candidates doesn't always work. The replacement transaction is a self-send, but it can happen if we have optional candidates that coin selection gives us a changeless solution without the optional candidates rather than one with change using the optional candidates. I think I'll need to use just mandatory candidates, adding one at a time as originally suggested until we have a solution with change.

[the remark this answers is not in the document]
jp1ac4: I think this could also affect the case where we want to bump the fee of a self-send: the higher fee may lead to a solution with no change. I'll use a similar approach for that case too.

One of the available scoring metrics for coin selection is [`Changeless`](https://github.com/evanlinjin/bdk/blob/d620dc61e809a8c280d623d5275abe8413b4b00a/nursery/coin_select/src/metrics/changeless.rs#L5). In principle, we might be able to use a similar `WithChange` metric by editing this [line](https://github.com/evanlinjin/bdk/blob/d620dc61e809a8c280d623d5275abe8413b4b00a/nursery/coin_select/src/metrics/changeless.rs#L18) to end with `is_some()` instead of `is_none()`. Then we could just pass the mandatory and optional candidates without having to increment the number of candidates.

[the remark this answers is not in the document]
jp1ac4: I did a quick test using a modified version of the `LowestFee` metric:
```
pub struct LowestFeeWithChange<'c, C> {
    pub lowest_fee: LowestFee<'c, C>, // This is the metric we're currently using
    pub must_have_change: bool,
}
```
 so that its score is given by:
```
if cs.is_target_met(self.lowest_fee.target, drain)
    && (drain.is_some() | !self.must_have_change)
{
    self.lowest_fee.score(cs) // We reuse the `LowestFee` score function.
} else {
    None
}
```
So `LowestFeeWithChange` is basically the same as `LowestFee`, except that `score()` will return `None` in case there's no change (depending on `must_have_change`).

From my tests, this avoids the problems I was facing with changeless solutions and also avoids the need for the extra logic adding candidates one at a time.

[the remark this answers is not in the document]
jp1ac4: Hmm I think it's a bit more complicated as we would also need to change how `bound()` is implemented. I'll continue with the iterating approach for now, adding candidates one by one.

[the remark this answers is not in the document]
jp1ac4: Sorry for all the noise, but I think this approach is still working. I just had to apply the same logic to the fallback in case BnB gave no solution. I've made the changes in a separate branch:
https://github.com/wizardsardine/liana/compare/master...jp1ac4:liana:lowest-fee-with-change-metric

It may be worth considering this approach as it should be simpler than iterating over the set of candidates as well as the min fee.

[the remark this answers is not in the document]
darosior: This (the code and the explanation) makes sense to me.

It reminds me of the case where an output has more than one output paying to one of our addresses. Another approach could be to just set the output(s) as part of the destination, but it's ugly as it might create another change output while there is already one (or more). On the other hand it would conserve the value of the output(s). Can this ever be a reasonable expectation of the user? Hmm i don't think it'd be worth risking further duplication of the change.
TL;DR: i think you should go with your with-change metric.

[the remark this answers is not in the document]
jp1ac4: I also considered an approach of treating the change as a regular destination in order to avoid a changeless solution, but had similar concerns.

OK great, I'll add the with-change metric in a separate PR.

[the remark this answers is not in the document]
jp1ac4: PR for new metric: https://github.com/wizardsardine/liana/pull/832.

src/commands/mod.rs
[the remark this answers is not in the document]
darosior: > As discussed, if more than one of our change addresses is used in the existing transaction, we can keep one of them to use for change in the replacement transaction (and drop the others?).

No i don't think we should drop the others. We should just keep them like any other destinations.

> Perhaps the choice of change address to use in the replacement should not depend on its value, but rather its index? E.g. choosing the highest index so that there's less chance of reusing change addresses in a future transaction.

That's a good point but i think it's still better to choose the largest value change output, to take the fees from there to avoid having to increase the transaction size.

----

In general my thinking is: if the user has set the outputs the way they are, he has a reason. Let's avoid touching the structure of the outputs *except* if we can find a change output suitable to be reduced in order to pay for the fees, to avoid including more inputs and/or an additional change output.

src/commands/mod.rs
```diff
@@ -736,6 +753,403 @@ impl DaemonControl {
...
+        }
+        // Try with increasing feerate until fee paid by replacement transaction is high enough.
+        // Replacement fee must be at least:
+        // sum of fees paid by original transactions + incremental feerate * replacement size.
+        // Loop will continue until either we find a suitable replacement or we have insufficient funds.
+        let mut rbf_vsize = 0_u64;
+        loop {
+            // Note that `rbf_vsize` strictly increases on each iteration as otherwise `min_fee`
+            // will ensure that the PSBT meets the required replacement fee and the loop will exit.
+            let min_fee = mempool_entry.fees.descendant.to_sat() + rbf_vsize;
```
darosior, line 1096: Why start at 0? This will always fail.

jp1ac4: I think with 0 we could still get a replacement with a high enough fee due to the higher feerate, and also in case the change output falls below the dust threshold and is added to fees.

darosior: Fair enough.

src/commands/mod.rs
darosior: Looks good to me!

ICYMI: https://github.com/wizardsardine/liana/pull/816/files#diff-bf2f650cdc66abd13076c47b25e8c9f783f06ec335bbd30822fdd690547615fdR465-R466 the comment here is now outdated.

jp1ac4: Great, thanks for the quick feedback. I'll update that comment and check the others.

src/commands/utils.rs
```diff
@@ -244,3 +244,26 @@ pub fn select_coins_for_spend(
...
+
+/// An unsigned transaction's maximum possible size in vbytes after satisfaction.
+///
+/// This assumes all inputs are internal (or have the same `max_sat_weight` value).
+///
+/// `tx` is the unsigned transaction.
+///
+/// `max_sat_weight` is the maximum weight difference of an input in the
+/// transaction before and after satisfaction. Must be in weight units.
+pub fn unsigned_tx_max_vbytes(tx: &bitcoin::Transaction, max_sat_weight: u64) -> u64 {
```
jp1ac4, line 256: This function could be used in `sanity_check`, but I'll wait in case any changes are required here (we could also change `sanity_check` in a separate PR).

src/commands/mod.rs
jp1ac4: I added a comment to the API doc that additional coins may be used, including for self-sends.

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: I changed the logic so that it chooses the change address with the highest value and then the one with the highest index, in case there's more than one. The other change addresses are then treated in the same way as non-change outputs.

src/commands/mod.rs
```diff
@@ -736,6 +774,372 @@ impl DaemonControl {
...
+                .extract_tx()
+                .vsize()
+                .try_into()
+                .expect("vsize must fit in a u64");
+            prev_abs_fee
+                .checked_div(prev_tx_size)
+                .ok_or(CommandError::InsaneFees(InsaneFeeInfo::InvalidFeerate))?
+        };
+        // Check replacement transaction's feerate is greater than previous feerate (the only directly conflicting transaction).
+        // TODO: Should we throw error or just set feerate_vb = prev_feerate_vb + 1 and check `rbf_psbt` below?
```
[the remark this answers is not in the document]
jp1ac4: I've added some logic along these lines, although it might need some tweaking.

src/commands/mod.rs
jp1ac4: Instead of this initial idea, I instead pass the next change address directly to `create_spend_internal`. It might be that we still increment the change index once in an intermediate call and then the final transaction won't have change, but this approach seems simpler.

src/commands/mod.rs
```diff
@@ -780,6 +840,229 @@ impl DaemonControl {
...
+                must_select: !is_cancel,
+            })
+            .collect();
+        let confirmed_cands: Vec<CandidateCoin> = db_conn
+            .coins(&[CoinStatus::Confirmed], &[])
+            .into_values()
+            .filter_map(|c| {
+                // In case the user attempts RBF before the previous coins have been updated in the DB,
+                // they would be returned as confirmed and not spending.
+                if !prev_coins.contains_key(&c.outpoint) {
```
jp1ac4, line 1001: An alternative would be to wait for the spend info of all coins to be updated.

src/commands/mod.rs
```diff
@@ -780,6 +840,229 @@ impl DaemonControl {
...
+                feerate_vb,
+                min_fee,
+                Some(change_address.clone()),
+            ) {
+                Ok(psbt) => psbt,
+                // If we get a coin selection error due to insufficient funds and we want to cancel the
+                // transaction, then set all previous coins as mandatory and add confirmed coins as
+                // optional, unless we have already done this.
+                Err(CommandError::CoinSelectionError(_))
+                    if is_cancel && candidate_coins.iter().all(|c| !c.must_select) =>
```
jp1ac4, line 1041: An alternative approach later on might be for coin selection to pick at least one from a subset. E.g. we could use a metric that only returns a score if the selection includes at least one candidate from this subset. So for the cancel case, we would want at least one from the previous inputs, and then zero or more from the optional.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
+            .try_into()
+            .expect("it must fit");
+        // Try with increasing feerate until fee paid by replacement transaction is high enough.
+        // Replacement fee must be at least:
+        // sum of fees paid by original transactions + incremental feerate * replacement size.
+        // Loop will continue until either we find a suitable replacement or we have insufficient funds.
+        let mut rbf_vsize = 0_u64;
+        loop {
+            // Note that `rbf_vsize` strictly increases on each iteration as otherwise `min_fee`
+            // will ensure that the PSBT meets the required replacement fee and the loop will exit.
```
darosior, line 1027: nit: i think the sentence is a bit off.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
+            candidate_coins.extend(&confirmed_cands);
+        }
+        let max_sat_weight: u64 = self
+            .config
+            .main_descriptor
+            .max_sat_weight()
+            .try_into()
+            .expect("it must fit");
+        // Try with increasing feerate until fee paid by replacement transaction is high enough.
+        // Replacement fee must be at least:
```
darosior, line 1021: With increasing *fee*. We do increase the feerate but what it's doing is increasing the fee, so i think it's better guidance for the reader to say that.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
+        // sum of fees paid by original transactions + incremental feerate * replacement size.
+        // Loop will continue until either we find a suitable replacement or we have insufficient funds.
+        let mut rbf_vsize = 0_u64;
+        loop {
+            // Note that `rbf_vsize` strictly increases on each iteration as otherwise `min_fee`
+            // will ensure that the PSBT meets the required replacement fee and the loop will exit.
+            let min_fee = mempool_entry.fees.descendant.to_sat() + rbf_vsize;
```
darosior, line 1028: Actually `rbf_vsize` does not strictly increase, as you re-set it to the same value on every iteration below. We should have a test which exercises the case where there is >1 iteration, this would have caught the infinite loop here.

Suggestion for a fix:
```diff
diff --git a/src/commands/mod.rs b/src/commands/mod.rs
index c662300..d6ee1d9 100644
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1021,11 +1021,12 @@ impl DaemonControl {
         // Replacement fee must be at least:
         // sum of fees paid by original transactions + incremental feerate * replacement size.
         // Loop will continue until either we find a suitable replacement or we have insufficient funds.
-        let mut rbf_vsize = 0_u64;
-        loop {
+        let mut replacement_vsize = 0;
+        for incremental_feerate in 0.. {
             // Note that `rbf_vsize` strictly increases on each iteration as otherwise `min_fee`
             // will ensure that the PSBT meets the required replacement fee and the loop will exit.
-            let min_fee = mempool_entry.fees.descendant.to_sat() + rbf_vsize;
+            let min_fee =
+                mempool_entry.fees.descendant.to_sat() + replacement_vsize * incremental_feerate;
             let rbf_psbt = match self.create_spend_internal(
                 &destinations,
                 &candidate_coins,
@@ -1050,17 +1051,20 @@ impl DaemonControl {
                     return Err(e);
                 }
             };
-            rbf_vsize = unsigned_tx_max_vbytes(&rbf_psbt.psbt.unsigned_tx, max_sat_weight);
+            replacement_vsize = unsigned_tx_max_vbytes(&rbf_psbt.psbt.unsigned_tx, max_sat_weight);
 
+            // Make sure it satisfies RBF rule 4.
             if rbf_psbt
                 .psbt
                 .fee()
                 .expect("has already been sanity checked")
-                >= mempool_entry.fees.descendant + bitcoin::Amount::from_sat(rbf_vsize)
+                >= mempool_entry.fees.descendant + bitcoin::Amount::from_sat(replacement_vsize)
             {
                 return Ok(rbf_psbt);
             }
         }
+
+        unreachable!("We keep increasing the min fee until we run out of funds or satisfy rule 4.")
     }
 
     /// Trigger a rescan of the block chain for transactions involving our main descriptor between
```

src/commands/mod.rs
```diff
@@ -462,240 +447,63 @@ impl DaemonControl {
...
-            version: 2,
-            lock_time: absolute::LockTime::Blocks(absolute::Height::ZERO), // TODO: randomized anti fee sniping
-            input: Vec::with_capacity(coins_outpoints.len()), // Will be zero capacity for coin selection.
-            output: Vec::with_capacity(destinations.len()),
-        };
-        // Add the destinations outputs to the transaction and PSBT. At the same time
+        // Check the destination addresses are valid for the network and
         // sanity check each output's value.
-        let mut psbt_outs = Vec::with_capacity(destinations.len());
+        let mut destinations_checked: HashMap<bitcoin::Address, bitcoin::Amount> =
```
darosior, line 452: No need for the type declaration here

darosior (commented):
Haven't looked at the tests yet. This is looking good, i like the create_spend_internal interface, definitely paving the way for completely abstracting out the spend creation. Nice commits organization too, although i'd have made a ~~simple~~ single commit of 5fe6ae90d4191f3ce737d4e9532b956207d417d6 and 41d6a4bedf5269d7b1b280b069f9ddd31ecba40a so it can be reviewed using `--color-moved=dimmed-zebra`.

src/bitcoin/d/mod.rs
```diff
@@ -944,7 +944,10 @@ impl BitcoinD {
                         .get("walletconflicts")
                         .and_then(Json::as_array)
                         .expect("A valid list of wallet conflicts must always be present.");
-                    if confs == 0 && !conflicts.is_empty() && !self.is_in_mempool(&spending_txid) {
+                    if confs == 0
+                        && !conflicts.is_empty()
+                        && self.mempool_entry(&spending_txid).is_none()
```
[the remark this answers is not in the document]
jp1ac4: I'll revert this change and follow your suggestion instead. It's in its own commit so should be straightforward.

jp1ac4:
> i'd have made a simple commit of [5fe6ae9](https://github.com/wizardsardine/liana/commit/5fe6ae90d4191f3ce737d4e9532b956207d417d6) and [41d6a4b](https://github.com/wizardsardine/liana/commit/41d6a4bedf5269d7b1b280b069f9ddd31ecba40a) so it can be reviewed using `--color-moved=dimmed-zebra`.

Sure, I'll squash those two commits. I kept them separate just to keep options open as I wasn't sure what would be best.

src/commands/mod.rs
jp1ac4: Ah yes thanks, this was a leftover comment from the first approach increasing the feerate directly. Will change it.

src/commands/mod.rs
jp1ac4: Do you mean there's an infinite loop in the case that `rbf_vsize` stays the same or always? In some of my tests, there was more than one iteration and, printing the value of `rbf_vsize`, I saw that it increased from one iteration to the next.

I might have overlooked something, but my thinking was that
```
rbf_fee_{i+1} >= minfee_{i+1} = descendant_fee + rbf_vsize_{i}
```
so that if `rbf_vsize_{i} >= rbf_vsize_{i+1}` then 
```
rbf_fee_{i+1} >= descendant_fee + rbf_vsize_{i+1}
```
and so we would exit the loop. So we only continue in the loop if `rbf_vsize` strictly increases.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
+            return Err(CommandError::RbfError(RbfErrorInfo::TxAlreadyConfirmed(
+                *txid,
+            )));
+        }
+        let mempool_entry = self
+            .bitcoin
+            .mempool_entry(txid)
+            .ok_or(CommandError::RbfError(RbfErrorInfo::TxNotYetBroadcast(
+                *txid,
+            )))?;
```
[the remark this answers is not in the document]
jp1ac4: That sounds good. Will change it to handle if tx is not in the mempool.

src/commands/mod.rs
darosior: Yes, i'm not quite convinced `rbf_size` would always increase on every iteration.

src/commands/mod.rs
```diff
@@ -780,6 +840,229 @@ impl DaemonControl {
...
+            .map(|(addr, _, _)| addr);
+        // If not cancel, use all previous outputs as destinations, except for
+        // the output corresponding to the change address we found above.
+        // If cancel, the replacement will not have any destinations, only a change output.
+        let destinations: HashMap<bitcoin::Address, bitcoin::Amount> = if !is_cancel {
+            prev_derivs
+                .iter()
+                .filter_map(|(addr, amt, _)| {
+                    if prev_change_address != Some(addr) {
+                        Some((addr.clone(), *amt))
```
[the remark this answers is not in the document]
darosior: Probably a good idea to move `prev_change_address`'s `cloned()` from below to just above if you do that.

darosior:
Just realized that actually this doesn't support RBF'ing a recovery transaction as in this case we need to set the right sequence in the inputs.

src/commands/mod.rs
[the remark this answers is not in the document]
darosior: I don't think there is a way to distinguish both.

Good point. I had overlooked that. You can use `gettxspendingprevout` to list the transactions spending a given set of coins. This can be a more generalistic way of getting the target min fee: always use `gettxspendingprevout` and query `getmempoolentry` for each of them. If the list is empty, assume it was purged and use a minfee of 0?

src/commands/mod.rs
```diff
@@ -780,6 +840,229 @@ impl DaemonControl {
...
+            .ok_or(CommandError::UnknownSpend(*txid))?;
+        let prev_coins = {
+            let prev_outpoints: &Vec<bitcoin::OutPoint> = &prev_psbt
+                .unsigned_tx
+                .input
+                .iter()
+                .map(|txin| txin.previous_output)
+                .collect();
+            let prev_coins = db_conn.coins_by_outpoints(prev_outpoints);
+            assert_eq!(prev_outpoints.len(), prev_coins.len());
```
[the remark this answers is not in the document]
darosior: Presumably the tx would have been dropped because its fees were below the mempool minfee so they would need to bump the fees before re-broadcasting it.

src/commands/mod.rs
jp1ac4: My reasoning above uses the `min_fee` parameter in `create_spend_internal` to ensure the replacement pays at least this fee. and should apply in any case that `rbf_vsize_{i} >= rbf_vsize_{i+1}`. As long as the `min_fee` parameter works as expected, I don't think there can be an infinite loop here.

But I may have overlooked something, so will make the change you suggested to be on the safe side.

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: Would it be possible to look up the `spend_txid` for each coin in the DB instead of using `gettxspendingprevout`?

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: I've now made this change.

src/commands/mod.rs
jp1ac4: I've removed this and a similar one in `rbf_psbt()`.

src/commands/mod.rs
jp1ac4: I've now updated this using your suggestion.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
+        // In case no feerate has been given, set it to prev_feerate_vb + 1.
+        let feerate_vb = {
+            let min_feerate_vb = prev_feerate_vb.to_sat().checked_add(1).unwrap();
+            if let Some(rate) = feerate_vb {
+                if rate < min_feerate_vb {
+                    if !is_cancel {
+                        return Err(CommandError::RbfError(RbfErrorInfo::TooLowFeerate(rate)));
+                    } else {
+                        min_feerate_vb
+                    }
```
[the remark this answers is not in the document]
jp1ac4: This will now also error for cancel.

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: I've updated it now to get spending transactions from the DB.

Even if it's not in the DB or mempool, I always include the `txid` of the transaction to be replaced in the set of directly conflicting transactions so that the replacement's feerate will be higher. If the transaction to be replaced is not in the mempool, I make the assumption that the PSBT has been finalized so that we can get its fee and vsize, but perhaps I need to check this is the case as it's now possible to call the `rbfpsbt` command on an unsigned PSBT. I've added comments regarding this in the code.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
+        // The inputs in the replacement transaction will be either those coins used as inputs in the transaction
+        // to be replaced or "confirmed" coins. Therefore, we can assume the transaction to be replaced is
+        // the only one directly conflicting with the replacement transaction.
+        let mut db_conn = self.db.connection();
+
+        let prev_psbt = db_conn
+            .spend_tx(txid)
+            .ok_or(CommandError::UnknownSpend(*txid))?;
+        let prev_coins = {
+            let prev_outpoints: &Vec<bitcoin::OutPoint> = &prev_psbt
```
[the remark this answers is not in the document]
jp1ac4: I've now changed this.

doc/API.md
```diff
@@ -257,6 +258,29 @@ This command does not return anything for now.
 | Field          | Type      | Description                                          |
 | -------------- | --------- | ---------------------------------------------------- |
 
+### `rbfpsbt`
+
+Create PSBT to replace the given transaction using RBF.
```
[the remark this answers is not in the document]
jp1ac4: I've added this.

src/commands/mod.rs
```diff
@@ -1009,6 +1032,229 @@ impl DaemonControl {
...
 
+    /// Create PSBT to replace the given transaction using RBF.
+    ///
+    /// If the previous transaction includes a change output to one of our own change addresses,
+    /// this same address will be used for change in the RBF transaction, if required. If the previous
+    /// transaction pays to more than one of our change addresses, then the one receiving the highest
+    /// value will be used as a change address and the others will be treated as non-change outputs.
+    ///
+    /// `is_cancel` indicates whether to "cancel" the transaction by removing non-change outputs from
+    /// the replacement or otherwise to keep the same non-change outputs and simply bump the fee.
```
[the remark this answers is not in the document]
jp1ac4: I've updated the wording now so hopefully it's a bit clearer.

doc/API.md
```diff
@@ -257,6 +258,29 @@ This command does not return anything for now.
...
+transaction pays to more than one of our change addresses, then the one receiving the highest
+value will be used as a change address and the others will be treated as non-change outputs.
+
+#### Request
+
+| Field    | Type   | Description                                                                              |
+| -------- | ------ | ---------------------------------------------------------------------------------------- |
+| `txid`   | string | Hex encoded txid of the Spend transaction to be replaced.                                |
+| `is_cancel`   | bool | Whether to "cancel" the transaction by removing non-change outputs from the replacement or otherwise to keep the same non-change outputs and simply bump the fee. If `true`, the only output of the RBF transaction will be change and the inputs will include at least one of the inputs from the previous transaction. If `false`, all inputs from the previous transaction will be used in the replacement. In both cases, the RBF transaction may include additional confirmed coins as inputs if required in order to pay the higher fee (this applies also when replacing a self-send).|
```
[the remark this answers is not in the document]
jp1ac4: I've updated this now.

src/bitcoin/d/mod.rs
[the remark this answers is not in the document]
jp1ac4: I've now reverted the change and redefined `is_in_mempool`.

src/commands/mod.rs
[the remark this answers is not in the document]
jp1ac4: I now check for outpoints for which we have no coin in the DB and return an error.

jp1ac4:
> Sure, I'll squash those two commits. I kept them separate just to keep options open as I wasn't sure what would be best.

I've now done this.

src/commands/mod.rs
jp1ac4: I've now updated this.

jp1ac4:
I've now addressed the comments.

darosior:
@edouardparis can you have a quick look at 01d50c956ac0ad7ca7247fe7085c5ca8ab13ee0c?

src/commands/mod.rs
```diff
@@ -171,6 +173,19 @@ pub enum InsaneFeeInfo {
     TooHighFeerate(u64),
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RbfErrorInfo {
+    TooLowFeerate(u64),
```
jp1ac4, line 178: I think it would be helpful to include the min feerate here as well as / instead of the chosen feerate.

src/commands/mod.rs
```diff
@@ -816,6 +831,265 @@ impl DaemonControl {
...
+                min_feerate_vb.max(conflict_feerate_vb.to_sat().checked_add(1).unwrap());
+        }
+        // Check replacement transaction's target feerate, if set, is high enough,
+        // and otherwise set it to the min feerate found above.
+        let feerate_vb = feerate_vb.unwrap_or(min_feerate_vb);
+        if feerate_vb < min_feerate_vb {
+            return Err(CommandError::RbfError(RbfErrorInfo::TooLowFeerate(
+                feerate_vb,
+            )));
+        }
```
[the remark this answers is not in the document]
darosior: Conversely, i don't think there is any reason for setting the feerate when using a cancel.

src/commands/mod.rs
```diff
@@ -816,6 +831,265 @@ impl DaemonControl {
...
+                    .ok_or(CommandError::InsaneFees(InsaneFeeInfo::InvalidFeerate))?
+            };
+            // Feerate should be at least 1 sat/vb more than conflicting transactions.
+            min_feerate_vb =
+                min_feerate_vb.max(conflict_feerate_vb.to_sat().checked_add(1).unwrap());
+        }
+        // Check replacement transaction's target feerate, if set, is high enough,
+        // and otherwise set it to the min feerate found above.
+        let feerate_vb = feerate_vb.unwrap_or(min_feerate_vb);
+        if feerate_vb < min_feerate_vb {
```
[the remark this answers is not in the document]
darosior: I was wrong here, it's already accounted for in `min_feerate_vb` through the `checked_add` above.

darosior:
I've addressed my comments in separate commits for you to review and squash them in the appropriate ones.
- 575025da04dc756ce8a154d26eb6ddae4e677df8, cf10f1396f45b2190f844916196a5faabc98ea8a and af4730b28da123887b83eebdbbfd1c0bb95f342d address https://github.com/wizardsardine/liana/pull/816#discussion_r1416863484 and https://github.com/wizardsardine/liana/pull/816#discussion_r1416882081
- de2f5197573504d27d5b91c006a95a38f0176593 addresses https://github.com/wizardsardine/liana/pull/816#discussion_r1416075264
- c663bb1f3c40e9ae9540cbf7d7ae0639d356d590 addresses https://github.com/wizardsardine/liana/pull/816#discussion_r1416097845
- 96f2a5dd002127d8d24fa2336fb93f69329af298 addresses https://github.com/wizardsardine/liana/pull/816#discussion_r1416909534

I've also played with this a bit more. I think this is good to go if you agree with my changes.

src/commands/mod.rs
```diff
@@ -858,29 +868,32 @@ impl DaemonControl {
...
-            if let Some(op) = prev_outpoints
-                .iter()
-                .find(|op| !prev_coins.contains_key(op))
-            {
-                return Err(CommandError::UnknownOutpoint(*op));
-            }
-            prev_coins
-        };
+        if !prev_psbt.unsigned_tx.is_explicitly_rbf() {
+            return Err(CommandError::RbfError(RbfErrorInfo::NotSignaling));
```
jp1ac4, line 881: Should we rather check all the spending transactions in the mempool? Maybe using the `bip125-replaceable` field in `getmempoolentry`?

jp1ac4: I think the directly conflicting transactions should correspond to the `mempool_spenders` for the previous outpoints, but happy to come back to that in a follow-up.

src/commands/mod.rs
jp1ac4: I haven't changed this yet. We could leave for a follow-up.

darosior:
ACK 0ab00cd6580ddc1964f4f328a90ab9c5f3f87efa

PR #821, opened 2023-11-17
Make it possible to create a sweep transaction using `createspend`

darosior:
We add a way to specify what change address to use. This allows the caller to set an external address which has for effect to sweep all the inputs' value to this address, after deduction of the fees and the other (optionally) set destination addresses.

For instance, combined with a self-send (no other destination) and by setting all the wallet's unspent coins, this allows one to sweep all the funds of the wallet to an external address.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -156,15 +156,23 @@ create a single change output. This may be useful to "refresh" coins whose timel
 may be close to expiry without having to bear the complexity of computing the correct amount for the
 change output.
 
+The optional `change_address` parameter allows the caller to specify what address to use for the
+leftover funds after all destinations have been set. This can be used to "sweep" the wallet: use all
+the unspent coins as input, set the other destination(s), if any, then set the `change_address` to
+the address of the wallet to sweep the funds to. Note however this output would only be created if
+there is enough remaining funds after sending to the specified destinations. This command WILL NOT
+ERROR if there isn't enough leftover funds to create the change/sweep output.
+
 This command will refuse to create any output worth less than 5k sats.
 
 #### Request
 
-| Field          | Type              | Description                                                       |
-| -------------- | ----------------- | ----------------------------------------------------------------- |
-| `destinations` | object            | Map from Bitcoin address to value.                                |
-| `outpoints`    | list of string    | List of the coins to be spent, as `txid:vout`.                    |
-| `feerate`      | integer           | Target feerate for the transaction, in satoshis per virtual byte. |
+| Field            | Type              | Description                                                       |
+| ---------------- | ----------------- | ----------------------------------------------------------------- |
+| `destinations`   | object            | Map from Bitcoin address to value.                                |
+| `outpoints`      | list of string    | List of the coins to be spent, as `txid:vout`.                    |
+| `feerate`        | integer           | Target feerate for the transaction, in satoshis per virtual byte. |
+| `change_address` | string            | Address to be used for leftover amount, if any.                   |
 
 #### Response
 
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -412,16 +412,20 @@ impl DaemonControl {
         destinations: &HashMap<bitcoin::Address<bitcoin::address::NetworkUnchecked>, u64>,
         coins_outpoints: &[bitcoin::OutPoint],
         feerate_vb: u64,
+        change_address: Option<bitcoin::Address<bitcoin::address::NetworkUnchecked>>,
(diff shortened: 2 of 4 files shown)


darosior:
I should either explicit in the doc that a sweep address will never be treated as internal (no BIP32 derivs set in PSBT output) or detect internal addresses passed.

doc/API.md
```diff
@@ -156,15 +156,23 @@ create a single change output. This may be useful to "refresh" coins whose timel
 may be close to expiry without having to bear the complexity of computing the correct amount for the
 change output.
 
+The optional `change_address` parameters allows the caller to specify what address to use for the
```
edouardparis, line 159:
```suggestion
The optional `change_address` parameter allows the caller to specify what address to use for the
```

src/commands/mod.rs
```diff
@@ -412,16 +412,20 @@ impl DaemonControl {
...
         // with a target feerate and outputs. In addition, we support different modes (coin control
-        // vs automated coin selection, self-spend, etc..) which make the logic a bit more
+        // vs automated coin selection, self-spend, sweep, etc..) which make the logic a bit more
         // intricate. Here is a brief overview of what we're doing here:
         // 1. Create a transaction with all the target outputs (if this is a self-send, none are
         //    added at this step the only output will be added as a change output).
         // 2. Automatically select the coins if necessary and determine whether a change output
         //    will be necessary for this transaction from the set of (automatically or manually)
         //    selected coins. The output for a self-send is added there.
+        //    The change output is also (ab)used to implemet a "sweep" functionality. We allow to
```
edouardparis, line 426:
```suggestion
        //    The change output is also (ab)used to implement a "sweep" functionality. We allow to
```

edouardparis (approved):
Seems nice !

doc/API.md
darosior: Done.

src/commands/mod.rs
darosior: Done.

src/commands/mod.rs
```diff
@@ -478,16 +482,32 @@ impl DaemonControl {
...
+        // The change address may be externally specified for the purpose of a "sweep": the user
+        // would set the value of some outputs (or none) and fill-in an address to be used for "all
+        // the rest". This is the same logic as for a change output, except it's external.
+        struct InternalChangeInfo {
+            pub desc: descriptors::DerivedSinglePathLianaDesc,
+            pub index: bip32::ChildNumber,
+        }
+        let (change_spk, int_change_info) = if let Some(addr) = change_address {
+            let addr = self.validate_address(addr)?;
+            (addr.script_pubkey(), None)
```
jp1ac4, line 495: Should we check in case one of our own change addresses has been passed in the parameter? This would happen as part of RBF, for example, in order for a replacement transaction to use the same change address.

In the case of RBF, I don't think we would need to increment the next index in the DB, as it should already be bigger than the one we pass. But perhaps we should check that and increment the next index accordingly?

darosior: Yes, thanks for the reminder i need to address https://github.com/wizardsardine/liana/pull/821#issuecomment-1816362834.

darosior: Done.

tests/test_spend.py
```diff
@@ -322,3 +330,66 @@ def test_coin_selection(lianad, bitcoind):
...
+        lianad.rpc.getnewaddress()["address"]: 0.2,
+        lianad.rpc.getnewaddress()["address"]: 0.1,
+    }
+    txid = bitcoind.rpc.sendmany("", destinations)
+    bitcoind.generate_block(1, wait_for_mempool=txid)
+    wait_for(lambda: len(lianad.rpc.listcoins(["confirmed"])["coins"]) == 3)
+    received_coins = lianad.rpc.listcoins(["confirmed"])["coins"]
+    spent_coin = next(c for c in received_coins if c["amount"] == 0.5 * COIN)
+    destinations = {
+        "bcrt1qmm5t0ch7vh2hryx9ctq3mswexcugqe4atkpkl2tetm8merqkthas3w7q30": int(
```
jp1ac4, line 383: nit: Was there a reason to hardcode the address instead of using `bitcoind.rpc.getnewaddress()`?

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/6bd6218d64495e6b5ac844861434ee8b1564ea02.

tests/test_spend.py
darosior: I don't remember. :sweat_smile:

PR #823, opened 2023-11-20
gui: bump rust version

edouardparis:


darosior:
If we are going to do this we should bump the clippy version checked in CI to make sure we don't reintroduce any new warnings..

darosior:
Obviously if you are only going to fix the lints for the GUI, only bump the version in the CI for the GUI, not the daemon

gui/Cargo.lock
```diff
@@ -239,9 +239,9 @@ checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "base64"
-version = "0.21.0"
+version = "0.21.6"
```
darosior, line 242: This commit says it bumps the Liana version but it doesn't.

edouardparis: Yes, my bad. I need to update it

edouardparis:
Self-ACK 3196a8025ea9686cf76f407043b3ef2f85e0b4fa

PR #825, opened 2023-11-20
handle psbt missing input coins

edouardparis:


darosior (commented):
I'm not sure we should prevent users from importing PSBTs without witness_utxo: it doesn't seem right to only be checking one of the necessary fields. Why not just displaying a placeholder instead of the fees for PSBTs with no coin in inputs?

gui/src/app/view/psbt.rs
```diff
@@ -729,12 +739,12 @@ pub fn inputs_and_outputs_view<'a>(
...
+    coin: Option<&'a Coin>,
     labels: &'a HashMap<String, String>,
     labels_editing: &'a HashMap<String, form::Value<String>>,
 ) -> Element<'a, Message> {
-    let outpoint = coin.outpoint.to_string();
-    let addr = coin.address.to_string();
+    let outpoint = outpoint.to_string();
     Column::new()
         .width(Length::Fill)
         .push(
```
jp1ac4, line 750: Can the labelling be turned off for an `outpoint` with no associated `coin`? I'm still given the option to add a label, and then any value I enter disappears after saving.

gui/src/app/view/psbt.rs
```diff
@@ -572,12 +577,17 @@ pub fn inputs_and_outputs_view<'a>(
                         .style(theme::Button::TransparentBorder)
                     },
                     move || {
-                        coins
+                        tx.input
```
jp1ac4, line 580: Can we use `tx.input` also to get the number of coins spent above? For one of my PSBTs, it tells me "1 coin spent", but there are two inputs, one of which no longer exists.

jp1ac4 (commented):
I've added a couple of comments from some tests I did now using the same wallet I've been using for testing RBF.

gui/src/app/view/psbt.rs
jp1ac4: Also, should we use `if !tx.input.is_empty()` instead of `if !coins.is_empty()` above , as otherwise the inputs section will not be shown for a PSBT with only non-existent coins.

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/825/commits/40ade751e0a00cde2a8eec3260eb267534a75f61.

The three issues I found have now been resolved:
- the displayed number of coins spent includes the missing inputs
- I can edit and save labels also for missing inputs
- the inputs tab is shown for a PSBT with only missing inputs

PR #829, opened 2023-11-21
bitcoin: poller: fix spend expiration from mempool, again

darosior:
Always mark a spend dropped from mempool as expired if it, or a conflict spending this coin, wasn't mined. The previous logic was confused: it would only do so if a conflict was detected, but it's not the only reason for a tx to be dropped from the mempool.


src/bitcoin/mod.rs
```diff
@@ -269,44 +269,30 @@ impl BitcoinInterface for d::BitcoinD {
...
-                            // will just be set at the next poll in `spending_coins()`.
-                            if self.is_in_mempool(txid) {
-                                Some(Conflict::Dropped)
-                            } else {
-                                None
+                    tx.block.and_then(|block| {
+                        // Being part of our watchonly wallet isn't enough, as it could be a
+                        // conflicting transaction which spends a different set of coins. Make sure
+                        // it does actually spend this coin.
+                        for txin in tx.tx.input {
```
darosior, line 278: Could use a `find_map` to stick to the functional style here.

darosior: Done.

jp1ac4:
ACK https://github.com/wizardsardine/liana/commit/758a4d6578aba487ae6ea06d32ece247525aba8b.

I tested with the following scenario:
```
A ---\	    / D
     T_1 --- 
B ---/	    \ E ----\
		     T_2
C ------------------/
```
Transaction `T_2` spends an output from `T_1`. I then replaced `T_1` using RBF, which also removed `T_2`. The spend info from coin C was cleared as expected.

jp1ac4 (approved):

PR #832, opened 2023-11-23
commands: require change for self-send

jp1ac4:
This adds a new `LowestFeeChangeCondition` metric for use in coin selection. It's the same as `LowestFee` with the option to only find solutions with change.

This option is then used for self-sends to ensure only a solution with change will be returned.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -550,10 +550,13 @@ impl DaemonControl {
                 feerate_vb,
                 0, // We only constrain the feerate.
                 max_sat_wu,
+                is_self_send, // Must have change if self-send.
             )
             .map_err(CommandError::CoinSelectionError)?
         };
         // If necessary, add a change output.
+        // For a self-send, coin selection will only find solutions with change and will otherwise
+        // return an error. In any case, the PSBT sanity check will catch a transaction with no outputs.
         if change_amount.to_sat() > 0 {
             // Don't forget to update our next change index!
             let next_index = change_index
@@ -569,12 +572,6 @@ impl DaemonControl {
                 bip32_derivation: change_desc.bip32_derivations(),
                 ..PsbtOut::default()
             });
-        } else if is_self_send {
-            return Err(CommandError::InsufficientFunds(
-                selected_coins.iter().map(|c| c.amount).sum(),
-                None,
-                feerate_vb,
-            ));
         }
 
         // Iterate through selected coins and add necessary information to the PSBT inputs.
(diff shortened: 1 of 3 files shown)


darosior (commented):
How about making this error:
https://github.com/wizardsardine/liana/blob/ed35a796ffa07948ca27b1d58312ae600839c70e/src/commands/mod.rs#L556-L578

A sanity check failure instead? It should never happen now: we would always fail the coin selection above before.

src/commands/utils.rs
```diff
@@ -157,23 +191,43 @@ pub fn select_coins_for_spend(
...
         );
         selector.sort_candidates_by_descending_value_pwu();
-        // If more coins still need to be selected to meet target, then `change_policy(&selector, target)`
-        // will give `Drain::none()`, i.e. no change, and this will simply select more coins until
-        // they cover the target.
-        selector.select_until_target_met(target, change_policy(&selector, target))?;
+        // Select more coins until target is met and change condition satisfied.
+        loop {
+            let drain = change_policy(&selector, target);
+            if selector.is_target_met(target, drain) && (drain.is_some() | !must_have_change) {
```
darosior, line 213: Did you mean to use a binary or here?
```suggestion
            if selector.is_target_met(target, drain) && (drain.is_some() || !must_have_change) {
```

jp1ac4:
> A sanity check failure instead? It should never happen now: we would always fail the coin selection above before.

Ah yes, that's true. I think in that case the sanity check would already catch it, as it checks for `tx.output.is_empty()`. So it should be enough just to remove the `else if is_self_send { ... }` check in the snippet above.

src/commands/utils.rs
jp1ac4: Oops, thanks for spotting. That should be a logical or. I'll change it.

src/commands/mod.rs
```diff
@@ -569,12 +569,6 @@ impl DaemonControl {
                 bip32_derivation: change_desc.bip32_derivations(),
                 ..PsbtOut::default()
             });
-        } else if is_self_send {
```
darosior, line 572: Would have been nice to have a comment above about how `change_amount` must be positive in case of a self send, and how if this doesn't hold the sanity checks will catch this.

src/commands/utils.rs
```diff
@@ -157,23 +191,43 @@ pub fn select_coins_for_spend(
...
-            change_policy: &change_policy,
-        },
-        100_000,
-    ) {
+    let lowest_fee = LowestFee {
+        target,
+        long_term_feerate,
+        change_policy: &change_policy,
+    };
+    let must_have_change = base_tx.output.is_empty(); // must have change if it is a self-send
```
darosior, line 199: The doc-comment of the function should be updated to document this.

darosior: I think even better would be to pass it as a parameter, so this function does not have any notion of self-send: it just exposes an interface to select coins such as there is always a change. And we set this in `create_spend` where we already have this notion of self-spend.

darosior (commented):
ACK 7a188f253f3494541dce7559581d23d4517a5c5c

Looks good to me, but since more concepts from `create_spend` are bleeding into `select_coins_for_spend` (the self-send) we should either document it or tweak its interface a bit.

src/commands/utils.rs
jp1ac4: The new parameter sounds good, will add that.

src/commands/mod.rs
jp1ac4: Ah yes, I'll do that.

jp1ac4:
I've made the suggested changes and also fixed, in a separate commit, an unrelated part of documentation for `select_coins_for_spend`.

darosior:
ACK 6123b87a024c04906338aa68c6afbcf57d57fc0b

PR #848, opened 2023-12-05
Grey out harware wallets when unrelated or do not support method

### LINKED ISSUE #830: GUI: list but grey out devices incompatible with address verification
Some hardware signing devices do not support address verification via USB. Currently post #819 those would not be listed when a user clicks on "Verify address on hardware device". This can be quite confusing for the user. Instead we should list them but grey them out. Maybe with a tooltip telling them to derive it manually on the device themselves using the provided derivation index.

We could also grey out devices which aren't part of our descriptor, but that should be part of a wider issue to do so in all cases (signing, desc import, etc..). See https://github.com/wizardsardine/liana/issues/831 about this.

edouardparis:
close #830

### DIFF
--- a/gui/src/app/view/hw.rs
+++ b/gui/src/app/view/hw.rs
@@ -6,6 +6,7 @@ use crate::{
     app::view::message::*,
     hw::{HardwareWallet, UnsupportedReason},
 };
+use async_hwi::DeviceKind;
 
 pub fn hw_list_view(
     i: usize,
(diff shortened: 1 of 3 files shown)


jp1ac4 (commented):
I've tested https://github.com/wizardsardine/liana/pull/848/commits/2c9a862dfab2b0dc28dcc35b39645a1497d4ee4e using Specter.

When it's not part of the wallet, it is greyed out and I cannot click it when verifying the receive address. When it's part of the wallet, it's again greyed out, but I'm still able to click it and an error message appears. Not sure if that's expected.

edouardparis:
Fixed the bug that let select grey out specter, fixed wording, rebased

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/848/commits/eb57d9b003a90fe06bbb9caca2cf615f49fccc85.

I tested again with Specter and it is now greyed out and not clickable in both cases (part of the wallet or not).

> what is the error message please ?

The error message before was the same as the tooltip saying to check manually on the device.

jp1ac4:
ACK https://github.com/wizardsardine/liana/commit/3372e2f1e0be711dac71becf57541e48309e93e4.

PR #842, opened 2023-11-30
[PoC] Standalone module for creating a spend transaction

darosior:
Based on top of #816, this introduces a new `spend` module with a helper to create a transaction spending coins from the wallet. It can be leveraged to create regular or recovery transactions, and also replacement for them. All the data structures used by the exposed spend creation function are contained with this module, in order to make it usable without a `lianad`-specific database and Bitcoin interface.

This PR is structured in an incremental fashion. First we pull out the `create_spend_internal` method introduced in #816 into a standalone `spend` module, then we incrementally remove the cruft and the ties from the spend module to the other components.

PR #856, opened 2023-12-07
spend: Scale down the number of BnB rounds in debug mode and with a high number of candidates

### LINKED ISSUE #846: Reduce the number of BnB iterations depending on the number of coins

darosior:
Based on #842.

Performing BnB with a large number of candidates can be computation expensive. Especially in debug mode. It can make the software lag in situations where it's called repeatedly.
Reasonably scale down the number of rounds performed depending on the number of candidates and whether optimizations are enabled.

Fixes #846

### DIFF
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -312,12 +312,15 @@ fn select_coins_for_spend(
         lowest_fee,
         must_have_change,
     };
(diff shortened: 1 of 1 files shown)


darosior:
Rebased now that #842 is merged.

Issue #840, opened 2023-11-27
gui: Don't increment change index when running (auto) coin selection


jp1ac4:
This relates to https://github.com/wizardsardine/liana/pull/563#discussion_r1397277629.

Currently, when a user is creating a new spend, the GUI may call `create_spend` several times in order to select coins automatically each time the user edits the input values. Each time `create_spend` is called, the next change index in the DB is incremented.

Following https://github.com/wizardsardine/liana/pull/821, we can now fix the change address to use when calling `create_spend`. We could use this to prevent the change index incrementing on these intermediate calls to `create_spend`. We could either pass an arbitrary change address, e.g. for index 0, or we could get the next change address from the DB and use this.

jp1ac4:
Yep, this has been fixed.

For future reference, it should be noted that once a draft PSBT has been generated (so after clicking Next on the Send page), the change index will be incremented if this draft PSBT has a change output (see also https://github.com/wizardsardine/liana/pull/963).

Issue #826, opened 2023-11-20
Let coin selection use (some?) unconfirmed coins


darosior:
Currently, our automated coin selection only considers confirmed coins. Considering unconfirmed coins as well introduces new challenges and was left for future work (see https://github.com/wizardsardine/liana/pull/560#issuecomment-1784834125).

It would be nice to consider them though. Should we have a bias toward spending confirmed coins first? It's maybe part of a larger bias toward preferring to spend older coins (see https://github.com/wizardsardine/liana/issues/827)? Should we only consider unconfirmed coins we created ourselves (change) and not unconfirmed payments?

pythcoiner:
What about running 3 rounds in coin selection:
 - first round only confirmed coins, if  match break
 - second round confirmed coins + our unconfirmed coins, if  match break
 - third round confirmed coins + all unconfirmed coins, if match at least notice user we will try spend unconfirmed coins

darosior:
We keep stumbling upon new situations where https://github.com/wizardsardine/liana/issues/811 (i mean in the general sens of being able to pass warnings from the spend creation code to the GUI) would be useful.

darosior:
I think it's only ever reasonable to spend our own unconfirmed change by default. We could allow to spend unconfirmed external deposits but only if the user explicitly ask for it, and i think we can leave this toggle to future work (they can already set them manually).

Note that if we ever implement this toggle, a later "safe" spend could spend one of our unconfirmed change which itself was created by and "unsafe" spend which spends an unconfirmed deposit. We should either check this or be explicit about it (maybe the toggle should be a global configuration option in this case).

jp1ac4:
With regard to RBF, we currently consider as candidates those (possibly unconfirmed) input coins from the transaction to be replaced together with confirmed coins.

Policy 2 in https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy restricts which unconfirmed coins can be used in the replacement:
- If the transaction to be replaced is in the mempool, then this should be the only directly conflicting transaction with the replacement and so we would not be able to include any other unconfirmed coins, and so no changes are needed.
- In case the transaction to be replaced is not in the mempool and there is some other directly conflicting transaction, it may be that this other transaction is spending unconfirmed change that is not an input in the transaction to be replaced. According to the policy, I think we could include this unconfirmed change as a candidate in the replacement.

For now, I'll make the change for a regular new spend and consider RBF separately.

[...]

PR #863, opened 2023-12-08
Use create_spend to calculate amount left to select and coin selection

### LINKED ISSUE #822: GUI: setting a too high feerate for a Spend crashes the software
We should really avoid careless conversions. Or idk sanitize inputs. 

```
  2023-11-17T13:19:00.521450Z ERROR liana:55: panic occurred at line 278 of file src/app/state/spend/step.rs: Some("attempt to multiply with overflow")                                                            
   0: liana::setup_panic_hook::{{closure}}                                                                                                                                                                         
             at /home/darosior/.cargo/git/checkouts/liana-efb3a908181ef2cf/2d303b1/src/lib.rs:49:18                                                                                                                
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call                                                                                                                                              
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/boxed.rs:2021:9                                                                                                                  
      std::panicking::rust_panic_with_hook                                                                                                                                                                         
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:711:13                                                                                                                
   2: std::panicking::begin_panic_handler::{{closure}}                                                                                                                                                             
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:597:13                                                                                                                
   3: std::sys_common::backtrace::__rust_end_short_backtrace                                                                                                                                                       
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/sys_common/backtrace.rs:170:18                                                                                                     
   4: rust_begin_unwind                                                                                                                                                                                            
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5                                                                                                                 
   5: core::panicking::panic_fmt                                                                                                                                                                                   
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14                                                                                                                
   6: core::panicking::panic                                                                                                                                                                                       
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:117:5                                                                                                                
   7: liana_gui::app::state::spend::step::DefineSpend::amount_left_to_select                                                                                                                                       
             at src/app/state/spend/step.rs:278:34                                                                                                                                                                 
   8: <liana_gui::app::state::spend::step::DefineSpend as liana_gui::app::state::spend::step::Step>::update                                                                                                        
             at src/app/state/spend/step.rs:320:29                                                                                                                                                                 
   9: <liana_gui::app::state::spend::CreateSpendPanel as liana_gui::app::state::State>::update                                                                                                                     
             at src/app/state/spend/mod.rs:105:20                                                                                                                                                                  
  10: liana_gui::app::App::update                                                                                                                                                                                  
             at src/app/mod.rs:174:18                                                                                                                                                                              
  11: <liana_gui::GUI as iced::application::Application>::update                                                                                                                                                   
             at src/main.rs:248:17                                                                                                                                                                                 
  12: <iced::application::Instance<A> as iced_native::program::Program>::update                                                                                                                                    
             at /home/darosior/.cargo/registry/src/index.crates.io-6f17d22bba15001f/iced-0.9.0/src/application.rs:227:9                                                                                            
  13: iced_winit::application::update::{{closure}}
             at /home/darosior/.cargo/registry/src/index.crates.io-6f17d22bba15001f/iced_winit-0.9.1/src/application.rs:664:40
  14: iced_futures::backend::native::tokio::<impl iced_futures::executor::Executor for tokio::runtime::runtime::Runtime>::enter                                                                                     
             at /home/darosior/.cargo/git/checkouts/iced-01eedf742052ef90/2d8318b/futures/src/backend/native/tokio.rs:19:9     
  15: iced_futures::runtime::Runtime<Hasher,Event,Executor,Sender,Message>::enter                         
             at /home/darosior/.cargo/git/checkouts/iced-01eedf742052ef90/2d8318b/futures/src/runtime.rs:53:9                                                                                                       
  16: iced_winit::application::update                                                                     
             at /home/darosior/.cargo/registry/src/index.crates.io-6f17d22bba15001f/iced_winit-0.9.1/src/application.rs:664:23                                                                                      
```

https://github.com/wizardsardine/liana/blob/514535d8d6fec705c7271241f68276c42b918150/gui/src/app/state/spend/step.rs#L173-L236

edouardparis:
Fixes #822. based on #865

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -10,6 +10,7 @@ use liana::{
     },
     spend::{
         create_spend, CandidateCoin, SpendCreationError, SpendOutputAddress, SpendTxFees, TxGetter,
+        MAX_FEERATE,
     },
 };
 
@@ -346,7 +347,7 @@ impl Step for DefineSpend {
                     view::CreateSpendMessage::FeerateEdited(s) => {
                         if let Ok(value) = s.parse::<u64>() {
                             self.feerate.value = s;
-                            self.feerate.valid = value != 0;
+                            self.feerate.valid = value != 0 && value <= MAX_FEERATE;
                         } else if s.is_empty() {
                             self.feerate.value = "".to_string();
                             self.feerate.valid = true;
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -6,7 +6,10 @@ use iced::{Command, Subscription};
 use liana::{
     descriptors::LianaDescriptor,
     miniscript::bitcoin::{
-        self, address, psbt::Psbt, Address, Amount, Denomination, Network, OutPoint,
+        self, address, psbt::Psbt, secp256k1, Address, Amount, Denomination, Network, OutPoint,
+    },
+    spend::{
+        create_spend, CandidateCoin, SpendCreationError, SpendOutputAddress, SpendTxFees, TxGetter,
     },
 };
 
@@ -19,7 +22,7 @@ use crate::{
     app::{cache::Cache, error::Error, message::Message, state::psbt, view, wallet::Wallet},
     daemon::{
         model::{remaining_sequence, Coin, SpendTx},
-        Daemon, DaemonError,
+        Daemon,
     },
 };
 
@@ -73,7 +76,9 @@ pub struct DefineSpend {
     is_valid: bool,
     is_duplicate: bool,
 
+    network: Network,
     descriptor: LianaDescriptor,
+    curve: secp256k1::Secp256k1<secp256k1::VerifyOnly>,
     timelock: u16,
     coins: Vec<(Coin, bool)>,
     coins_labels: HashMap<String, String>,
@@ -85,7 +90,12 @@ pub struct DefineSpend {
 }
 
 impl DefineSpend {
-    pub fn new(descriptor: LianaDescriptor, coins: &[Coin], timelock: u16) -> Self {
+    pub fn new(
+        network: Network,
+        descriptor: LianaDescriptor,
+        coins: &[Coin],
+        timelock: u16,
+    ) -> Self {
         let balance_available = coins
             .iter()
             .filter_map(|coin| {
@@ -99,7 +109,7 @@ impl DefineSpend {
         let coins: Vec<(Coin, bool)> = coins
             .iter()
             .filter_map(|c| {
-                if c.spend_info.is_none() {
+                if c.spend_info.is_none() && !c.is_immature {
                     Some((c.clone(), false))
                 } else {
                     None
@@ -109,7 +119,9 @@ impl DefineSpend {
 
         Self {
             balance_available,
+            network,
             descriptor,
+            curve: secp256k1::Secp256k1::verification_only(),
             timelock,
             generated: None,
             coins,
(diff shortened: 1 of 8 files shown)


darosior:
Does this also fix #840?

edouardparis:
20231211_15h18m02s_grim

So it does not crash anymore, but I think I should add a limit of 1000 to the form to just have an "invalid" form in red instead of having the big error banner.

pythcoiner:
> So it does not crash anymore, but I think I should add a limit of 1000 to the form to just have an "invalid" form in red instead of having the big error banner.

do this will 'lock' the step so the user cannot continue further? or just a hint?
1000 is just 3 times the sats/vbytes we experience few weeks ago, what if the fee go over this amount in few weeks/month?

darosior:
> do this will 'lock' the step so the user cannot continue further? or just a hint?

The user will encounter the same error if they click on "Next".

>  1000 is just 3 times the sats/vbytes we experience few weeks ago, what if the fee go over this amount in few weeks/month?

Yeah we might have to update our "max sane feerate", but that should be separate from this which just reuses what we use elsewhere.

gui/src/app/state/spend/step.rs
```diff
@@ -175,110 +188,121 @@ impl DefineSpend {
...
+                            sequence: None,
+                            must_select: *selected,
+                        })
+                    } else {
+                        None
+                    }
+                })
+                .collect()
+        } else {
+            self.coins
```
jp1ac4, line 233: For automated coin selection, we currently only consider confirmed coins, so I think we'll need to filter for those here to keep the same behaviour (see also https://github.com/wizardsardine/liana/issues/826).

gui/src/app/state/spend/step.rs
```diff
@@ -175,110 +188,121 @@ impl DefineSpend {
...
+
+        let destinations: Vec<(SpendOutputAddress, Amount)> = self
+            .recipients
+            .iter()
+            .map(|recipient| {
+                (
+                    SpendOutputAddress {
+                        addr: Address::from_str(&recipient.address.value)
+                            .expect("Checked before")
+                            .assume_checked(),
```
jp1ac4, line 206: I'm not sure if this matters when creating the spend, but `assume_checked` would set the network of a signet address to testnet. Following https://github.com/wizardsardine/liana/pull/813, we ignore the network when looking up addresses, but there might be some other effect of having a different network set.

gui/src/app/state/spend/step.rs
```diff
@@ -175,110 +188,121 @@ impl DefineSpend {
...
-        let mut outputs: HashMap<Address<address::NetworkUnchecked>, u64> = HashMap::new();
-        for recipient in &self.recipients {
-            outputs.insert(
-                Address::from_str(&recipient.address.value).expect("Checked before"),
-                recipient.amount().expect("Checked before"),
-            );
+    /// redraft calculates the amount left to select and auto selects coins
+    /// if the user did not select a coin manually
+    fn redraft(&mut self, daemon: Arc<dyn Daemon + Sync + Send>) {
+        if !self.form_values_are_valid() || self.recipients.is_empty() {
```
jp1ac4, line 194: Perhaps in a follow-up we could also use this function for a self-transfer to check if the user sets a feerate that is too high and will not give any change.

gui/src/app/view/spend/mod.rs
```diff
@@ -176,7 +176,7 @@ pub fn create_spend_tx<'a>(
                                         Message::CreateSpend(CreateSpendMessage::FeerateEdited(msg))
                                     },
                                 )
-                                .warning("Invalid feerate")
+                                .warning("Feerate must be an integer inferior or equal to 1000 sats/vbyte")
```
jp1ac4, line 179: nit: I think it's more common to say "less than or equal to"

jp1ac4 (commented):
I've done some testing and it's looking good. I just had a few comments, mainly about using confirmed coins only for auto coin selection.

gui/src/app/state/spend/step.rs
edouardparis: Ok, Thanks for the info.

gui/src/app/state/spend/step.rs
edouardparis: I do not think it matters, signet address are testnet address. https://github.com/wizardsardine/liana/blob/master/src/commands/mod.rs#L232 also address verification is done in the edit address message cycle and does not concern this PR.

gui/src/app/state/spend/step.rs
edouardparis: Yes, we could also use this function to calculate the amount a `max` button feature.

darosior: Opened https://github.com/wizardsardine/liana/issues/868 to track this.

gui/src/app/state/spend/step.rs
darosior: > Following https://github.com/wizardsardine/liana/pull/813, we ignore the network when looking up addresses

Since this is using the spend module it's not connected to the database anymore at all.

> but there might be some other effect of having a different network set.

As long as we are only using this dummy spend for selecting coins, the address encoding should have any effect. In fact we could be taking `Script`s instead of addresses in: https://github.com/wizardsardine/liana/blob/3e0f82a71e031604fcf8bb0fd757ae95908e8ca6/src/spend.rs#L385-L389

This would avoid requiring this kind of ugliness from the callers.

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/a7324681e1072e4b53cd932cb49d462e9cc35956.

PR #864, opened 2023-12-08
gui: add option to delete wallet

### LINKED ISSUE #543: Add a remove wallet feature
In case someone want to remove his wallet for create a new one, or just before uninstall.

jp1ac4:
This is to resolve #543.

I've used a trash icon instead of a cross, but can change it to a cross if preferred.

### DIFF
--- a/gui/src/launcher.rs
+++ b/gui/src/launcher.rs
@@ -1,10 +1,14 @@
 use std::path::PathBuf;
 
-use iced::{widget::Space, Alignment, Command, Length, Subscription};
+use iced::{
+    widget::{tooltip, Space},
+    Alignment, Command, Length, Subscription,
+};
 
 use liana::{config::ConfigError, miniscript::bitcoin::Network};
 use liana_ui::{
-    component::{badge, card, text::*},
+    color,
+    component::{badge, button, card, modal::Modal, notification, text::*},
     icon, image, theme,
     util::*,
     widget::*,
@@ -12,10 +16,22 @@ use liana_ui::{
 
 use crate::app;
 
+fn wallet_name(network: &Network) -> String {
+    match network {
+        Network::Bitcoin => "Bitcoin Mainnet",
+        Network::Testnet => "Bitcoin Testnet",
+        Network::Signet => "Bitcoin Signet",
+        Network::Regtest => "Bitcoin Regtest",
+        _ => "Bitcoin unknown",
+    }
+    .to_string()
+}
+
 pub struct Launcher {
     choices: Vec<Network>,
     datadir_path: PathBuf,
     error: Option<String>,
+    delete_wallet_modal: Option<DeleteWalletModal>,
 }
 
 impl Launcher {
@@ -35,6 +51,7 @@ impl Launcher {
             datadir_path,
             choices,
             error: None,
+            delete_wallet_modal: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


darosior:
Played a bit with this. Using a trash icon is fine by me.

The presentation isn't the prettiest but fine, clear and usable:
image

The modal however i think could be improved. Currently:
image
I tried to align the content to the center (as i think in all other modals?) but i'm not sure it really improves things:
image
I'd like to grey-out the line about the managed bitcoind or convey it's less important information. Also i think the WARNING line should be last. Here is my best attempt so far:
image

What do you think? Here is my diff in case you want to take it / improve over:
```diff
diff --git a/gui/src/launcher.rs b/gui/src/launcher.rs
index 8ef01fc..70b3c7c 100644
--- a/gui/src/launcher.rs
+++ b/gui/src/launcher.rs
@@ -276,8 +276,8 @@ impl DeleteWalletModal {
             "Are you sure you want to delete the wallet and all associated data for {}?",
             wallet_name(&self.network)
         );
-        let help_text_2 = "WARNING: This cannot be undone.";
-        let help_text_3 = "If you are using a Liana-managed Bitcoin node, it will not be affected by this action.";
+        let help_text_2 = "(If you are using a Liana-managed Bitcoin node, it will not be affected by this action.)";
+        let help_text_3 = "WARNING: This cannot be undone.";
 
         Into::<Element<ViewMessage>>::into(
             card::simple(
@@ -289,7 +289,8 @@ impl DeleteWalletModal {
                             .width(Length::Fill),
                     ))
                     .push(Row::new().push(text(help_text_1)))
-                    .push(Row::new().push(text(help_text_2)))
+                    .push(Row::new().push(caption(help_text_2).style(color::GREY_3).small()))
+                    .push(Row::new())
                     .push(Row::new().push(text(help_text_3)))
                     .push_maybe(self.warning.as_ref().map(|w| {
                         notification::warning(w.to_string(), w.to_string()).width(Length::Fill)
@@ -302,9 +303,12 @@ impl DeleteWalletModal {
                             .align_items(Alignment::Center)
                             .push(icon::circle_check_icon().style(color::GREEN))
                             .push(text("Wallet successfully deleted").style(color::GREEN))
-                    }),
+                    })
+                    .align_items(Alignment::Center),
             )
-            .width(Length::Fixed(700.0)),
+            .width(Length::Fixed(700.0))
+            .center_x()
+            .center_y(),
         )
         .map(Message::View)
     }
```

gui/src/launcher.rs
```diff
@@ -164,6 +225,89 @@ pub enum Message {
...
+            wallet_datadir,
+            warning: None,
+            deleted: false,
+        }
+    }
+
+    fn update(&mut self, message: Message) -> Command<Message> {
+        if let Message::View(ViewMessage::DeleteWallet(DeleteWalletMessage::Confirm)) = message {
+            self.warning = None;
+            if let Err(e) = std::fs::remove_dir_all(&self.wallet_datadir) {
```
darosior, line 258: This would remove the bitcoind watchonly wallet file. Is it an issue? Depends. It's certainly not the best to be removing a file managed by a process from under it. On the other hand it's unlikely to make bitcoind crash, and even if it did i don't think it would be worse than any other alternative i can think of for handling this.

Therefore i think it's fine like that, but wanted to share in case other think about something i would have missed.

darosior (commented):
Tested ACK aside from my artistic sensibilities.

gui/src/launcher.rs
jp1ac4: Would it be worth adding a message relating to this in the modal (if they are not using Liana-managed node for that wallet)?

jp1ac4:
> I tried to align the content to the center (as i think in all other modals?) but i'm not sure it really improves things

Perhaps we could keep left-alignment for the text and then center for the button, which is how the installer modal for editing keys is.

> I'd like to grey-out the line about the managed bitcoind or convey it's less important information.

It makes sense to grey it out. I could also check whether that wallet is using a Liana-managed node and only show the message in that case.

jp1ac4:
One other thing that might be nice would be to see the wallet be deleted in the blurred background while the modal is open. Currently, it's only removed from the list when the user closes the modal.

gui/src/launcher.rs
darosior: I don't think so.

gui/src/launcher.rs
```diff
@@ -96,31 +137,42 @@ impl Launcher {
...
-                                                        .align_items(Alignment::Center)
-                                                        .push(
-                                                            badge::Badge::new(icon::bitcoin_icon())
-                                                                .style(match choice {
-                                                                    Network::Bitcoin => {
-                                                                        theme::Badge::Bitcoin
-                                                                    }
-                                                                    _ => theme::Badge::Standard,
-                                                                }),
+                                                Row::new()
```
edouardparis, line 140: add some spacing

jp1ac4:
I've made changes based on the feedback so far. I've kept them as separate commits for now in case we only want some of them, but I'll then squash them.

In https://github.com/wizardsardine/liana/pull/864/commits/91d4c9a385f2ac7cda03071b859c56de55e74cda:
- added spacing between wallet button and trash icon
- changed wording order and formatting of text (I used `p1_regular` for managed bitcoind message as it should be equivalent to `caption().small()`)
- changed alignment of "Delete wallet" button to centre

In https://github.com/wizardsardine/liana/pull/864/commits/8b8300a2136465f5b9a65ae26615c72b6507782f, we now refresh the list of available wallets in the blurred background while the modal is still open.

In https://github.com/wizardsardine/liana/pull/864/commits/c32f4b9d040eec881e57e253394d2d3175d061d3, I check whether the wallet uses managed bitcoind. There are now three different cases:

1) Wallet uses managed bitcoind:
image

2) Wallet does not use managed bitcoind:
image

3) We are not sure if wallet uses managed bitcoind (error reading config file):
image

darosior:
This looks great to me. From some light testing it's really slick. Will let @edouardparis have a look at the code before merging.

Slightly-tested-not-reviewed ACK c32f4b9d040eec881e57e253394d2d3175d061d3

edouardparis:
ACK c32f4b9d040eec881e57e253394d2d3175d061d3

jp1ac4:
I'll squash commits now.

edouardparis:
ACK 279c1c75a1dfe6617798b59cbe3932c803aeb90f

darosior:
Great work.

PR #865, opened 2023-12-11
Expose spend module and coins fields

edouardparis:
In order to externally use create_spend, the spend module must be exposed and the listcoinsentry must include the derivation_index and is_change fields.


darosior (commented):
Makes sense to expose those fields. Need to document them in doc/API.md though.

PR #867, opened 2023-12-11
spend: a temporary partial fix for `LowestFee`

jp1ac4:
This is a temporary partial fix for https://github.com/bitcoindevkit/coin-select/issues/6 that should be reverted once the upstream fix has been made.

When calculating the score, the excess should be added to changeless solutions instead of those with change.

Given a solution has been found, this fix adds or removes the excess to its incorrectly calculated score as required so that two changeless solutions can be differentiated if one has higher excess (and therefore pays a higher fee).

Note that the `bound` function is also affected by this bug, which could mean some branches are not considered when running BnB, but at least this fix will mean the score for those solutions that are found is correct.

The new functional test included here fails without this fix.

### DIFF
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -201,7 +201,27 @@ where
         if drain.is_none() && self.must_have_change {
             None
         } else {
-            self.lowest_fee.score(cs)
+            // This is a temporary partial fix for https://github.com/bitcoindevkit/coin-select/issues/6
+            // until it has been fixed upstream.
+            // TODO: Revert this change once upstream fix has been made.
+            // When calculating the score, the excess should be added to changeless solutions instead of
+            // those with change.
+            // Given a solution has been found, this fix adds or removes the excess to its incorrectly
+            // calculated score as required so that two changeless solutions can be differentiated
+            // if one has higher excess (and therefore pays a higher fee).
+            // Note that the `bound` function is also affected by this bug, which could mean some branches
+            // are not considered when running BnB, but at least this fix will mean the score for those
+            // solutions that are found is correct.
+            self.lowest_fee.score(cs).map(|score| {
+                // See https://github.com/bitcoindevkit/coin-select/blob/29b187f5509a01ba125a0354f6711e317bb5522a/src/metrics/lowest_fee.rs#L35-L45
+                assert!(cs.selected_value() >= self.lowest_fee.target.value);
+                let excess = (cs.selected_value() - self.lowest_fee.target.value) as f32;
+                bdk_coin_select::float::Ordf32(if drain.is_none() {
+                    score.0 + excess
+                } else {
+                    score.0 - excess
+                })
+            })
         }
     }
 
--- a/tests/test_spend.py
+++ b/tests/test_spend.py
@@ -1,5 +1,5 @@
 from fixtures import *
-from test_framework.serializations import PSBT
+from test_framework.serializations import PSBT, uint256_from_str
 from test_framework.utils import wait_for, COIN, RpcError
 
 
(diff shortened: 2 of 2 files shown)


tests/test_spend.py
```diff
@@ -332,6 +335,24 @@ def test_coin_selection(lianad, bitcoind):
...
+    wait_for(lambda: len(lianad.rpc.listcoins(["confirmed"])["coins"]) == 2)
+    # Send an amount that can be paid by just one of our coins.
+    res = lianad.rpc.createspend({bitcoind.rpc.getnewaddress(): 30800}, [], 1)
+    psbt = PSBT.from_base64(res["psbt"])
+    # Only one input needed.
+    assert len(psbt.i) == 1
+    # Coin A is used as input.
+    assert [psbt_in.map[PSBT_IN_NON_WITNESS_UTXO] for psbt_in in psbt.i][
+        0
+    ] == bytes.fromhex(bitcoind.rpc.gettransaction(txid_a)["hex"])
```
darosior, line 353: Hmm. No need to iterate. And you can also simply check the txid without having to check the whole transaction with an RPC call:
```diff
diff --git a/tests/test_spend.py b/tests/test_spend.py
index d735d1e..94ba1a5 100644
--- a/tests/test_spend.py
+++ b/tests/test_spend.py
@@ -1,8 +1,5 @@
 from fixtures import *
-from test_framework.serializations import (
-    PSBT,
-    PSBT_IN_NON_WITNESS_UTXO,
-)
+from test_framework.serializations import PSBT, uint256_from_str
 from test_framework.utils import wait_for, COIN, RpcError
 
 
@@ -348,9 +345,8 @@ def test_coin_selection_changeless(lianad, bitcoind):
     # Only one input needed.
     assert len(psbt.i) == 1
     # Coin A is used as input.
-    assert [psbt_in.map[PSBT_IN_NON_WITNESS_UTXO] for psbt_in in psbt.i][
-        0
-    ] == bytes.fromhex(bitcoind.rpc.gettransaction(txid_a)["hex"])
+    txid_a = uint256_from_str(bytes.fromhex(txid_a)[::-1])
+    assert psbt.tx.vin[0].prevout.hash == txid_a
 
 
 def test_sweep(lianad, bitcoind):
```

darosior (commented):
The comment makes sense but i feel i would have to dig into the CS implementation to be able to meaningfully review this. I'll ask Evan if he can have a quick look at your hotfix.

tests/test_spend.py
jp1ac4: Thanks, I've changed this now.

src/spend.rs
```diff
@@ -201,7 +201,27 @@ where
...
+            // solutions that are found is correct.
+            self.lowest_fee.score(cs).map(|score| {
+                // See https://github.com/bitcoindevkit/coin-select/blob/29b187f5509a01ba125a0354f6711e317bb5522a/src/metrics/lowest_fee.rs#L35-L45
+                assert!(cs.selected_value() >= self.lowest_fee.target.value);
+                let excess = (cs.selected_value() - self.lowest_fee.target.value) as f32;
+                bdk_coin_select::float::Ordf32(if drain.is_none() {
+                    score.0 + excess
+                } else {
+                    score.0 - excess
+                })
```
darosior, line 223: So the bug is that in case there is a change output, it will decrease the score by the value of the excess:
https://github.com/bitcoindevkit/coin-select/blob/29b187f5509a01ba125a0354f6711e317bb5522a/src/metrics/lowest_fee.rs#L38C19-L42. But when there is none it won't be changing the score: https://github.com/bitcoindevkit/coin-select/blob/29b187f5509a01ba125a0354f6711e317bb5522a/src/metrics/lowest_fee.rs#L43.

But here you not only increase it to compensate the incorrect decrease in case of no-change, you also decrease it in case of change. Shouldn't we only be compensating in case there was no change? Like:
```suggestion
                bdk_coin_select::float::Ordf32(if drain.is_none() {
                    score.0 + excess
                } else {
                    score.0
                })
```

jp1ac4: I think the intention of the `LowestFee` metric is that the excess should only be added to the score for changeless solutions (see also [L6-L13](https://github.com/bitcoindevkit/coin-select/blob/29b187f5509a01ba125a0354f6711e317bb5522a/src/metrics/lowest_fee.rs#L6-L13)). As it is currently adding the excess to the solution with change, I subtract it here to undo that, so that in the end the excess is only being added to changeless solutions.

We may want to add something to the score for solutions with change in order to give preference to smaller or larger change amounts, but I think that would be beyond the current intention of the `LowestFee` metric.

darosior: Oh, yeah, got that backward. It's adding the excess to the solution *with* change, and you undo it here. But it doesn't remove anything from the solution without change, so why do you add the excess to the solution without change here?

The diff i should have suggested above is:
```suggestion
                bdk_coin_select::float::Ordf32(if drain.is_none() {
                    score.0
                } else {
                    score.0 - excess
                })
```

jp1ac4: As the excess isn't currently being added to the changeless score in the upstream, I need to add it here.

darosior: My bad, i'm slow!

darosior:
ACK 5f534eb988b903c28b8c2edb1db9a6389aadfdc0

PR #869, opened 2023-12-12
gui: set max feerate in RBF modal

jp1ac4:
This applies a similar change to feerate form validation as in #863 to ensure value is not too high.

image


edouardparis (approved):
utACK dfc4342d44cc2a3bcc080050fa636c5db78a2768

PR #878, opened 2023-12-14
Backport #879 : fix bitbox unsupported status when no wallet yet

edouardparis:
During the install process, no wallet yet is created and bitbox should not be marked as unsupported if
its fingerprint is not in any descriptor.


edouardparis (commented):
Self-ACK ed19a9836f4df6a1b7812fc5f770cf41b2505bdd
backport of #879

PR #879, opened 2023-12-14
fix bitbox unsupported status when no wallet yet

edouardparis:
During the install process, no wallet yet is created and bitbox should not be marked as unsupported if
its fingerprint is not in any descriptor.


edouardparis (commented):
Self-ACK 1fa414be75cc99b4d95d6a7f3048cbfe62bdeb9b

PR #882, opened 2023-12-14
gui: use descriptor helper function to get unsigned tx max size

### LINKED ISSUE #880: GUI: Estimated feerate displayed for PSBTs can be too low
The PSBT screen shows an estimated feerate based on each transaction input having the max satisfaction size.

This feerate is sometimes underestimated, especially when there are several inputs, as in this example showing an estimated feerate of 0 sats/vb:
image

In a similar way to #734, the total weight of the transaction should first be calculated before being converted to vbytes in order to avoid rounding issues.

This issue only concerns the estimated feerate displayed and does not affect how the transaction is created.

jp1ac4:
This is to resolve #880 using the new helper function from #881 to get the max possible size of the unsigned transaction.

The following screenshot is taken using both new and old implementations together and shows the new estimated feerate changes from 0 to 1 sat/vbyte:
image

### DIFF
--- a/gui/src/daemon/model.rs
+++ b/gui/src/daemon/model.rs
@@ -38,9 +38,9 @@ pub struct SpendTx {
     pub change_indexes: Vec<usize>,
     pub spend_amount: Amount,
     pub fee_amount: Option<Amount>,
-    /// The maximum size difference (in virtual bytes) of
-    /// an input in this transaction before and after satisfaction.
-    pub max_sat_vbytes: usize,
+    /// Maximum possible size of the unsigned transaction after satisfaction
+    /// (assuming all inputs are for the same descriptor).
+    pub max_vbytes: u64,
     pub status: SpendStatus,
     pub sigs: PartialSpendInfo,
     pub updated_at: Option<u32>,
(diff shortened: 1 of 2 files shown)


jp1ac4:
I'll leave this as draft until #881 has been merged.

gui/src/daemon/model.rs
```diff
@@ -203,10 +203,11 @@ impl SpendTx {
...
     /// Feerate obtained if all transaction inputs have the maximum satisfaction size.
     pub fn min_feerate_vb(&self) -> Option<u64> {
-        // This assumes all inputs are internal (have same max satisfaction size).
-        let max_tx_vbytes =
-            self.psbt.unsigned_tx.vsize() + (self.max_sat_vbytes * self.psbt.inputs.len());
-        self.fee_amount.map(|a| a.to_sat() / max_tx_vbytes as u64)
+        self.fee_amount.map(|a| {
+            a.to_sat()
+                .checked_div(self.max_vbytes)
+                .expect("max_vbytes should be greater than zero")
```
darosior, line 209: nit: Hopefully, yes! But why? I think that's the info which should go here. For instance "a PSBT with an empty unsigned_tx would never deserialize therefore this is never 0" or simply "a descriptor's satisfaction size is never 0".

darosior (commented):
Nice! Just need to make the GUI point back to master.

gui/src/daemon/model.rs
jp1ac4: I've now changed this to the second suggestion.

jp1ac4:
The GUI is now pointing to the latest master.

darosior:
ACK e373ad4

Issue #356, opened 2023-02-23
Support for `rpcuser` / `rpcpassword` instead of cookie file auth for bitcoind RPC


darosior:
It was requested to me several times. I'd like to centralize the discussion in this issue.

Typically the situation where it could make sense is: you have a `bitcoind` running on your local network, and you'd rather reuse it instead of running another one on your laptop alongside Liana.

This is reasonable, however i think the usefulness of such a setup is overstated mainly due to 2 confusions:
1. It is *trivial* and *lightweight* to run bitcoind on your laptop. People still see `bitcoind` as a heavy process for some reason. You can trivially run a pruned bitcoind on your laptop. Once you've passed the IBD you'll barely notice it even on a low end machine. If you are not too concern about the bandwidth it's also the KISS solution to needing a bitcoind, instead of trying to over-optimize your setup so that all your process that need a bitcoind connect to a single box. 
2. The RPC interface is *not supposed to be accessed remotely*. Sure, it *might* be reasonable to connect to a box running bitcoind on your LAN. But then what, when you want to use your Liana wallet from outside your LAN you start doing SSH tunnels?... Meh. Back to step 1. It's over-optimization, just run `bitcoind` locally on your laptop and save yourself the configuration time and brain process.

Finally, sure, Electrum is an alternative. See https://github.com/wizardsardine/liana/issues/56. Ideally we'd have something like https://github.com/wizardsardine/liana/issues/101 and everybody wouldn't even notice they are running a full node locally.

vincenzopalazzo:
Ok, your approach makes total sense to me.

I also need to access my remote machine with `user` and `password`, but I can run a bitcoind and connect only to my remote instances as well.

However, the syncing time can be a problem, so it would be great to have a fallback backend that allows me to work with the wallet and switch to Bitcoin when it is ready.

For the core lightning plugin, I have designed it in a way that when the bitcoind are in pruning mode or syncing, it will fetch the block from the network directly or with the nakamoto implementation (a fork of it). I have implemented this design in the following repository: https://github.com/coffee-tools/folgore.

It would also be helpful to have a similar design with liana and use a pure bitcoind backend, along with another pure Bitcoin backend for faster startup and not have electrum-like solution.

darosior:
> However, the syncing time can be a problem, so it would be great to have a fallback backend that allows me to work with the wallet and switch to Bitcoin when it is ready.

[...]

PR #890, opened 2023-12-19
config: add user:password option for RPC authentication

jp1ac4:
As a first step towards #356, this adds the option to specify a user and password in Liana's config file for bitcoind RPC authentication.

The GUI installer & settings will need to be updated in a follow-up PR to add this option.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,6 +35,43 @@ pub fn serialize_duration<S: Serializer>(duration: &Duration, s: S) -> Result<S:
     s.serialize_u64(duration.as_secs())
 }
 
+fn deserialize_rpc_auth<'de, D>(deserializer: D) -> Result<BitcoindRpcAuth, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    pub struct BitcoindRpcAuthHelper {
+        cookie_path: Option<PathBuf>,
+        auth: Option<String>,
+    }
+    let BitcoindRpcAuthHelper { cookie_path, auth } =
+        BitcoindRpcAuthHelper::deserialize(deserializer)?;
+    let rpc_auth = match (cookie_path, auth) {
+        (Some(_), Some(_)) => {
+            return Err(de::Error::custom(
+                "must not set both `cookie_path` and `auth`",
+            ));
+        }
+        (Some(path), None) => BitcoindRpcAuth::CookieFile(path),
+        (None, Some(auth)) => auth
+            .split_once(':')
+            .ok_or(de::Error::custom("`auth` must be 'user:password'"))
+            .map(|(user, pass)| BitcoindRpcAuth::UserPass(user.to_string(), pass.to_string()))?,
+        (None, None) => {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -291,17 +292,24 @@ fn setup_bitcoind(
         // test this codepath in a functional test.
         log::info!(
             "A data directory exists with no watchonly wallet. This is most likely due to \
-             having been created to an older version of Liana on Windows, where the watchonly \
-             wallet would live under bitcoind's datadir. Trying to find the older wallet and copy it in our own datadir."
+            having been created to an older version of Liana on Windows, where the watchonly \
+            wallet would live under bitcoind's datadir. Trying to find the older wallet and copy it in our own datadir. \
+            Make sure to set the cookie_path option in your config file in order for this operation to succeed."
```
darosior, line 297: Hmm. We should just get rid of this monstrosity.

darosior (commented):
I think this can be simplified a little bit

src/lib.rs
darosior: Done in https://github.com/wizardsardine/liana/pull/892.

darosior:
@edouardparis can you have a look at the GUI CI error? Looks like a rust nightly bug, right?

src/config.rs
```diff
@@ -48,15 +48,127 @@ fn default_daemon() -> bool {
...
 pub struct BitcoindConfig {
-    /// Path to bitcoind's cookie file, to authenticate the RPC connection
-    pub cookie_path: PathBuf,
+    /// Authentication credentials for bitcoind's RPC server.
+    pub rpc_auth: BitcoindRpcAuth,
     /// The IP:port bitcoind's RPC is listening on
     pub addr: SocketAddr,
 }
 
+impl Serialize for BitcoindConfig {
```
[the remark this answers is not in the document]
jp1ac4: Thanks for this, much simpler :relieved: I hadn't considered using the `#[serde(flatten)]` attribute.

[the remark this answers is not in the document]
jp1ac4: I've now changed it to use serde attributes. The deserializer is slightly different so that it returns an error if both "cookie_path" and "auth" are set in the config file.

src/config.rs
```diff
@@ -35,6 +35,43 @@ pub fn serialize_duration<S: Serializer>(duration: &Duration, s: S) -> Result<S:
...
+    D: Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    pub struct BitcoindRpcAuthHelper {
+        cookie_path: Option<PathBuf>,
+        auth: Option<String>,
+    }
+    let BitcoindRpcAuthHelper { cookie_path, auth } =
+        BitcoindRpcAuthHelper::deserialize(deserializer)?;
+    let rpc_auth = match (cookie_path, auth) {
```
darosior, line 49: Neat.

darosior (commented):
ACK 552a49edb2dbccda5204f80188212fc6b59e029e

PR #892, opened 2023-12-20
Drop watchonly wallet migration for old Windows datadir

darosior:
It was introduced as a temporary hack in v2 (released in August 2023) for migrating datadirs created on Windows using Liana v1 (May 2023).

Presumably Windows users wouldn't have used the beta versions which preceded v1 release with any real funds. This removal would go in v5, released around end of January / early February. Presumably any Windows user who created a mainnet datadir within the 3 months period where the bug was present would have opened its Liana at least once in the following 6 months after the migration was released.

In the unlikely case a user on Windows created a mainnet wallet on the very first release of Liana, funded it, didn't touch it for more than 6 months and tried to open it later using the latest version of Liana available they are instructed to first run Liana v4 (so the migration is performed automatically at startup) before running the latest version of Liana.

Nits on the wording welcome!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,8 @@ pub enum StartupError {
     Bitcoind(BitcoindError),
     #[cfg(unix)]
     Daemonization(&'static str),
+    #[cfg(windows)]
+    NoWatchonlyInDatadir,
 }
 
 impl fmt::Display for StartupError {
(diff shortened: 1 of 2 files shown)


darosior:
The GUI ci failure looks unrelated.

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/892/commits/6d2833fc1857daf140f7efc8d0b81c662076edd4.

Wording looks good to me.

PR #894, opened 2023-12-21
gui: truncate config file before writing

jp1ac4:
This is required in case the updated config file has a shorter length than the existing file.

For example, if changing cookie path from absolute path to relative path, the updated config file will have invalid content without this change.

### DIFF
--- a/gui/src/app/mod.rs
+++ b/gui/src/app/mod.rs
@@ -202,22 +202,19 @@ impl App {
         let daemon = EmbeddedDaemon::start(cfg)?;
         self.daemon = Arc::new(daemon);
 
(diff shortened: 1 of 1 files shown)


darosior (commented):
Makes sense but i'm missing context surrounding this code. Github utACK a4b982d972b9a5d823b48ae1f1b979481e6785b6. Will let Edouard confirm it's OK.

darosior:
Well i'll just go ahead with this one as we start piling PRs depending on it. @edouardparis please do still have a look whenever you can.

edouardparis:
Thanks for the ping, I read and understood to need for the change. Will test it.

PR #898, opened 2023-12-28
gui: support user/password for RPC authentication

jp1ac4:
This adds support to the GUI for user and password RPC authentication as part of #356.

The first commit adds the user/password option to the settings page, and only updates the installer as required for it to compile. It also changes how the managed bitcoind gets the cookie location when starting so that it doesn't rely on the config file.

The second commit adds the user/password option to the installer when using a self-managed node.

Updating the managed node to use user/password can be done in a follow-up PR.

### DIFF
--- a/gui/src/app/view/settings.rs
+++ b/gui/src/app/view/settings.rs
@@ -3,11 +3,14 @@ use std::str::FromStr;
 
 use iced::{
     alignment,
-    widget::{scrollable, Space},
+    widget::{radio, scrollable, Space},
     Alignment, Length,
 };
 
-use liana::miniscript::bitcoin::{bip32::Fingerprint, Network};
+use liana::{
+    config::BitcoindRpcAuth,
+    miniscript::bitcoin::{bip32::Fingerprint, Network},
+};
 
 use super::{dashboard, message::*};
 
@@ -26,6 +29,7 @@ use crate::{
         menu::Menu,
         view::{hw, warning::warn},
     },
+    bitcoind::{RpcAuthType, RpcAuthValues},
     hw::HardwareWallet,
 };
 
@@ -208,7 +212,8 @@ pub fn bitcoind_edit<'a>(
     network: Network,
     blockheight: i32,
     addr: &form::Value<String>,
-    cookie_path: &form::Value<String>,
+    rpc_auth_vals: &RpcAuthValues,
+    selected_auth_type: &RpcAuthType,
     processing: bool,
 ) -> Element<'a, SettingsEditMessage> {
     let mut col = Column::new().spacing(20);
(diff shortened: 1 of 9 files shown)


gui/src/app/view/settings.rs
```diff
@@ -244,18 +249,59 @@ pub fn bitcoind_edit<'a>(
...
+                        .spacing(10),
+                    |row, auth_type| {
+                        row.push(radio(
+                            format!("{}", auth_type),
+                            *auth_type,
+                            Some(*selected_auth_type),
+                            SettingsEditMessage::BitcoindRpcAuthTypeSelected,
+                        ))
+                        .spacing(30)
+                        .align_items(Alignment::Center)
```
darosior, line 266: Nice, i like the radio button. Just let's not make them green?

darosior: Looking through the rest of the interface most buttons do get green when hovered though.

gui/src/app/view/settings.rs
```diff
@@ -244,18 +249,59 @@ pub fn bitcoind_edit<'a>(
...
+                            .padding(5),
+                        )
+                        .push(
+                            form::Form::new_trimmed("Password", &rpc_auth_vals.password, |value| {
+                                SettingsEditMessage::FieldEdited("password", value)
+                            })
+                            .warning("Please enter a valid password")
+                            .size(20)
+                            .padding(5),
+                        ),
```
darosior, line 301: nit: can we add a small marging in-between the two text inputs?

gui/src/installer/view.rs
```diff
@@ -787,39 +788,93 @@ pub fn help_backup<'a>() -> Element<'a, Message> {
...
+                            |new_selection| {
+                                Message::DefineBitcoind(
+                                    message::DefineBitcoind::RpcAuthTypeSelected(new_selection),
+                                )
+                            },
+                        ))
+                        .spacing(30)
+                        .align_items(Alignment::Center)
+                    },
+                ),
```
darosior, line 832: Same comment regarding the radio buttons here:
image


They should have the same style as the other buttons around, i think.

darosior (commented):
Looks good from some quick testing!

gui/src/app/view/settings.rs
jp1ac4: I've now added `spacing(10)` between them in both settings and installer.

jp1ac4:
In a separate commit, I've modified the radio button theme to only have the green border when hovering over it. The selected button still uses a green dot, which is similar to a checkbox that has been ticked.

image

I'm happy to make further changes to the radio button theme as part of this PR.

darosior:
ACK 838550b917d96a6a6d11f5d0103c53706acd9a8c

PR #899, opened 2023-12-28
gui(settings): load updated bitcoind settings after saving

jp1ac4:
Currently, the Bitcoin Core settings page doesn't show the updated values after saving changes and needs to be reloaded in order to refresh. This change makes the Bitcoin Core settings page reload after changes are saved so that they are shown immediately.


edouardparis:
ACK bd9d9f2168fcc2cac965f40a561e550752ff33e6

edouardparis (approved):
ACK bd9d9f2168fcc2cac965f40a561e550752ff33e6

Issue #811, opened 2023-11-14
Warn when adding the change output's value to fees.


darosior:
When the change output value is below our own "dust" limit (5k sats at the moment, ie 1.5€) we do not include it at all thereby adding the value to fees.

We should display a warning when we do so. Maybe through a `"warnings"` entry in the result of `createspend`?

pythcoiner:
related to #790

darosior:
Half of this was merged with #905. What's left now is to display the warnings in the GUI.

jp1ac4:
I think we can display this warning, and the one from https://github.com/wizardsardine/liana/pull/873, when the user has generated a PSBT, but before saving it.

We can also consider whether to add the warning about change added to fee when the user creates a new RBF spend. We could display any such warnings in the RBF modal, but it might be more intuitive to link to the draft PSBT screen from the RBF modal and let the user view the warnings there. Currently, the PSBT is saved automatically from the RBF modal and we link to the saved PSBT.

We could also display these warnings on a saved PSBT screen, but in general we won't have access to these warnings after the PSBT is saved so I think it makes more sense to display them for draft PSBTs only.

For now, I could proceed on the assumption we'll display these warnings in the draft PSBT screen only and we can decide later whether to change the RBF modal.

PR #905, opened 2024-01-03
spend: warn when overpaying fee

jp1ac4:
This is a first step towards #811. The GUI will be updated in a follow-up PR.

It adds a warning if there is any change/excess available that has been added to the fee, building on the assumption from https://github.com/bitcoindevkit/coin-select/pull/14 that the decision whether to include change depends on whether the excess is over a threshold. The function `select_coins_for_spend()` now returns this threshold so that the caller can check if the possible change amount is above it.

I've used an `enum` for the different warnings, but could use strings directly.

If the approach looks fine, I'll add some tests.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -179,9 +179,10 @@ This command will refuse to create any output worth less than 5k sats.
 
 #### Response
 
-| Field          | Type      | Description                                          |
-| -------------- | --------- | ---------------------------------------------------- |
-| `psbt`         | string    | PSBT of the spending transaction, encoded as base64. |
+| Field          | Type              | Description                                          |
+| -------------- | ----------------- | ---------------------------------------------------- |
+| `psbt`         | string            | PSBT of the spending transaction, encoded as base64. |
+| `warnings`     | list of string    | Warnings, if any, generated during spend creation.   |
 
 
 ### `updatespend`
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -490,7 +490,9 @@ impl DaemonControl {
         // derivation index in case any address in the transaction outputs was ours and from the
         // future.
         let change_info = change_address.info;
-        let CreateSpendRes { psbt, has_change } = create_spend(
+        let CreateSpendRes {
+            psbt, has_change, ..
+        } = create_spend(
             &self.config.main_descriptor,
             &self.secp,
             &mut tx_getter,
@@ -810,6 +812,7 @@ impl DaemonControl {
             let CreateSpendRes {
                 psbt: rbf_psbt,
                 has_change,
+                ..
             } = match create_spend(
                 &self.config.main_descriptor,
                 &self.secp,
@@ -985,7 +988,9 @@ impl DaemonControl {
         }
 
         let sweep_addr_info = sweep_addr.info;
-        let CreateSpendRes { psbt, has_change } = create_spend(
+        let CreateSpendRes {
+            psbt, has_change, ..
+        } = create_spend(
             &self.config.main_descriptor,
             &self.secp,
             &mut tx_getter,
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -491,7 +491,9 @@ impl DaemonControl {
         // future.
         let change_info = change_address.info;
         let CreateSpendRes {
-            psbt, has_change, ..
+            psbt,
+            has_change,
+            warnings,
         } = create_spend(
             &self.config.main_descriptor,
             &self.secp,
@@ -508,7 +510,10 @@ impl DaemonControl {
             self.maybe_increase_next_deriv_index(&mut db_conn, &change_info);
         }
 
-        Ok(CreateSpendResult { psbt })
+        Ok(CreateSpendResult {
+            psbt,
+            warnings: warnings.iter().map(|w| w.to_string()).collect(),
+        })
     }
 
     pub fn update_spend(&self, mut psbt: Psbt) -> Result<(), CommandError> {
@@ -812,7 +817,7 @@ impl DaemonControl {
             let CreateSpendRes {
                 psbt: rbf_psbt,
                 has_change,
-                ..
+                warnings,
             } = match create_spend(
                 &self.config.main_descriptor,
                 &self.secp,
@@ -857,7 +862,10 @@ impl DaemonControl {
                     self.maybe_increase_next_deriv_index(&mut db_conn, &change_address.info);
                 }
 
-                return Ok(CreateSpendResult { psbt: rbf_psbt });
+                return Ok(CreateSpendResult {
+                    psbt: rbf_psbt,
+                    warnings: warnings.iter().map(|w| w.to_string()).collect(),
+                });
             }
         }
 
@@ -1103,6 +1111,7 @@ pub struct ListCoinsResult {
 pub struct CreateSpendResult {
     #[serde(serialize_with = "ser_to_string", deserialize_with = "deser_fromstr")]
     pub psbt: Psbt,
+    pub warnings: Vec<String>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -1141,6 +1150,7 @@ mod tests {
     use super::*;
     use crate::{bitcoin::Block, database::BlockInfo, spend::InsaneFeeInfo, testutils::*};
 
+    use bdk_coin_select::InsufficientFunds;
     use bitcoin::{
         bip32::{self, ChildNumber},
         blockdata::transaction::{TxIn, TxOut, Version as TxVersion},
@@ -1374,6 +1384,8 @@ mod tests {
         assert_eq!(tx.input.len(), 1);
         assert_eq!(tx.input[0].previous_output, dummy_op);
         assert_eq!(tx.output.len(), 2);
+        // It has change so no warnings expected.
+        assert!(res.warnings.is_empty());
         assert_eq!(
             tx.output[0].script_pubkey,
             dummy_addr.payload().script_pubkey()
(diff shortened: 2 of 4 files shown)


src/spend.rs
```diff
@@ -172,6 +172,29 @@ pub struct CandidateCoin {
...
+    /// A non-zero value implies that either a possible change output has not been added
+    /// by the change policy or otherwise that the excess amount is too small to pay
+    /// for a change output.
+    pub excess: bitcoin::Amount,
+    /// Change amount that could be included for this selection while still respecting the
+    /// target feerate, irrespective of the change policy used for selection.
+    pub change_amount: bitcoin::Amount,
+    /// Minimum change amount required in order for a change output to be added
+    /// to the transaction, as determined by the change policy used for this selection.
+    pub change_policy_min_amount: bitcoin::Amount,
```
darosior, line 196: Why can't we just have:
- `excess`: the amount overpaid, with or without change
- `change_amount`: the change amount, if non-zero

It seems to me that's all we need? If `change_amount` is 0 and `excess` is >0 (or to a given threshold?) emit a warning.

src/spend.rs
```diff
@@ -462,11 +462,37 @@ pub enum SpendTxFees {
     Rbf(u64, u64),
 }
 
+pub enum CreateSpendWarning {
+    /// A possible change output has not been included.
+    ChangeNotIncluded(u64),
+    /// Excess is insufficient to pay for a change output.
+    ExcessTooSmallForChange(u64),
```
darosior, line 471: As a consequence of my previous comment, this would be a single variant. I think it would be a good thing, it's really unclear what the difference between the two is here.

src/commands/mod.rs
```diff
@@ -1099,10 +1104,11 @@ pub struct ListCoinsResult {
     pub coins: Vec<ListCoinsEntry>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
darosior, line 1107: Instead of removing them here, add them to `CreateSpendWarning`.

darosior (commented):
Approach ACK for the interface, i'm just unsure we need more than just the excess value and change amount.

src/spend.rs
jp1ac4: Hmm yes I think we could just include the `excess` taking into account any change added according to the policy and keep the `change_amount` as the amount of change according to the policy, and remove `change_policy_min_amount` (the caller could then continue to use `change_amount` directly without needing to check the threshold). This should be clearer as then all values would be according to the change policy.

Then the `excess` would be the amount being overpaid with respect to the target feerate, which is what we want. The possible amount of change in the absence of any min threshold would be smaller than the excess as we'd need to pay for the change output, but I think the important thing is to know how much is being overpaid rather than how much change there might have been if a change output had been included.

src/commands/mod.rs
jp1ac4: Thanks, I've done that now. Something before made me think it wasn't working, but it does :relieved:

src/spend.rs
jp1ac4: I've now combined them into a single `FeeOverpayment`.

src/spend.rs
jp1ac4: This has now been done.

doc/API.md
```diff
@@ -182,6 +182,7 @@ This command will refuse to create any output worth less than 5k sats.
 | Field          | Type      | Description                                          |
 | -------------- | --------- | ---------------------------------------------------- |
 | `psbt`         | string    | PSBT of the spending transaction, encoded as base64. |
+| `warnings`     | list      | Warnings, if any, generated during spend creation.   |
```
jp1ac4, line 185: Should the `createspend` command return the string `Display` value?

darosior (commented):
ACK. Good unit test. Got a couple comments mostly on documentation stuff. Needs rebase.

It's always good to have a sanity check of the interface as consumed externally from the functional tests:
```diff
diff --git a/tests/test_spend.py b/tests/test_spend.py
index 94ba1a5..446acf7 100644
--- a/tests/test_spend.py
+++ b/tests/test_spend.py
@@ -26,6 +26,9 @@ def test_spend_change(lianad, bitcoind):
     assert len(spend_psbt.o) == 3
     assert len(spend_psbt.tx.vout) == 3
 
+    # Since the transaction contains a change output there is no warning.
+    assert len(res["warnings"]) == 0
+
     # Sign and broadcast this first Spend transaction.
     signed_psbt = lianad.signer.sign_psbt(spend_psbt)
     lianad.rpc.updatespend(signed_psbt.to_base64())
@@ -46,6 +49,7 @@ def test_spend_change(lianad, bitcoind):
     }
     res = lianad.rpc.createspend(destinations, outpoints, 2)
     spend_psbt = PSBT.from_base64(res["psbt"])
+    assert len(res["warnings"]) == 0
 
     # We can sign and broadcast it.
     signed_psbt = lianad.signer.sign_psbt(spend_psbt)
@@ -110,6 +114,7 @@ def test_coin_marked_spent(lianad, bitcoind):
     res = lianad.rpc.createspend(destinations, [outpoint], 6)
     psbt = PSBT.from_base64(res["psbt"])
     sign_and_broadcast(psbt)
+    assert len(res["warnings"]) == 0
 
     # Spend the second coin without a change output
     outpoint = next(
@@ -123,6 +128,8 @@ def test_coin_marked_spent(lianad, bitcoind):
     res = lianad.rpc.createspend(destinations, [outpoint], 1)
     psbt = PSBT.from_base64(res["psbt"])
     sign_and_broadcast(psbt)
+    assert len(res["warnings"]) == 1
+    assert res["warnings"][0] == ""
 
     # Spend the third coin to an address of ours, no change
     outpoints = [
@@ -136,6 +143,8 @@ def test_coin_marked_spent(lianad, bitcoind):
     res = lianad.rpc.createspend(destinations, [outpoints[0]], 1)
     psbt = PSBT.from_base64(res["psbt"])
     sign_and_broadcast(psbt)
+    assert len(res["warnings"]) == 1
+    assert res["warnings"][0] == ""
 
     # Spend the fourth coin to an address of ours, with change
     destinations = {
@@ -144,6 +153,7 @@ def test_coin_marked_spent(lianad, bitcoind):
     res = lianad.rpc.createspend(destinations, [outpoints[1]], 18)
     psbt = PSBT.from_base64(res["psbt"])
     sign_and_broadcast(psbt)
+    assert len(res["warnings"]) == 0
 
     # Batch spend the fourth and fifth coins
     outpoint = next(
@@ -159,6 +169,7 @@ def test_coin_marked_spent(lianad, bitcoind):
     res = lianad.rpc.createspend(destinations, [outpoints[2], outpoint], 2)
     psbt = PSBT.from_base64(res["psbt"])
     sign_and_broadcast(psbt)
+    assert len(res["warnings"]) == 0
 
     # All the spent coins must have been detected as such
     all_deposits = (deposit_a, deposit_b, deposit_c, deposit_d)
```

I've left out the error string assertion as `""` since the current string is going to change.

src/spend.rs
```diff
@@ -172,6 +173,26 @@ pub struct CandidateCoin {
...
+    /// Change amount that should be included according to the change policy used
+    /// for selection.
+    pub change_amount: bitcoin::Amount,
+    /// Amount the selected candidates overpay the amount required to pay all recipients
+    /// at the target feerate, taking into account the change output, if any, as determined
+    /// by the change policy.
+    ///
+    /// A non-zero value implies that either a possible change output has not been added
+    /// by the change policy or otherwise that the excess amount is too small to pay
+    /// for a change output.
```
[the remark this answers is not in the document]
jp1ac4: One is due to the potential change amount being below the min value we set in the change policy, while the other is that there's not enough excess to create a change output even if the min value were 0 (excess less than 43 sats in my tests).

src/spend.rs
```diff
@@ -417,11 +447,30 @@ pub enum SpendTxFees {
...
+pub enum CreateSpendWarning {
+    FeeOverpayment(u64),
+}
+
+impl fmt::Display for CreateSpendWarning {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            CreateSpendWarning::FeeOverpayment(amt) => write!(
+                f,
+                "The transaction fee is {} sats higher than required to achieve the target feerate.",
```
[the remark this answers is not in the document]
jp1ac4: Currently, the excess value returned doesn't take into account the extra fee for creating the change output so would be slightly bigger than the potential change amount. I could return the potential change amount instead, but then we would ignore if there's some excess that is too small to pay the fee for a change output. I suppose for such small amounts (e.g. less than 43 sats), a warning is unnecessary.

doc/API.md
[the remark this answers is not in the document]
jp1ac4: Might it be useful for the GUI to have access to the enum value, e.g. if the GUI needs to parse the warning to extract a value, e.g. to create a link to a transaction using its txid, it would be simpler with the enum.

src/spend.rs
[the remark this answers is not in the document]
darosior: I don't think it's worth differentiating the two. It's always about paying for an output (or spending it).

src/spend.rs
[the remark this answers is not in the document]
darosior: Yes exactly.

doc/API.md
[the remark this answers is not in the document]
darosior: The GUI needs to have access to the same interface when consuming the JSONRPC interface or directly using the Rust functions.

jp1ac4: I've now updated this to return a list of strings.

src/spend.rs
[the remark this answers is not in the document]
jp1ac4: I've now replaced `excess` with `max_change_amount` in the return value.

src/spend.rs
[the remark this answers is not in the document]
jp1ac4: I've renamed the warning to `ChangeAddedToFee` and now the value takes into account the extra fee for creating the change output.

jp1ac4:
I've now made the suggested changes.

src/spend.rs
```diff
@@ -374,14 +390,27 @@ fn select_coins_for_spend(
...
-    Ok((
+    // Max available change is given by the excess when adding a change output with zero value.
+    let drain_novalue = bdk_coin_select::Drain {
+        weights: drain_weights,
+        value: 0,
+    };
+    let max_change_amount = bitcoin::Amount::from_sat(
         selector
+            .excess(target, drain_novalue)
+            .max(0) // negative excess would mean insufficient funds to pay for change output
```
darosior, line 401: So this can only happen if we've got a selection without change such as adding a change output would decrease the feerate below the target? Any other error here would be a logic error because at this point we do indeed have enough funds?

darosior (commented):
ACK 5a15c744e796c9ebe4f6c1eb3010b824189f7855

src/spend.rs
jp1ac4: That's right, the selected coins here have enough value without a change output and so the negative excess would mean that the selected coins don't have enough value to pay the extra weight for the change output at the target feerate.

PR #906, opened 2024-01-04
Bump MSRV, upgrade dependencies and clippification

darosior:
New years' cleanup!

Our MSRV was becoming too restrictive to keep being able to benefit from the latest releases of our dependencies. Strike a better balance between not requiring a recklessly new compiler and still being able to get upstream bug fixes.

From this update most of our dependencies to their latest point releases. Then upgrade rusqlite, jsonrpc and rust-[bitcoin, miniscript] to their latest major version. Note their is a ton of updates. I've done a minimal due diligence to check what was updated, to what version, what changed, but i can't realistically audit all the code changes this is effectively pulling. I think we need to look into `cargo-crev`.

Take the opportunity to apply some clippy cleanups, and bump the version we're using in CI.

See commit messages for details.

### DIFF
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -547,8 +547,7 @@ pub fn create_spend(
             })?;
             fr
         }
-        .try_into()
-        .expect("u16 must fit in f32");
+        .into();
         let max_sat_wu = main_descriptor
             .max_sat_weight()
             .try_into()
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -23,7 +23,7 @@ pub const DUST_OUTPUT_SATS: u64 = 5_000;
 pub const LONG_TERM_FEERATE_VB: f32 = 10.0;
 
 /// Assume that paying more than 1BTC in fee is a bug.
-pub const MAX_FEE: u64 = bitcoin::blockdata::constants::COIN_VALUE;
+pub const MAX_FEE: bitcoin::Amount = bitcoin::Amount::ONE_BTC;
 
 /// Assume that paying more than 1000sat/vb in feerate is a bug.
 pub const MAX_FEERATE: u64 = 1_000;
@@ -53,7 +53,7 @@ impl fmt::Display for SpendCreationError {
             Self::InvalidOutputValue(amount) => write!(f, "Invalid output value '{}'.", amount),
             Self::InsaneFees(info) => write!(
                 f,
-                "We assume transactions with a fee larger than {} sats or a feerate larger than {} sats/vb are a mistake. \
+                "We assume transactions with a fee larger than {} or a feerate larger than {} sats/vb are a mistake. \
                 The created transaction {}.",
                 MAX_FEE,
                 MAX_FEERATE,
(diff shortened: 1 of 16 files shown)


src/spend.rs
```diff
@@ -295,13 +302,20 @@ fn select_coins_for_spend(
...
                 .try_into()
-                .expect("tx size must always fit in u32")
+                .expect("tx size must always fit in u32");
+            // Starting with version 0.31, rust-bitcoin now accounts for the segwit marker when
+            // serializing transactions with no input. See comment above.
+            if tx_with_change.input.is_empty() {
+                weight.saturating_sub(2)
+            } else {
+                weight
+            }
```
jp1ac4, line 318: I think we could alternatively subtract `base_tx.weight().to_wu()` instead of `base_weight` and then the extra 2 would be cancelled out from both weight values:
```diff
diff --git a/src/spend.rs b/src/spend.rs
index a373542..fc71280 100644
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -300,22 +300,15 @@ fn select_coins_for_spend(
     let long_term_feerate = FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_VB);
     let drain_weights = DrainWeights {
         output_weight: {
-            let mut tx_with_change = base_tx;
+            let mut tx_with_change = base_tx.clone();
             tx_with_change.output.push(change_txo);
-            let weight: u32 = tx_with_change
+            tx_with_change
                 .weight()
                 .to_wu()
-                .checked_sub(base_weight.into())
+                .checked_sub(base_tx.weight().to_wu())
                 .expect("base_weight can't be larger")
                 .try_into()
-                .expect("tx size must always fit in u32");
-            // Starting with version 0.31, rust-bitcoin now accounts for the segwit marker when
-            // serializing transactions with no input. See comment above.
-            if tx_with_change.input.is_empty() {
-                weight.saturating_sub(2)
-            } else {
-                weight
-            }
+                .expect("tx size must always fit in u32")
         },
         spend_weight: max_input_weight,
     };

```

darosior: Good call. Did something similar:
```diff
diff --git a/src/spend.rs b/src/spend.rs
index e8e2e10..091839f 100644
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -300,22 +300,18 @@ fn select_coins_for_spend(
     let long_term_feerate = FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_VB);
     let drain_weights = DrainWeights {
         output_weight: {
+            // We don't reuse the above base_weight.since 2 WU may have been substracted from it.
+            // See comment above for details.
+            let nochange_weight = base_tx.weight().to_wu();
             let mut tx_with_change = base_tx;
             tx_with_change.output.push(change_txo);
-            let weight: u32 = tx_with_change
+            tx_with_change
                 .weight()
                 .to_wu()
-                .checked_sub(base_weight.into())
+                .checked_sub(nochange_weight)
                 .expect("base_weight can't be larger")
                 .try_into()
-                .expect("tx size must always fit in u32");
-            // Starting with version 0.31, rust-bitcoin now accounts for the segwit marker when
-            // serializing transactions with no input. See comment above.
-            if tx_with_change.input.is_empty() {
-                weight.saturating_sub(2)
-            } else {
-                weight
-            }
+                .expect("tx size must always fit in u32")
         },
         spend_weight: max_input_weight,
     };
```

(Using a different var name to avoid shadowing if ever refactoring this code in the future.)

darosior:
reACK 0e99136404fcbbdee09eef4b116f17222cb2dee0 -- re-applying jp1ac4's ACK after addressing review

PR #908, opened 2024-01-08
gui(daemon): fix `rbfpsbt` params for external daemon

jp1ac4:
The `is_cancel` parameter gives an "invalid parameter" error when converted to string.

The `feerate_vb` parameter should only be passed if it has a value as otherwise it gives an "invalid parameter" error.


darosior:
Good catch.

edouardparis (approved):
utACK 5c06eef586ebd540f14f3ded4d3639dcbcfb0209

Issue #566, opened 2023-07-13
Support `bitcoind` being restarted from under us


darosior:
https://github.com/wizardsardine/liana/pull/562 cleaned up our bitcoind error handling code, and introduce a functional test for transient bitcoind failures. The test can trivially be modified to showcase that if bitcoind gets restarted while `lianad` is running, it would eventually panic.

This is because upon restart `bitcoind` would generate a new cookie file, and `lianad` would still be using the old credentials. Reading the new cookie to update our credentials would be trivial. However, this means introducing some mutable state in the `BitcoinD` struct.

PR #909, opened 2024-01-10
Bubble up errors from the Bitcoin interface instead of panic'ing there

darosior:
We assume a `bitcoind` is always running beneath Liana. As such, we make a [reasonable attempt to stay connected on spurious errors](https://github.com/wizardsardine/liana/blob/15b1f0c789d251dc3d420de50e859b6dd4bfda75/src/bitcoin/d/mod.rs#L342-L376) but we panic if we can't reach it [for more than a minute](https://github.com/wizardsardine/liana/blob/15b1f0c789d251dc3d420de50e859b6dd4bfda75/src/bitcoin/d/mod.rs#L40-L43). We also panic on unexpected API breaks (if an expected field isn't present or of the right type).
The rationale for this is pretty simple:
1. We simply can't continue without a running `bitcoind`. On a `bitcoind` failure there is nothing else to than stopping the program.
2. Failing directly in the `bitcoind` module (or any other implementation of the Bitcion interface) avoids cluttering the whole API with `Result`'s.

That said, we manifested some interest recently in actually bearing the cost of cumbersome `Result`'s in the API in order to:
1. Have nicer failures. Sure, we can't continue without `bitcoind`. But it doesn't mean we should straight away crash from under the user who would probably be using the GUI and have no clue what's going on.
2. On some non-fatal failures (which ones? cc @edouardparis), try to restart Liana. (Maybe for https://github.com/wizardsardine/liana/issues/566? If so there is nicer way to achieve this..)

I open this as a draft as it's based on #906 and i still have open questions:
- Is it worth it? Not going to pretend i don't have a huge sunk cost here since i've implemented it all already, but still worth asking ourselves this question.
- What do you think of the API?
  - The poller thread now needs to be externally managed, which i think makes the most sense.
  - I couldn't come up with anything better than `Result<T, Box<dyn std::error::Error>>` for the Bitcoin interface trait. I haven't looked too hard into it but if you've got a better suggestion i'd love to hear it.


edouardparis:
> Failing directly in the bitcoind module (or any other implementation of the Bitcion interface) avoids cluttering the whole API with Result's.
 
I personally believe Result is not cluttering the API. Rust has great ways to write error handling code in contrary of GO where you have if err != nil everywhere, and even with GO I was told to never use `panic` except in the internal of encapsulated code that have very clear invariants. Bitcoin node is totally external to the code and should not be considered a such.

> Have nicer failures. Sure, we can't continue without bitcoind. But it doesn't mean we should straight away crash from under the user who would probably be using the GUI and have no clue what's going on.
    On some non-fatal failures (which ones? cc @edouardparis), try to restart Liana. (Maybe for 
    https://github.com/wizardsardine/liana/issues/566? If so there is nicer way to achieve this..)
    
We agree about the user UX/UI, better to show him the issue. I see three kinds of errors:

* specific errors returned by commands (client request is not good, context changed, coin was already spent...): the error is already displayed to the user in a orange banner. 
* Failure error in the commands (means the poller is still running): we can display an error to the user to debug in an orange banner.
* Failure error in the poller (the poller stopped): we can redirect user to a full page like the one when the bitcoin node is not running with the same layout and a restart button.

> Is it worth it? Not going to pretend i don't have a huge sunk cost here since i've implemented it all already, but still worth asking ourselves this question.

Yes, especially in the long term. If we have multiple backend support we may have errors that we want to propagate instead of panicking.
    
> What do you think of the API? poller thread now needs to be externally managed, which i think makes the most sense.

I like that we have a final error handling in the `main.rs`.
    
>I couldn't come up with anything better than Result<T, Box<dyn std::error::Error>> for the Bitcoin interface trait. I haven't looked too hard into it but if you've got a better suggestion i'd love to hear it.

I believe it is fine for now, we can always rework on top if we need fined grained error matching.

darosior:
> specific errors returned by commands (client request is not good, context changed, coin was already spent...): the error is already displayed to the user in a orange banner.

If we are making a bad request i don't see how restarting could make a difference. We should stop as soon as possible. (Except maybe in rare mishandled edge cases where the delay might solve the issue.)

> Failure error in the poller (the poller stopped): we can redirect user to a full page like the one when the bitcoin node is not running with the same layout and a restart button.

I'm not sure. It seems to be very specific while the error could be due to many other things than the connection with `bitcoind`?

Maybe if the poller crash, a modal with the error should be displayed. We could let the user change their `bitcoind` settings there and try to restart but we shouldn't make it feel as if it's necessarily going to fix the issue.

edouardparis:
> If we are making a bad request i don't see how restarting could make a difference. We should stop as soon as possible. (Except maybe in rare mishandled edge cases where the delay might solve the issue.)

Then not a banner, but a full page error (not in ugly orange) with a warning icon and all the details.

> I'm not sure. It seems to be very specific while the error could be due to many other things than the connection with bitcoind?

> Maybe if the poller crash, a modal with the error should be displayed. We could let the user change their bitcoind settings there and try to restart but we shouldn't make it feel as if it's necessarily going to fix the issue.

In both case, it is good to have an error message to the user and maybe a button "retrieve the logs", that open the file system management window and let user register the logs of the day in a file he chose the location.
(Instead of pointing him toward the not trivial "<datadir>/<network>/liana.log")

And I agree that _we shouldn't make it feel as if it's necessarily going to fix the issue._

darosior:
Yes, that's all good points.

src/bitcoin/poller/looper.rs
```diff
@@ -208,13 +208,11 @@ fn new_tip(bit: &impl BitcoinInterface, current_tip: &BlockChainTip) -> TipUpdat
 }
 
 fn updates(
+    db_conn: &mut Box<dyn DatabaseConnection>,
```
darosior, line 210: Commit a8ca8500a4c60d2e135508d7efcbb1f6ef784f2e "poller: hold the same database lock across one update round" is confused. This is not a lock, it's a connection.

Issue #805, opened 2023-11-07
[GUI] Display wallet creation date at installer and in settings


kloaec:
Core know the wallet creation date, a user should:
- write it down, probably at the same time as they backup their descriptor
- be able to find it on the settings too.

darosior:
Yeah it would be nice to know the wallet creation date in the "wallet" section of the settings in the GUI.

We shouldn't get it from bitcoind though but from our own database: https://github.com/wizardsardine/liana/blob/d5b7a3a7de8fb36b5995bc25fc5171e425a22d8b/src/database/sqlite/schema.rs#L21-L31

An approach to this would be to:
1. Include this field in the `getinfo` RPC command
2. Display this field in the "wallet" section of the settings in the GUI by calling `getinfo`.

pythcoiner:
I can tackle that, but not sure I can deliver for v4

darosior:
Related to #620.

PR #920, opened 2024-01-14
Wallet creation time

pythcoiner:
address #805


darosior:
Concept ACK. I think the GUI integration should go in a latter PR (since it needs to update the daemon). So a small addition to `doc/API.md` seems to be all that's left to do here.

pythcoiner:
docs added, now ready for review

src/testutils.rs
```diff
@@ -5,11 +5,13 @@ use crate::{
     descriptors, DaemonHandle,
 };
 
+use std::convert::TryInto;
 use std::{
     collections::{HashMap, HashSet},
     env, fs, io, path, process,
     str::FromStr,
     sync, thread, time,
+    time::{SystemTime, UNIX_EPOCH},
```
darosior, line 14: Nit, for next time:
```suggestion
use std::{
    collections::{HashMap, HashSet},
    convert::TryInto, env, fs, io, path, process,
    str::FromStr,
    sync, thread, time,
    time::{SystemTime, UNIX_EPOCH},
```

tests/test_rpc.py
```diff
@@ -22,6 +22,7 @@
 
 def test_getinfo(lianad):
     res = lianad.rpc.getinfo()
+    assert 'timestamp' in res.keys()
```
darosior, line 25: nit: could've sanity checked the value. For instance that it's after now - 1 day and lower or equal to now.

darosior (commented):
ACK 79177945ada4803af409f963dd6185d436aa6e21

PR #921, opened 2024-01-14
Add optional txids param to listspendtxs

### LINKED ISSUE #862: Add an optional `txids` parameter to the `listspendtxs` command.
If set, only return transactions whose txid is in the list.

pythcoiner:
fix #862 
- [x] daemon
- [x] tests
- [x] doc

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -218,12 +218,14 @@ This command does not return anything for now.
 
 List stored Spend transactions.
 
+
+If `txids` is specified, only list transactions whose `txid` is in `txids`(empty list of `txids` is not allowed).
+
 #### Request
 
-This command does not take any parameter for now.
-
-| Field         | Type              | Description                                                 |
-| ------------- | ----------------- | ----------------------------------------------------------- |
+| Field         | Type                       | Description                          |
+| ------------- | -------------------------- | ------------------------------------ |
+| `txids`       | array of string (optional) | Ids of the transactions to retrieve  |
 
 #### Response
 
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -62,6 +62,7 @@ pub enum CommandError {
     /// Overflowing or unhardened derivation index.
     InvalidDerivationIndex,
     RbfError(RbfErrorInfo),
+    EmptyFilterList,
 }
 
 impl fmt::Display for CommandError {
@@ -114,6 +115,7 @@ impl fmt::Display for CommandError {
                 write!(f, "Unhardened or overflowing BIP32 derivation index.")
             }
             Self::RbfError(e) => write!(f, "RBF error: '{}'.", e),
+            Self::EmptyFilterList => write!(f, "Filter list is empty, should supply None instead."),
         }
     }
 }
@@ -589,14 +591,32 @@ impl DaemonControl {
         }
     }
 
-    pub fn list_spend(&self) -> ListSpendResult {
+    pub fn list_spend(
+        &self,
+        txids: Option<Vec<bitcoin::Txid>>,
+    ) -> Result<ListSpendResult, CommandError> {
+        if let Some(ids) = &txids {
+            if ids.is_empty() {
+                return Err(CommandError::EmptyFilterList);
+            }
+        }
+
         let mut db_conn = self.db.connection();
-        let spend_txs = db_conn
-            .list_spend()
+        let spend_psbts = db_conn.list_spend();
+
+        let txids_set: Option<HashSet<_>> = txids.as_ref().map(|list| list.iter().collect());
+        let spend_txs = spend_psbts
             .into_iter()
-            .map(|(psbt, updated_at)| ListSpendEntry { psbt, updated_at })
+            .filter_map(|(psbt, updated_at)| {
+                if let Some(set) = &txids_set {
+                    if !set.contains(&psbt.unsigned_tx.txid()) {
+                        return None;
+                    }
+                }
+                Some(ListSpendEntry { psbt, updated_at })
+            })
             .collect();
-        ListSpendResult { spend_txs }
+        Ok(ListSpendResult { spend_txs })
     }
 
     pub fn delete_spend(&self, txid: &bitcoin::Txid) {
--- a/src/jsonrpc/api.rs
+++ b/src/jsonrpc/api.rs
@@ -10,7 +10,7 @@ use std::{
     str::FromStr,
 };
 
-use miniscript::bitcoin::{self, psbt::Psbt};
+use miniscript::bitcoin::{self, psbt::Psbt, Txid};
 
 fn create_spend(control: &DaemonControl, params: Params) -> Result<serde_json::Value, Error> {
     let destinations = params
@@ -226,6 +226,32 @@ fn list_confirmed(control: &DaemonControl, params: Params) -> Result<serde_json:
     ))
 }
 
+fn list_spendtxs(
+    control: &DaemonControl,
+    params: Option<Params>,
+) -> Result<serde_json::Value, Error> {
+    let txids: Option<Vec<bitcoin::Txid>> = if let Some(p) = params {
+        let tx_ids = p.get(0, "txids");
+        if let Some(ids) = tx_ids {
+            let ids: Vec<Txid> = ids
+                .as_array()
+                .and_then(|arr| {
+                    arr.iter()
+                        .map(|entry| entry.as_str().and_then(|e| bitcoin::Txid::from_str(e).ok()))
+                        .collect()
+                })
+                .ok_or_else(|| Error::invalid_params("Invalid 'txids' parameter."))?;
+            Some(ids)
+        } else {
+            None
+        }
+    } else {
+        None
+    };
+
+    Ok(serde_json::json!(&control.list_spend(txids)?))
+}
+
 fn list_transactions(control: &DaemonControl, params: Params) -> Result<serde_json::Value, Error> {
     let txids: Vec<bitcoin::Txid> = params
         .get(0, "txids")
@@ -391,7 +417,7 @@ pub fn handle_request(control: &DaemonControl, req: Request) -> Result<Response,
             })?;
             list_confirmed(control, params)?
         }
-        "listspendtxs" => serde_json::json!(&control.list_spend()),
+        "listspendtxs" => list_spendtxs(control, req.params)?,
         "listtransactions" => {
             let params = req.params.ok_or_else(|| {
                 Error::invalid_params(
--- a/src/jsonrpc/mod.rs
+++ b/src/jsonrpc/mod.rs
@@ -165,6 +165,7 @@ impl From<commands::CommandError> for Error {
             | commands::CommandError::AlreadyRescanning
             | commands::CommandError::InvalidDerivationIndex
             | commands::CommandError::RbfError(..)
+            | commands::CommandError::EmptyFilterList
             | commands::CommandError::RecoveryNotAvailable => {
                 Error::new(ErrorCode::InvalidParams, e.to_string())
             }
--- a/tests/test_rpc.py
+++ b/tests/test_rpc.py
@@ -22,7 +22,7 @@ from test_framework.utils import (
 
 def test_getinfo(lianad):
     res = lianad.rpc.getinfo()
-    assert 'timestamp' in res.keys()
+    assert "timestamp" in res.keys()
     assert res["version"] == "4.0.0-dev"
     assert res["network"] == "regtest"
     wait_for(lambda: lianad.rpc.getinfo()["block_height"] == 101)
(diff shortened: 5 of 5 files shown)


src/commands/mod.rs
```diff
@@ -580,13 +581,25 @@ impl DaemonControl {
...
+                    if tx.txid() == txid {
+                        spend_txs.push(ListSpendEntry {
+                            psbt: psbt.clone(),
+                            updated_at,
+                        });
+                    }
+                }
+            } else {
+                spend_txs.push(ListSpendEntry { psbt, updated_at });
+            };
```
darosior, line 601: Meh at the `O(N * M)` complexity here. I'd rather create a `HashSet` from the `Vector` of txids to be able in the loop to check presence from the set in `O(1)`. (That effectively makes it `O(N) + O(M)` instead.)

Side nit: your can also use `filter_map`.

darosior:
Concept ACK. Ping me once you're done adding the new field to `doc/API.md`.

doc/API.md
```diff
@@ -210,12 +210,14 @@ This command does not return anything for now.
 
 List stored Spend transactions.
 
-#### Request
 
-This command does not take any parameter for now.
+If `txids` specified, only transactions with an id matching will be retrieved.
```
darosior, line 214:
```suggestion
If `txids` is specified, only list transactions whose `txid` is in `txids`.
```

doc/API.md
```diff
@@ -210,12 +210,14 @@ This command does not return anything for now.
...
 
-| Field         | Type              | Description                                                 |
-| ------------- | ----------------- | ----------------------------------------------------------- |
+#### Request
+
+| Field         | Type                       | Description                           |
+| ------------- | -------------------------- | ------------------------------------- |
+| `txids`       | array of string (optional) | Ids of the transactions  to retrieve  |
```
darosior, line 220: nit
```suggestion
| `txids`       | array of string (optional) | Ids of the transactions to retrieve  |
```

src/commands/mod.rs
```diff
@@ -580,13 +580,23 @@ impl DaemonControl {
...
-    pub fn list_spend(&self) -> ListSpendResult {
+    pub fn list_spend(&self, txids: Option<Vec<bitcoin::Txid>>) -> ListSpendResult {
         let mut db_conn = self.db.connection();
-        let spend_txs = db_conn
-            .list_spend()
+        let spend_psbts = db_conn.list_spend();
+
+        let txids_set = txids
+            .as_ref()
+            .map(|list| list.iter().cloned().collect::<HashSet<_>>())
```
darosior, line 589: nit (untested):
```suggestion
        let txids_set: HashSet<_> = txids
            .as_ref()
            .map(|list| list.iter().cloned().collect())
```

src/commands/mod.rs
```diff
@@ -580,13 +580,23 @@ impl DaemonControl {
...
+            .unwrap_or_default();
+
+        let spend_txs: Vec<ListSpendEntry> = spend_psbts
             .into_iter()
+            .filter(|(psbt, _)| {
+                txids_set.is_empty() || txids_set.contains(&psbt.unsigned_tx.txid())
```
darosior, line 595: I think this is incorrect. Or is it intended that passing an empty list would behave the same as not passing a list?

src/commands/mod.rs
```diff
@@ -580,13 +580,23 @@ impl DaemonControl {
...
-        let spend_txs = db_conn
-            .list_spend()
+        let spend_psbts = db_conn.list_spend();
+
+        let txids_set = txids
+            .as_ref()
+            .map(|list| list.iter().cloned().collect::<HashSet<_>>())
+            .unwrap_or_default();
+
+        let spend_txs: Vec<ListSpendEntry> = spend_psbts
```
darosior, line 592: nit:
```suggestion
        let spend_txs = spend_psbts
```

(deduced from the returned type)

tests/test_rpc.py
```diff
@@ -457,10 +457,35 @@ def test_list_spend(lianad, bitcoind):
...
+        lianad.rpc.listspendtxs(txids=['abc'])
+
+    with pytest.raises(RpcError, match="Invalid params: Invalid 'txids' parameter."):
+        lianad.rpc.listspendtxs(txids=['123'])
+
     # Listing all Spend transactions will list them both. It'll tell us which one has
     # change and which one doesn't.
     list_res = lianad.rpc.listspendtxs()["spend_txs"]
     assert len(list_res) == 2
+
```
darosior, line 488: nit: spurious line break in an unrelated part of the test?

darosior (commented):
Is it on purpose that passing an empty list of txids behaves the same as not passing a list at all?

tests/test_rpc.py
pythcoiner: fixed

doc/API.md
pythcoiner: fixed

src/commands/mod.rs
pythcoiner: right, its incorrect, should i error if an empty txid list is passed or return an empty Vec?

darosior: I'm not sure. For instance in `listcoins` i think the behaviour is the same as what you implemented here, right @jp1ac4?

jp1ac4: Yep, for `listcoins`, an empty list means no filter, but in the GUI's `list_spend_transactions`, I was not consistent and used `None` for no filter:
https://github.com/wizardsardine/liana/blob/0e295acc8240e55b039d8e0fd56595a6d2ae13a0/gui/src/daemon/mod.rs#L96-L109

I guess we can still choose either option here depending on which makes more sense and then change any other commands to be consistent. I tend to think `None` for no filter makes sense and then an empty list doesn't require any extra logic as we just return whatever it contains, which could be nothing.

darosior: What i find the most logical is (in JSON):
- `null`: no filter
- `[]`: a filter which rules out any spend tx

Now, the `[]` doesn't make any sense so i wouldn't mind if it meant "no filter". Just gotta be consistent across the board for when an empty list means "no filter" or "most extreme filter".

darosior: I'm starting to think the best course of action here is to error in case of `[]` filter. If we find a use for it later on we can always change it. Better to go from Error -> No Error than to change its meaning.

pythcoiner: i've followed this way and added a test for this error case

src/commands/mod.rs
```diff
@@ -580,13 +580,25 @@ impl DaemonControl {
...
 
-    pub fn list_spend(&self) -> ListSpendResult {
+    pub fn list_spend(&self, txids: Option<Vec<bitcoin::Txid>>) -> ListSpendResult {
         let mut db_conn = self.db.connection();
-        let spend_txs = db_conn
-            .list_spend()
+        let spend_psbts = db_conn.list_spend();
+
+        let txids_set: Option<HashSet<_>> =
+            txids.as_ref().map(|list| list.iter().cloned().collect());
```
darosior, line 588: This clone is unnecessary.
```suggestion
            txids.as_ref().map(|list| list.into_iter().collect());
```

src/commands/mod.rs
```diff
@@ -580,13 +580,25 @@ impl DaemonControl {
...
+
+        let spend_txs = spend_psbts
             .into_iter()
+            .filter(|(psbt, _)| {
+                if let Some(set) = &txids_set {
+                    set.contains(&psbt.unsigned_tx.txid())
+                } else {
+                    true
+                }
+            })
```
darosior, line 598: nit: a `filter` followed by a `map` can be done more succintly in a single `filter_map`:
```diff
diff --git a/src/commands/mod.rs b/src/commands/mod.rs
index b54ec80..083b3a4 100644
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -589,14 +589,14 @@ impl DaemonControl {
 
         let spend_txs = spend_psbts
             .into_iter()
-            .filter(|(psbt, _)| {
+            .filter_map(|(psbt, updated_at)| {
                 if let Some(set) = &txids_set {
-                    set.contains(&psbt.unsigned_tx.txid())
-                } else {
-                    true
+                    if !set.contains(&psbt.unsigned_tx.txid()) {
+                        return None;
+                    }
                 }
+                Some(ListSpendEntry { psbt, updated_at })
             })
-            .map(|(psbt, updated_at)| ListSpendEntry { psbt, updated_at })
             .collect();
 
         ListSpendResult { spend_txs }
```

src/commands/mod.rs
```diff
@@ -580,13 +580,25 @@ impl DaemonControl {
         }
     }
 
-    pub fn list_spend(&self) -> ListSpendResult {
+    pub fn list_spend(&self, txids: Option<Vec<bitcoin::Txid>>) -> ListSpendResult {
```
darosior, line 583: You are missing a check for an empty list of `txids` in this function?

src/jsonrpc/api.rs
```diff
@@ -226,6 +226,34 @@ fn list_confirmed(control: &DaemonControl, params: Params) -> Result<serde_json:
...
+                .as_array()
+                .and_then(|arr| {
+                    arr.iter()
+                        .map(|entry| entry.as_str().and_then(|e| bitcoin::Txid::from_str(e).ok()))
+                        .collect()
+                })
+                .ok_or_else(|| Error::invalid_params("Invalid 'txids' parameter."))?;
+            if ids.is_empty() {
+                return Err(Error::invalid_params("Invalid 'txids' parameter."));
+            }
```
darosior, line 246: This check should be present for both interfaces (through the crate and through the JSONRPC), not only the JSONRPC.

src/commands/mod.rs
pythcoiner: i've got a clippy warning on `into_iter()`:

```
warning: this `.into_iter()` call is equivalent to `.iter()` and will not consume the `Vec`
   --> src/commands/mod.rs:587:76
    |
587 |         let txids_set: Option<HashSet<_>> = txids.as_ref().map(|list| list.into_iter().collect());
    |                                                                            ^^^^^^^^^ help: call directly: `iter`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#into_iter_on_ref
    = note: `#[warn(clippy::into_iter_on_ref)]` on by default
```
i change to `iter()`

src/commands/mod.rs
pythcoiner: done

src/jsonrpc/api.rs
pythcoiner: i've moved the check to the command level, created `CommandError::EmptyFilterList` ,a quite generalistic name in case we want to raise it later for other filter list, but can rename it to `CommandError::EmptyTxIdsList` if you want something more specific.

src/commands/mod.rs
```diff
@@ -580,14 +582,32 @@ impl DaemonControl {
...
+        if let Some(ids) = &txids {
+            if ids.is_empty() {
+                return Err(CommandError::EmptyFilterList);
+            }
+        }
+
         let mut db_conn = self.db.connection();
-        let spend_txs = db_conn
-            .list_spend()
+        let spend_psbts = db_conn.list_spend();
```
jp1ac4, line 596: I'm not sure if this has been discussed, but would it be worth also adding a filter on the DB query in `list_spend`? Perhaps at least if the number of txids is not too large so that the query string is not too long.

jp1ac4 (commented):
I did some quick testing and it's looking good. I've added some minor comments.

Although it's beyond the scope of this PR, would adding `txid` to the response be useful?

src/commands/mod.rs
pythcoiner: you mean in order to apply the filtering at the DB level?

jp1ac4: That's right. We did a similar thing for listing coins by status and/or outpoint.

tests/test_rpc.py
```diff
@@ -1188,9 +1215,7 @@ def test_rbfpsbt_cancel(lianad, bitcoind):
     # But we can't set the feerate explicitly.
     with pytest.raises(
         RpcError,
-        match=re.escape(
-            "A feerate must not be provided if creating a cancel."
-        ),
+        match=re.escape("A feerate must not be provided if creating a cancel."),
```
pythcoiner, line 1218: these lines have been reformated by black

tests/test_rpc.py
```diff
@@ -457,6 +457,33 @@ def test_list_spend(lianad, bitcoind):
     lianad.rpc.updatespend(res["psbt"])
     lianad.rpc.updatespend(res_b["psbt"])
 
+    # Check 'txids' parameter
+    list_res = lianad.rpc.listspendtxs()["spend_txs"]
+
+    txid = bitcoind.rpc.decodepsbt(list_res[0]['psbt'])['tx']['txid']
```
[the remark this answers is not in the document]
pythcoiner: done

tests/test_rpc.py
```diff
@@ -457,6 +457,33 @@ def test_list_spend(lianad, bitcoind):
...
+        lianad.rpc.listspendtxs(txids=[txid, 123])
+
+    with pytest.raises(RpcError, match="Invalid params: Invalid 'txids' parameter."):
+        lianad.rpc.listspendtxs(txids=[0])
+
+    with pytest.raises(RpcError, match="Invalid params: Invalid 'txids' parameter."):
+        lianad.rpc.listspendtxs(txids=[123])
+
+    with pytest.raises(RpcError, match="Invalid params: Invalid 'txids' parameter."):
+        lianad.rpc.listspendtxs(txids=['abc'])
```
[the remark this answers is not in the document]
pythcoiner: done

pythcoiner:
> Although it's beyond the scope of this PR, would adding `txid` to the response be useful?

i can do that on a follow-up

src/commands/mod.rs
pythcoiner: not strong idea, but i guess it should be done at only one place to avoid duplicate 'filtering feature', i can move the filtering to DB level if required

tests/test_rpc.py
jp1ac4: You could commit those lines separately, but I think it's also fine as part of the same commit.

src/jsonrpc/mod.rs
```diff
@@ -174,6 +174,9 @@ impl From<commands::CommandError> for Error {
             commands::CommandError::TxBroadcast(_) => {
                 Error::new(ErrorCode::ServerError(BROADCAST_ERROR), e.to_string())
             }
+            commands::CommandError::EmptyFilterList => {
+                Error::invalid_params("Invalid 'txids' parameter.")
```
jp1ac4, line 178: As this is a general error type, you could make this message more generic or otherwise pass through the message from the `CommandError` itself.

pythcoiner: adressed

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/da1ebce5b6328f8f5636192e7434d1ba598d5516.

Issue #903, opened 2024-01-03
RBF: warn the user if they are about to shoot them in the foot.


darosior:
If a user spends the unconfirmed change output from their transaction A in a later transaction B we should prevent them from RBF'ing A (and thereby invalidating B).

jp1ac4:
Should we just warn the user when creating the replacement or prevent them from doing so?

In case the transaction to be replaced is not in the mempool, I'll apply the check to any directly conflicting transactions. I could check these directly conflicting transactions for any outputs that are coins in our database that are being spent (in case this conflicting transaction is paying to one of our receive addresses or more than one change address).

darosior:
I'm leaning toward asking confirmation. If they don't know what they are doing they wouldn't confirm. And an experienced user can still proceed.

I'm not sure i understood the latter part. How i was thinking about it was: if we get a request to replace a tx `A`, we check in database if it has a change output and if it's marked as `spending` in the DB (ie we have a tx `B` chained onto our own change). We would return an error with the txid in this case.

Thinking through it, making it a confirmation instead of an error would be more work (would probably have to include another parameter, like `force`). A warning could work but it's maybe too "soft"?

jp1ac4:
> I'm not sure i understood the latter part. How i was thinking about it was: if we get a request to replace a tx A, we check in database if it has a change output and if it's marked as spending in the DB (ie we have a tx B chained onto our own change). We would return an error with the txid in this case.

My thinking was that this check should apply to the directly conflicting transactions of the replacement transaction, which may or may not be `A`. If `A` is in the mempool, then this will be the only directly conflicting transaction, but if not, then there may be one or more other directly conflicting transactions. Also, it's possible that one of the outputs of these directly conflicting transactions is a non-change coin in our DB, so I was thinking that we could check for any outputs that are coins marked as spending, not just for change outputs.

> I'm leaning toward asking confirmation. If they don't know what they are doing they wouldn't confirm. And an experienced user can still proceed.

I agree we should allow the user to proceed. If there's a clear warning of some kind, then we can assume they know what they're doing.

> Thinking through it, making it a confirmation instead of an error would be more work (would probably have to include another parameter, like force). A warning could work but it's maybe too "soft"?

[...]

Issue #887, opened 2023-12-15
Refresh the state of the wallet after broadcasting a transaction


darosior:
This is in view of having a nicer UX when sending a transaction in the GUI. Our [broadcast command](https://github.com/wizardsardine/liana/blob/48feee16f5180a505c21b037c0d716791c67e97b/src/commands/mod.rs#L618-L642) should trigger the poller to do a round of updates. I'm not sure how best to update [the poller](https://github.com/wizardsardine/liana/blob/48feee16f5180a505c21b037c0d716791c67e97b/src/bitcoin/poller/looper.rs#L328-L336) to be listening for such requests, or to otherwise be able to trigger a single round of updates.

kloaec:
this is much needed, it is frustrating to click home knowing it's displaying a past state but the poller is just taking its time.

edouardparis:
If we have the table `transactions` cache, the broadcast command could do:
- set the coins as spending (set the spend_txid) in database
- store the transaction in the table
- create change coins

Then no need to call for a poller update,

darosior:
I'd rather not share the same responsibility over multiple pieces of code..

darosior:
Also of course there is more to updating the database: we have some invariants which would be nice to not break just for the sake of not reusing the poller (what for?). For instance consistency with the chain tip. Trying to make these invariants hold would boil down to simply rewriting the poller's logic.

Issue #677, opened 2023-09-01
Use `listcoins` filters in the GUI


darosior:
After #676, make use of the newly introduced filters to avoid large queries in the GUI when not necessary.

PR #930, opened 2024-01-23
gui: add warnings for descendant & conflicting transactions that will be dropped

jp1ac4:
This is for #903.

It adds two warnings to the GUI:
- when creating a new RBF if there are any descendant transactions of the transaction to be replaced:
image

- when broadcasting a transaction if there are any conflicting transactions
image

These warnings are generated in the GUI using coins data from the DB and so will not appear if the DB coins have not yet been updated. Would resolving https://github.com/wizardsardine/liana/issues/887 help ensure the coins are updated before running these checks?

### DIFF
--- a/gui/src/app/state/transactions.rs
+++ b/gui/src/app/state/transactions.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     convert::TryInto,
     sync::Arc,
     time::{SystemTime, UNIX_EPOCH},
@@ -112,6 +112,16 @@ impl State for TransactionsPanel {
                     }
                 }
             },
+            Message::RbfModal(tx, is_cancel, prev_feerate_vb, res) => match res {
+                Ok(descendant_txids) => {
+                    let modal =
+                        CreateRbfModal::new(tx, is_cancel, prev_feerate_vb, descendant_txids);
+                    self.create_rbf_modal = Some(modal);
+                }
+                Err(e) => {
+                    self.warning = e.into();
+                }
+            },
             Message::View(view::Message::Close) => {
                 self.selected_tx = None;
             }
@@ -129,8 +139,30 @@ impl State for TransactionsPanel {
                                 .to_sat()
                                 .checked_div(tx.tx.vsize().try_into().unwrap())
                                 .unwrap();
-                            let modal = CreateRbfModal::new(tx.clone(), is_cancel, prev_feerate_vb);
-                            self.create_rbf_modal = Some(modal);
+                            let tx = tx.clone();
+                            let txid = tx.tx.txid();
+                            return Command::perform(
+                                async move {
+                                    daemon
+                                        // TODO: filter for spending coins when this is possible:
+                                        // https://github.com/wizardsardine/liana/issues/677
+                                        .list_coins()
+                                        .map(|res| {
+                                            res.coins
+                                                .iter()
+                                                .filter_map(|c| {
+                                                    if c.outpoint.txid == txid {
+                                                        c.spend_info.map(|info| info.txid)
+                                                    } else {
+                                                        None
+                                                    }
+                                                })
+                                                .collect()
+                                        })
+                                        .map_err(|e| e.into())
+                                },
+                                move |res| Message::RbfModal(tx, is_cancel, prev_feerate_vb, res),
+                            );
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


gui/src/app/view/psbt.rs
```diff
@@ -151,15 +159,50 @@ pub fn broadcast_action<'a>(warning: Option<&Error>, saved: bool) -> Element<'a,
...
+                .push_maybe(if conflicting_txids.is_empty() {
+                    None
+                } else {
+                    Some(
+                        Row::new()
+                            .push(icon::warning_icon())
+                            .push(text(format!(
+                                "WARNING: The following transaction{} spending one or more \
+                                inputs from the transaction to be broadcast and will be \
+                                dropped, along with any other transactions that depend on {}:",
```
darosior, line 172: Let's have a higher level first sentence before the technical transaction dependency description? I'm thinking something along the lines of "WARNING: replacing this transaction would invalidate some later payments. [...]".

gui/src/app/view/transactions.rs
```diff
@@ -169,15 +174,52 @@ pub fn create_rbf_modal<'a>(
...
+                None
+            } else {
+                Some(
+                    Row::new()
+                        .push(icon::warning_icon())
+                        .push(text(format!(
+                            "WARNING: The following transaction{} spending one or \
+                            more outputs from the transaction to be replaced and will \
+                            be dropped when the replacement is broadcast, along with \
+                            any other transactions that depend on {}:",
```
darosior, line 201: Same here.

darosior (commented):
Approach ACK.

> These warnings are generated in the GUI using coins data from the DB and so will not appear if the DB coins have not yet been updated. Would resolving https://github.com/wizardsardine/liana/issues/887 help ensure the coins are updated before running these checks?

In most cases yes. It wouldn't cover it if it was just broadcast from another wallet (one such realistic scenario would be in the case of a multi-user wallet such as a multisig). But i think that's fine.

jp1ac4:
I've made some changes to the wording as suggested. I also added the option to copy the transaction ID with a copy icon.

Here are some screenshots for the broadcast warning:

- a single transaction:
image

- more than one transaction:
image

For cancelling a transaction:
- a single transaction:
image

- more than one transaction:
image

And for bumping the fee:
image

I also reworded the existing "cancel transaction" message to mention sending coins "back to your wallet" instead of "back to us".

darosior:
That looks very good. Thank you for all the screenshots.

gui/src/app/state/transactions.rs
```diff
@@ -126,8 +127,24 @@ impl State for TransactionsPanel {
                                 .to_sat()
                                 .checked_div(tx.tx.vsize().try_into().unwrap())
                                 .unwrap();
-                            let modal =
-                                CreateRbfModal::new(tx.tx.txid(), is_cancel, prev_feerate_vb);
+                            let descendant_txids: HashSet<_> = cache
```
edouardparis, line 130: I would like we do not use the cache coins for this usage. cache is generally used to display something like the previous balance, while the new balance is calculated in order to not display zero values.

I think it would be good instead, to have CreateRbfModal::new() -> (modal, command)
and give the command result to modal.update so it updates its conflicting txids.

What do you think ?

jp1ac4: That sounds good thanks, I'll make this change. After we implement https://github.com/wizardsardine/liana/issues/677, then we'll be able to retrieve just the spending coins in this call.

jp1ac4: I've now made the change to use `daemon.list_coins` and pass the result to another message.

In the end, I didn't pass the result to `modal.update` because when I tried this, the modal would open without any warnings and then update shortly afterwards, and I felt it was better that the warnings would be there as soon as the modal opened. So the modal is created using the result of `list_coins`.

I made a similar change for the set of conflicting transactions when broadcasting a transaction.

jp1ac4:
As well as the changes discussed above in https://github.com/wizardsardine/liana/pull/930#discussion_r1473035686, I've also removed the `prev_feerate_vb` parameter from `CreateRbfModal::new()` as this value can now be obtained from the `tx` parameter.

In order for the new `BroadcastModal` message handler to access `Error` instead of `&Error`, I changed the `match` statement to use `message` instead of `&message`.

edouardparis:
ACK fde28bf571e4ba791617e62f0c124138cd58e1fe

edouardparis (approved):
ACK fde28bf571e4ba791617e62f0c124138cd58e1fe

PR #933, opened 2024-01-23
Fix error message when importing desc for wrong network

### LINKED ISSUE #781: GUI: Installer: clearer error on descriptor import for wrong network
When importing a descriptor, making a mistake when choosing the network on the first screen results in a pretty scary "INCOMPATIBLE DESCRIPTOR" error on the second screen when we copy paste the descriptor.

It would be nicer to just let them know the network is wrong.

edouardparis:
close #781

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -364,6 +364,7 @@ pub fn import_descriptor<'a>(
     network: bitcoin::Network,
     network_valid: bool,
     imported_descriptor: &form::Value<String>,
+    wrong_network: bool,
     error: Option<&String>,
 ) -> Element<'a, Message> {
     let row_network = Row::new()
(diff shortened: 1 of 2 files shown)


gui/src/installer/view.rs
```diff
@@ -392,7 +393,11 @@ pub fn import_descriptor<'a>(
             form::Form::new_trimmed("Descriptor", imported_descriptor, |msg| {
                 Message::DefineDescriptor(message::DefineDescriptor::ImportDescriptor(msg))
             })
-            .warning("Incompatible descriptor.")
+            .warning(if wrong_network {
+                "The descriptor is for an other network, you can change the network in the previous steps."
+            } else {
+                "Incompatible descriptor."
```
darosior, line 399: Could we have a less coloured error message in the general case? Like "error reading descriptor". "Incompatible" can freak a user out as it can be read as "we successfully read your descriptor and it's not compatible with this software", and it's often not the case.

jp1ac4 (commented):
I reviewed https://github.com/wizardsardine/liana/pull/933/commits/04e9d9d0c12ebe9e5b173689912d03a557097e1d and left a comment regarding the message wording.

edouardparis (commented):
Self-ACK 24f67d2a85b12730c669c4c7e311a534f75cc0ad

PR #941, opened 2024-01-29
gui: display error for rescan past pruned height

### LINKED ISSUE #900: GUI: errors when triggering rescan aren't displayed, at all
For instance here is when trying to rescan past the height after fixing #693.

image

edouardparis:
close #900

After discussion with @kloaec it is better to not return error if date is prior to genesis block date, just take genesis block timestamp

### DIFF
--- a/gui/src/app/state/settings/bitcoind.rs
+++ b/gui/src/app/state/settings/bitcoind.rs
@@ -1,4 +1,4 @@
-use std::convert::From;
+use std::convert::{From, TryInto};
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
@@ -8,7 +8,10 @@ use chrono::prelude::*;
 use iced::Command;
 use tracing::info;
 
-use liana::config::{BitcoinConfig, BitcoindConfig, BitcoindRpcAuth, Config};
+use liana::{
+    config::{BitcoinConfig, BitcoindConfig, BitcoindRpcAuth, Config},
+    miniscript::bitcoin::Network,
+};
 
 use liana_ui::{component::form, widget::Element};
 
@@ -40,7 +43,7 @@ impl BitcoindSettingsState {
             node_settings: config.map(|config| {
                 BitcoindSettings::new(
                     config.bitcoin_config.clone(),
-                    config.bitcoind_config.clone().unwrap(),
+                    config.bitcoind_config.unwrap(),
                     daemon_is_external,
                     bitcoind_is_internal,
                 )
@@ -315,6 +318,7 @@ pub struct RescanSetting {
     month: form::Value<String>,
     day: form::Value<String>,
     invalid_date: bool,
+    future_date: bool,
     past_possible_height: bool,
 }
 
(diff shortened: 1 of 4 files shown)


gui/src/app/view/settings.rs
```diff
@@ -533,8 +534,13 @@ pub fn rescan<'a>(
...
                         )
                         .push_maybe(if invalid_date {
+                            Some(p1_regular("Provided date is invalid").style(color::RED))
+                        } else {
+                            None
+                        })
+                        .push_maybe(if past_possible_height {
                             Some(
-                                p1_regular("Provided date is either invalid or anterior to the genesis block's timestamp.")
+                                p1_regular("Provided date earlier than the node prune height")
```
jp1ac4, line 543: I also get this error when passing a future date. Perhaps the message could include both possibilities.

gui/src/app/state/settings/bitcoind.rs
```diff
@@ -370,23 +359,46 @@ impl Setting for RescanSetting {
...
-                ) {
-                    if date < NaiveDate::from_str("2009-01-03").unwrap() {
-                        self.invalid_date = true;
-                        return Command::none();
-                    } else {
-                        self.invalid_date = false;
-                        date
+                )
+                .and_then(|d| d.and_hms_opt(0, 0, 0))
+                .map(|d| d.timestamp() as u32)
```
jp1ac4, line 368: I noticed from the logs that if I put dates earlier than 1970, a future timestamp is passed to the rescan. This shouldn't matter as future timestamps return an error, but perhaps if the timestamp is negative, it should be set to 0 before converting to `u32`?

jp1ac4 (commented):
I've added a couple of comments based on initial testing. Will continue testing.

gui/src/app/state/settings/bitcoind.rs
edouardparis: ah good catch !

edouardparis:
Thanks @jp1ac4 I fixed the comments and rebased

jp1ac4 (commented):
I've added a couple more comments regarding the negative timestamp value. Otherwise, my tests have been going well.

gui/src/app/state/settings/bitcoind.rs
```diff
@@ -370,27 +362,61 @@ impl Setting for RescanSetting {
...
+                        }
+                        Network::Signet => {
+                            if date < SIGNET_GENESIS_BLOCK_TIMESTAMP {
+                                info!("Date {} prior to genesis block, using genesis block timestamp {}", date, SIGNET_GENESIS_BLOCK_TIMESTAMP);
+                                SIGNET_GENESIS_BLOCK_TIMESTAMP
+                            } else {
+                                date
+                            }
+                        }
+                        _ => date,
```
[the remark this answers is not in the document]
edouardparis: date cannot be negative because in this case it would be inferior to genesis block timestamp and the genesis block timestamp would be chosen instead.

gui/src/app/state/settings/bitcoind.rs
```diff
@@ -370,27 +362,61 @@ impl Setting for RescanSetting {
...
+                }
                 self.processing = true;
                 info!("Asking deamon to rescan with timestamp: {}", t);
                 return Command::perform(
-                    async move { daemon.start_rescan(t).map_err(|e| e.into()) },
+                    async move {
+                        daemon
+                            .start_rescan(
+                                // t cannot be inferior to 0
+                                t as u32,
```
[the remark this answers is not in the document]
edouardparis: Yes you are right it is better to panic instead

edouardparis (commented):
Self-ACK 69fd1891f3d4c0c1d9fa3ec807a7a09bd8468a5c
Thanks @jp1ac4 for the good reviews

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/69fd1891f3d4c0c1d9fa3ec807a7a09bd8468a5c.

PR #943, opened 2024-01-29
[GUI] display wallet birthday

### LINKED ISSUE #805: [GUI] Display wallet creation date at installer and in settings

pythcoiner:
fixes #805

### DIFF
--- a/gui/src/app/state/settings/mod.rs
+++ b/gui/src/app/state/settings/mod.rs
@@ -78,7 +78,12 @@ impl State for SettingsState {
             }
             Message::View(view::Message::Settings(view::SettingsMessage::EditWalletSettings)) => {
                 self.setting = Some(
-                    WalletSettingsState::new(self.data_dir.clone(), self.wallet.clone()).into(),
+                    WalletSettingsState::new(
+                        self.data_dir.clone(),
+                        self.wallet.clone(),
+                        self.wallet.timestamp,
+                    )
+                    .into(),
                 );
                 self.setting
                     .as_mut()
--- a/gui/src/app/view/settings.rs
+++ b/gui/src/app/view/settings.rs
@@ -589,6 +589,7 @@ pub fn wallet_settings<'a>(
     keys_aliases: &[(Fingerprint, form::Value<String>)],
     processing: bool,
     updated: bool,
+    creation_date: u32,
 ) -> Element<'a, Message> {
     dashboard(
         &Menu::Settings,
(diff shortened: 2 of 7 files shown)


pythcoiner:
should I display date as MM/DD/YY, as it can be confusing to some european country that use DD/MM/YY?
maybe better to same as the rescan date settings:
image

gui/src/app/state/settings/mod.rs
```diff
@@ -78,7 +78,12 @@ impl State for SettingsState {
             }
             Message::View(view::Message::Settings(view::SettingsMessage::EditWalletSettings)) => {
                 self.setting = Some(
-                    WalletSettingsState::new(self.data_dir.clone(), self.wallet.clone()).into(),
+                    WalletSettingsState::new(
+                        self.data_dir.clone(),
+                        self.wallet.clone(),
+                        daemon.get_info().unwrap().timestamp,
```
pythcoiner, line 84: not sure about unwrap here?
buble up error?
change creation_date type as Option<u32>?

gui/src/app/view/settings.rs
```diff
@@ -738,3 +740,15 @@ pub fn register_wallet_modal<'a>(
...
         .into()
 }
+
+pub fn creation_date_message(creation_date: u32) -> Option<Row<'static, Message>> {
+    if let Some(datetime) = chrono::NaiveDateTime::from_timestamp_opt(creation_date as i64, 0) {
+        return Some(
+            Row::new()
+                .push(text("Wallet creation date:").bold())
+                .spacing(10)
+                .push(text(datetime.format("%m/%d/%Y").to_string())),
```
darosior, line 750: I'd say `%y/%m/%d` is the least ambiguous

gui/src/app/state/settings/mod.rs
edouardparis: Daemon request should be in an asynchronous iced command. But I think for this time it is better to pass the wallet timestamp to the Wallet struct in src/app/wallet.rs
```
impl Wallet {
    pub fn new(main_descriptor: LianaDescriptor, timestamp: u32) -> Self {

```

pythcoiner:
comments have been adressed

darosior:
Just to give an update on this, it's being stuck because we had a disagreement (Kevin and i) about how this should be displayed to the user. I'm not sure where this landed though.

pythcoiner:
This PR is stale for 2+ years, I'm closing, feel free  to re-openif you re-gain some interest in it.

PR #944, opened 2024-01-29
gui: duplicate labels for rbf psbt

### LINKED ISSUE #861: GUI: RBF: automated labelling?
Should we set a label for RBF'd transactions such as:
- If the transaction has a label, prepend it with "RBF of <og label>" or "Cancel of <og label>"
- If it doesn't use the txid in place of the og label

Should it be set as a transaction label? I'd say yes. But in most cases it's payments which'd have labels.

edouardparis:
close #861


edouardparis:
Updated to add check is_cancel on top of jp1ac4 approval.
Self-ACK 6dd1d87228efc691151cc603dae6844a85d63099

PR #947, opened 2024-01-30
gui: use `createspend` command for coin selection and missing amount

jp1ac4:
This PR is a follow-up to https://github.com/wizardsardine/liana/pull/927 and uses changes made in https://github.com/wizardsardine/liana/pull/938.

It reverts to using the `createspend` command in the GUI for automated coin selection and to determine the amount left to select when creating a new spend (which was previously changed in https://github.com/wizardsardine/liana/pull/863).

With this PR, the changes from https://github.com/wizardsardine/liana/pull/873 will become effective in the GUI so that (some) unconfirmed coins are used as candidates and any additional fee to pay for ancestors is included when calculating the amount left to select.

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -381,7 +381,7 @@ impl Step for DefineSpend {
                         return Command::perform(
                             async move {
                                 daemon
-                                    .create_spend_tx(&inputs, &outputs, feerate_vb)
+                                    .create_spend_tx(&inputs, &outputs, feerate_vb, None)
                                     .map_err(|e| e.into())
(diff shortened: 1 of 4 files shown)


jp1ac4:
Thanks to @edouardparis for spotting that the amount left to select would not update if the user deselected all coins. This was because passing an empty list of outpoints to `create_spend_tx` would use auto-selection rather than manual selection with an empty list of inputs. I've made a change as suggested to set the amount left to select in that case to the total destination value.

edouardparis (approved):
ACK cb5073c83a54b3e4d430a7b30ce590c8385f7f7c

PR #951, opened 2024-02-02
gui(installer): clear checkbox if descriptor is changed

jp1ac4:
Currently, once the box has been ticked to confirm the descriptor has been backed up, it remains ticked even if the user goes back and changes the descriptor.

This PR will clear the checkbox if the descriptor is changed so that the user knows it needs to be backed up


edouardparis (approved):
ACK 416dc9d2d83c3b06ce7e7f5e282a5b9251fa303d

PR #953, opened 2024-02-02
[Daemon] Add -v/--version/-h/--help args 

pythcoiner:
image

### DIFF
--- a/src/bin/daemon.rs
+++ b/src/bin/daemon.rs
@@ -5,13 +5,19 @@ use std::{
     process, thread, time,
 };
 
-use liana::{config::Config, DaemonHandle};
+use liana::{config::Config, DaemonHandle, VERSION};
 
-fn print_help_exit() {
+fn print_help_exit(code: i32) {
+    eprintln!("lianad version {}", VERSION);
     eprintln!("A TOML configuration file is required to run lianad. By default lianad looks for a 'config.toml' file in its data directory. A different one may be provided like so: '--conf <config file path>'.");
     eprintln!("A documented sample is available at 'contrib/lianad_config_example.toml' in the source tree (https://github.com/wizardsardine/liana/blob/v1.0/contrib/lianad_config_example.toml).");
     eprintln!("The default data directory path is a 'liana/' folder in the XDG standard configuration directory for all OSes but Linux ones, where it's '~/.liana/'.");
-    process::exit(1);
+    process::exit(code);
+}
+
+fn print_version() {
+    eprintln!("{}", VERSION);
+    process::exit(0);
 }
 
 fn parse_args(args: Vec<String>) -> Option<PathBuf> {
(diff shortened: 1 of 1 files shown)


src/bin/daemon.rs
```diff
@@ -19,7 +19,10 @@ fn parse_args(args: Vec<String>) -> Option<PathBuf> {
         return None;
     }
 
-    if args[1] != "--conf" {
+    if args[1] == "--version" || args[1] == "-v" {
+        eprintln!("{}", VERSION);
+        process::exit(1);
+    } else if args[1] != "--conf" {
         eprintln!("Only a single command line argument is supported: --conf. All other configuration parameters must be specified in the configuration file.");
```
jp1ac4, line 26: Would it be worth updating this message? Maybe not as this message is more about running the daemon.

pythcoiner: not sure it worth it, @darosior?

darosior:
I'm reluctant to add more than a single command line argument. Maybe the version could be displayed in the help?

pythcoiner:
let you decide about that, i think -v/--version is a kind of standard 'non-argument' as it's not a way to launch the program and i feel almost (CLI) program use it but no strong opinion.

darosior:
Fine, let's do this. A `--help` / `-h` which simply calls `print_help` (but doesn't return `1`) would be helpful too in the same spirit i guess.

src/bin/daemon.rs
```diff
@@ -19,7 +19,10 @@ fn parse_args(args: Vec<String>) -> Option<PathBuf> {
         return None;
     }
 
-    if args[1] != "--conf" {
+    if args[1] == "--version" || args[1] == "-v" {
+        eprintln!("{}", VERSION);
+        process::exit(1);
```
darosior, line 24: This isn't an error? Let's return `0`.

darosior (changes requested):

pythcoiner:
> Fine, let's do this. A `--help` / `-h` which simply calls `print_help` (but doesn't return `1`) would be helpful too in the same spirit i guess.

instead creating a nec `print_help()` method, i just add an argument to pass the return code to `print_help_exit(code)`, wdyt @darosior ?

src/bin/daemon.rs
darosior: Agree with @jp1ac4, but let's keep that for future work.

darosior:
I'm a bit confused. You introduce a new `--help` argument as i suggested, but then dropped the `--version` argument. Do you think they are redundant? I'm fine with having only `--help`, just in this case please squash your commits into one (and update the commit / PR titles).

> instead creating a new print_help() method, i just add an argument to pass the return code to print_help_exit(code), wdyt @darosior ?

Sounds good.

pythcoiner:
> I'm a bit confused. You introduce a new `--help` argument as i suggested, but then dropped the `--version` argument. Do you think they are redundant? I'm fine with having only `--help`, just in this case please squash your commits into one (and update the commit / PR titles).
> 
> > instead creating a new print_help() method, i just add an argument to pass the return code to print_help_exit(code), wdyt @darosior ?
> 
> Sounds good.

some misunderstanding, on my side, done

src/bin/daemon.rs
```diff
@@ -5,27 +5,37 @@ use std::{
...
+    process::exit(0);
 }
 
 fn parse_args(args: Vec<String>) -> Option<PathBuf> {
     if args.len() == 1 {
         return None;
     }
 
-    if args[1] != "--conf" {
+    if args[1] == "--help" || args[1] == "-h" {
```
darosior, line 28: If anybody looks at this and thinks it does not do proper bounds checks, the args.len() check above actually does. This is because args is never empty because it always contains the program's name. Still i keep having to double check this everytime i read this code so it'd be worth making the check above `args.len() <= 1`.

darosior (approved):
ACK cd0e69ab237b7fb33aebebb4ca9bad12f06239fe

Issue #797, opened 2023-11-03
Don't increment  address index on `receive` menu click


pythcoiner:
Every time the user clicks on the `receive` menu button, a new address is generated and so the receive address index is incremented, we can instead generate a new address when the user clicks on the copy button, this avoids generating unused addresses.

In the case the user generates a batch of addresses, we could also change the address display color once the copy button is clicked.

I also wondering how we handle the receive/change index when Liana is used from several machines, do we update the index if we detect receive or change addresses w/ an index over the DB index have already coins?

darosior:
I don't think the hassle would be worth it. The derivation index is incremented before the address is even passed to the GUI. The thinking is that no address should get out without us having increased our derivation index first. Better waste a few derivation indexes than risk reusing an address.

Yeah the derivation index is also adjusted based on onchain data: https://github.com/wizardsardine/liana/blob/d5b7a3a7de8fb36b5995bc25fc5171e425a22d8b/src/bitcoin/poller/looper.rs#L58-L66

pythcoiner:
I just wonder if this can cause some coins to be forget during rescan, does core use a gap of adresses?

kloaec:
There is a gap, it's should be large enough (100+ if i remember).

I understand it "feels weird" because we aren't used to that, but it really helps with the address reuse problem. People should see addresses as disposable, free to generate, things.

darosior:
Bitcoind will use a "gap limit" of `1000`. On our end we use a look-ahead in the address->deriv index mapping of 200.

https://github.com/bitcoin/bitcoin/blob/2b3f43b96ef0b674bf350f50f317477b8d3e1e56/src/wallet/scriptpubkeyman.h#L54-L55

https://github.com/wizardsardine/liana/blob/d5b7a3a7de8fb36b5995bc25fc5171e425a22d8b/src/database/sqlite/utils.rs#L7

I'm going to close this, feel free to reopen if you think there is still something to be done here.

pythcoiner:
should we let advanced user access LOOK_AHEAD_LIMIT, in .toml file by example?

PR #959, opened 2024-02-06
Refac app: keep every panel states

edouardparis:
This refac keeps each panel state in parallel.

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -1,4 +1,10 @@
-use std::{cmp::Ordering, collections::HashMap, str::FromStr, sync::Arc};
+use std::{
+    cmp::Ordering,
+    collections::{HashMap, HashSet},
+    iter::FromIterator,
+    str::FromStr,
+    sync::Arc,
+};
 
 use iced::{Command, Subscription};
 use liana::{
@@ -143,6 +149,11 @@ impl DefineSpend {
     }
 
     pub fn with_coins_sorted(mut self, blockheight: u32) -> Self {
+        self.sort_coins(blockheight);
+        self
+    }
+
+    fn sort_coins(&mut self, blockheight: u32) {
         let timelock = self.timelock;
         self.coins.sort_by(|(a, a_selected), (b, b_selected)| {
             if *a_selected && !b_selected || !a_selected && *b_selected {
@@ -157,7 +168,6 @@ impl DefineSpend {
                 a.block_height.cmp(&b.block_height)
             }
         });
-        self
     }
 
     pub fn self_send(mut self) -> Self {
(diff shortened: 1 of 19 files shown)


jp1ac4:
From an initial test, I noticed that the RBF modal showing "Go to replacement PSBT" remains open when re-opening that transaction. This could be useful in a way so that the user can get back to the replacement PSBT, but I think it might be better if we cleared the modal when the user clicks on "Go to replacement PSBT".

For this, we would probably need to add a new message for this button to clear the modal before opening the replacement PSBT. We could also do it in a follow-up.

jp1ac4:
If I use RBF on an unconfirmed transaction, after clicking on "Go to replacement" in the RBF modal and then clicking PSBTs in the left-hand menu, only the replacement PSBT is shown initially and the other PSBTs appear shortly afterwards, even if the PSBTs had previously been loaded by a previous click on the PSBTs menu. Is it possible to keep the PSBTs if they've already been loaded once?

pythcoiner:
Currently testing this, coins table seems missing in `send` page (but the coins displayed in `coins` and `home` page)

image

image

pythcoiner:
Currently the state of addresses generated in `receive` is keep if we leave and come back, we talk about a related topic in #797

jp1ac4:
I found a couple of issues on the Transactions screen:
- After replacing a transaction, the replaced transaction remains as Pending on the Transactions screen, whereas previously the replaced transaction would no longer appear.
- I also find the Pending label remains after a transaction has been confirmed, regardless of any RBF.

gui/src/app/state/spend/step.rs
```diff
@@ -399,6 +409,17 @@ impl Step for DefineSpend {
                         self.batch_label.valid = label.len() <= 100;
                         self.batch_label.value = label;
                     }
+                    view::CreateSpendMessage::Clear => {
```
jp1ac4, line 412: I think we also need to clear the MAX checkbox and unset the amount left to select:

image

Note here the "Next" button is still clickable and will cause a panic. So I guess `is_valid` also needs to be set to `false`.

Similarly for `is_duplicate`:

image

Perhaps for these latter values, we could let the `redraft` and `check_valid` functions update them at the end instead of using `return Command::none()`, so only the MAX checkbox would need be cleared here.

jp1ac4 (commented):
It's looking good. I just found one more issue relating to the "Clear" button on the Send page.

edouardparis:
Thanks @jp1ac4, fixed and rebased

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/18e040e51ff835f5f6c8218392496865b6f8737a.

PR #963, opened 2024-02-07
commands: increment address index if current value used for new spend

jp1ac4:
This is a fix to ensure the change index in the database is incremented if a new spend is created with a change address derived from the current index, regardless of whether this new spend is broadcast or not.

EDIT: The same fix has been applied to the receive index.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -267,7 +267,7 @@ impl DaemonControl {
                     .increment()
                     .expect("Must not get into hardened territory");
                 db_conn.set_change_index(next_index, &self.secp);
-            } else if !is_change && db_conn.receive_index() < *index {
+            } else if !is_change && db_conn.receive_index() <= *index {
                 let next_index = index
                     .increment()
                     .expect("Must not get into hardened territory");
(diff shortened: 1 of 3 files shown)


src/commands/mod.rs
```diff
@@ -261,7 +261,7 @@ impl DaemonControl {
         addr_info: &Option<AddrInfo>,
     ) {
         if let Some(AddrInfo { index, is_change }) = addr_info {
-            if *is_change && db_conn.change_index() < *index {
+            if *is_change && db_conn.change_index() <= *index {
```
darosior, line 264: Should also fix the check against receive index right below?

jp1ac4: Ah yes, I'll take a look.

jp1ac4: I've now made this fix for the receive index and updated the functional tests.

I didn't test for this in the rust tests as `derivation_index_by_address` returns `None` in the dummy DB and so we don't enter the affected function in relation to receive addresses (we do enter the affected function for change addresses that have been generated within `create_spend`).

jp1ac4:
Rebased on master.

darosior:
ACK 323470da04d80db53ea750b91eff6eb5cff18765 -- great catch, thanks.

darosior:
Rebased on master.

darosior:
re-ACK 3a12ea0f9f7311fe5f7b795a7d43a5e2996146b6

PR #965, opened 2024-02-08
poller: unspend expired before new spend

jp1ac4:
This change ensures that the spend txid of a coin is updated directly to another value in case a conflicting spend is detected.

The previous order caused the spend txid to first be cleared, which would misleadingly make the coin appear as confirmed
rather than spending.

I've added a new utils function for the functional tests that is a slight generalisation of `wait_for` with an additional condition that must always be met while waiting.

`wait_for` now calls this new function with the condition being one that is always true.

### DIFF
--- a/tests/test_framework/utils.py
+++ b/tests/test_framework/utils.py
@@ -35,17 +35,33 @@ def wait_for(success, timeout=TIMEOUT, debug_fn=None):
     debug_fn is logged at each call to success, it can be useful for debugging
     when tests fail.
     """
+    wait_for_while_condition_holds(success, lambda: True, timeout, debug_fn)
+
+
+def wait_for_while_condition_holds(success, condition, timeout=TIMEOUT, debug_fn=None):
+    """
+    Run success() either until it returns True, or until the timeout is reached,
+    as long as condition() holds.
+    debug_fn is logged at each call to success, it can be useful for debugging
+    when tests fail.
+    """
     start_time = time.time()
     interval = 0.25
-    while not success() and time.time() < start_time + timeout:
+    while True:
+        if time.time() >= start_time + timeout:
+            raise ValueError("Error waiting for {}", success)
+        if not condition():
+            raise ValueError(
+                "Condition {} not met while waiting for {}", condition, success
+            )
+        if success():
+            return
         if debug_fn is not None:
             logging.info(debug_fn())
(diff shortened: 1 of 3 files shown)


jp1ac4:
Refactored the new utils function.

jp1ac4:
Rebased on master.

tests/test_framework/utils.py
```diff
@@ -46,6 +56,10 @@ def wait_for(success, timeout=TIMEOUT, debug_fn=None):
             interval = 5
     if time.time() > start_time + timeout:
         raise ValueError("Error waiting for {}", success)
+    if not condition():
```
darosior, line 59: I think it's a bit fragile: it's possible for a caller to provide a `condition` function such as it returns `True` above and `False` here during the second call. It would be more robust to change the loop above to be:
```python
while True:
    if success() or time >= start_time + timeout:
        break
    if condition():
        raise
```

darosior (commented):
The rationale for this ordering is to make sure a coin marked as spending in this round which then gets marked as expired does get recorded correctly.

This is because we check the spend status of both the coins marked spending in DB but also the newly spending coins:
https://github.com/wizardsardine/liana/blob/147cc0b983faf910e9a082c3aa0b12077f839d48/src/bitcoin/poller/looper.rs#L136-L146

It seems this change could only cause issues in the unlikely scenario a coin gets detected as spending and "expires" right before we check its spend status in `spent_coins` and therefore we miss it did expire. But in this case it would just be marked as such at the next polling round, that's fine.

darosior (commented):
Concept ACK. It does make a lot of sense, i think that's a bugfix and is harmless. I'd like to sleep on it though.

darosior:
In a followup it would be worthwile to introduce this new `wait_for_while_condition_holds` helper for the rescan tests.

tests/test_framework/utils.py
jp1ac4: That's true. Perhaps we could also do:
```python
while True:
    if time.time() >= start_time + timeout:
        raise
    if not condition():
        raise
    if success():
        return
```

jp1ac4: I've now updated this function.

darosior:
ACK cc1de1d6d6710f1426a957806661e7f3461a7cb5

PR #958, opened 2024-02-06
database: convert `CoinStatus` to `listcoins` argument string

jp1ac4:
This is a preliminary PR towards https://github.com/wizardsardine/liana/issues/677 and is the counterpart to the existing `CoinStatus::from_arg()` method.

This is needed so that the GUI can pass the required string arguments to the `listcoins` RPC command when using an external Liana daemon.

PR #967, opened 2024-02-09
gui: add `listcoins` filters

jp1ac4:
This is to resolve https://github.com/wizardsardine/liana/issues/677.

As well as adding the filters to the daemon interface, I've applied filters in separate commits to different sections of the GUI.

This PR builds on changes from https://github.com/wizardsardine/liana/pull/958 and https://github.com/wizardsardine/liana/pull/965. The latter is required when filtering for pending transactions so that a coin whose spending txid changes (e.g. due to RBF) remains as spending.

### DIFF
--- a/gui/src/app/state/psbt.rs
+++ b/gui/src/app/state/psbt.rs
@@ -7,6 +7,7 @@ use iced::Subscription;
 
 use iced::Command;
 use liana::{
+    commands::CoinStatus,
     descriptors::LianaPolicy,
     miniscript::bitcoin::{bip32::Fingerprint, psbt::Psbt, Network, Txid},
 };
@@ -167,23 +168,15 @@ impl PsbtState {
                 return cmd;
             }
             Message::View(view::Message::Spend(view::SpendTxMessage::Broadcast)) => {
-                let outpoints: HashSet<_> = self.tx.coins.keys().cloned().collect();
+                let outpoints: Vec<_> = self.tx.coins.keys().cloned().collect();
                 return Command::perform(
                     async move {
                         daemon
-                            // TODO: filter for the outpoints in `tx.coins` when this is possible:
-                            // https://github.com/wizardsardine/liana/issues/677
-                            .list_coins(&[], &[])
+                            .list_coins(&[CoinStatus::Spending], &outpoints)
                             .map(|res| {
                                 res.coins
                                     .iter()
-                                    .filter_map(|c| {
-                                        if outpoints.contains(&c.outpoint) {
-                                            c.spend_info.map(|info| info.txid)
-                                        } else {
-                                            None
-                                        }
-                                    })
+                                    .filter_map(|c| c.spend_info.map(|info| info.txid))
                                     .collect()
                             })
                             .map_err(|e| e.into())
--- a/gui/src/app/state/psbt.rs
+++ b/gui/src/app/state/psbt.rs
@@ -173,7 +173,7 @@ impl PsbtState {
                         daemon
                             // TODO: filter for the outpoints in `tx.coins` when this is possible:
                             // https://github.com/wizardsardine/liana/issues/677
-                            .list_coins()
+                            .list_coins(&[], &[])
                             .map(|res| {
                                 res.coins
                                     .iter()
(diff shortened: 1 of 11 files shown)


edouardparis:
Beware that the application use the Coins message passing by to update its cache:
https://github.com/wizardsardine/liana/blob/master/gui/src/app/mod.rs#L155
If the coins message is filtered then the cache is wrong, so maybe before doing this kind of change, we should first merge an equivalent of https://github.com/wizardsardine/liana/pull/959 where every panel state is kept with its own coins list for example and the cache becomes unnecessary.

One of the main advantages of the new list_coins filter is the outpoints filter, we should use it in the daemon methods that retrieve the coins which outpoints are corresponding to the inputs and the outputs of the resulting HistoryTransaction, Psbt etc

gui/src/app/state/transactions.rs
```diff
@@ -146,23 +147,23 @@ impl State for TransactionsPanel {
...
                                         res.coins
                                             .iter()
-                                            .filter_map(|c| {
-                                                if c.outpoint.txid == txid {
-                                                    c.spend_info.map(|info| info.txid)
-                                                } else {
-                                                    None
-                                                }
-                                            })
+                                            .map(|c| c.spend_info.expect("coin is spending").txid)
```
edouardparis, line 166: I would prefer we keep a filter_map and no expect

gui/src/app/state/psbt.rs
```diff
@@ -167,23 +168,15 @@ impl PsbtState {
...
                                 res.coins
                                     .iter()
-                                    .filter_map(|c| {
-                                        if outpoints.contains(&c.outpoint) {
-                                            c.spend_info.map(|info| info.txid)
-                                        } else {
-                                            None
-                                        }
-                                    })
+                                    .map(|c| c.spend_info.expect("coin is spending").txid)
```
edouardparis, line 179: I would prefer we keep a filter_map and no expect

gui/src/app/state/psbt.rs
```diff
@@ -167,23 +168,15 @@ impl PsbtState {
...
             Message::View(view::Message::Spend(view::SpendTxMessage::Broadcast)) => {
-                let outpoints: HashSet<_> = self.tx.coins.keys().cloned().collect();
+                let outpoints: Vec<_> = self.tx.coins.keys().cloned().collect();
                 return Command::perform(
                     async move {
                         daemon
-                            // TODO: filter for the outpoints in `tx.coins` when this is possible:
-                            // https://github.com/wizardsardine/liana/issues/677
-                            .list_coins()
+                            .list_coins(&[CoinStatus::Spending], &outpoints)
```
edouardparis, line 175: we use the outpoint, it is enough, you can remove the status filter.

jp1ac4: I think it's OK to keep the status filter here as the previous `filter_map` was already filtering for coins with a non-null spend txid.

edouardparis: Ah you may right, I think we are good then.

gui/src/app/state/transactions.rs
```diff
@@ -146,23 +147,23 @@ impl State for TransactionsPanel {
...
+                                )
+                            })
+                            .collect();
                         return Command::perform(
                             async move {
                                 daemon
-                                    // TODO: filter for spending coins when this is possible:
-                                    // https://github.com/wizardsardine/liana/issues/677
-                                    .list_coins()
+                                    .list_coins(&[CoinStatus::Spending], &outpoints)
```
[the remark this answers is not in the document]
jp1ac4: As with the other comment, I've kept the status filter.

gui/src/app/state/psbt.rs
jp1ac4: I've changed this to filter_map.

gui/src/app/state/transactions.rs
jp1ac4: I've changed this to filter_map.

edouardparis (approved):
ACK 1c99376860458f350c3289cd995716aca96c4d87

PR #971, opened 2024-02-22
gui: prevent spend with duplicate addresses

jp1ac4:
Currently, if duplicate addresses are entered when creating a new spend, the user can click Next and only the second amount will be used for the draft PSBT.

This change prevents the user clicking Next if there are duplicate addresses and stops the `redraft` function running.


edouardparis (approved):
I did not manage to go to next step with two same addresses
ACK 8639f9b94f1fdceaee94aa927e8f9b7b07cc51f9

PR #972, opened 2024-02-22
ci: bump bitcoind version in functional tests to 26.0

darosior:
I thought there was some hacks i do in the functional tests i needed to update but apparently i was mistaken.


darosior:
Ran the whole test suite a dozen times locally without encountering any flaky test.

.cirrus.yml
```diff
@@ -54,10 +54,10 @@ task:
       export IS_BITCOIND_25=0
```
jp1ac4, line 54: Should we rename this variable to something like `IS_BITCOIND_LATEST`?

darosior: It's not the latest we need but anything above 24. Renamed.

darosior:
self-ACK bdf8b8f625030697ca58680fd0c99c3dd3bf77af - ran ci one last time and didn't notice any flakiness

Issue #546, opened 2023-05-29
When sending, have a `max` button to send the whole balance (sweep wallet)


pythcoiner:
In the send screen, it would be useful to have a `MAX` button that autofill the sum amount of all the wallet coins.
![image](https://github.com/wizardsardine/liana/assets/124568858/afa45ffa-7352-4a39-bf03-5298248ac8c6)

darosior:
Yes. It's something we've discussed. It's a bit intricate to make work as you change the feerate though.

pythcoiner:
> Yes. It's something we've discussed. It's a bit intricate to make work as you change the feerate though.

yes, i think if `max` has been selected (and until send value not changed by the user) the output amount might be implicitly (total - fee)

darosior:
This was brought up today again. A path forward could be to have the recovery command logic be moved to a "sweep" helper and add another command, `sweep`, that takes a single address and sweeps the whole balance there through the primary path.

The integration in the GUI would be more involved. If sweep was used, it should not let the user add more destinations. Nor select coins actually. @edouardparis wdyt?

darosior:
Adding the daemon label as ideally i think the sweep logic should be in the daemon and tested using the functional tests.

edouardparis:
Recovery panel could be renamed sweep and the primary path added to the list of recovery paths.
Wording can be:

2 paths are accessible next block, please select one
- The primary path
  - Signatures from Ed and Antoine can move 11 coins totalling 1.2334 BTC 
-  Recovery path:
   - Signatures from Ed can move 4 coins totalling 0.6244 BTC

darosior:
That seems reasonable and pretty straightforward.
------- Original Message -------

pythcoiner:
> `sweep`, that takes a single address and sweeps the whole balance there through the primary path.

Do you mean to move in batch _one to one_ or consolidate _many to one_?

darosior:
See also https://github.com/wizardsardine/liana/pull/821.

darosior:
Relevant: https://github.com/wizardsardine/liana/pull/863#discussion_r1423661296

darosior:
I'm not sure. Any opinion?

jp1ac4:
I think we could do both. Having the max button on the Send page is intuitive I think and users may expect to see that option there.

We could have the max button even for multiple recipients: for a given recipient, clicking the max button would return the max available for that recipient after taking into account the other recipients.

darosior:
[...]

PR #974, opened 2024-02-23
gui: add max button for spend recipients

jp1ac4:
This adds the MAX button from #546.

The max amount is calculated within the `redraft` function based on the change output. The recipient's amount is updated directly within this same function in order to avoid another call to `redraft` after updating the amount.

The MAX button only takes effect once when the user clicks it. If further changes are made (e.g. to feerate or other recipients), the value will not update and the user will need to click MAX again.

Currently, the MAX button is always clickable. It would be possible to determine for each recipient whether the MAX button should be clickable for that recipient (feerate is valid and all other recipients are valid and no duplicates), but it's not a trivial change so I'm not sure if it's worth doing as I think the behaviour is quite intuitive.

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -192,6 +192,13 @@ impl DefineSpend {
     /// if the user did not select a coin manually
     fn redraft(&mut self, daemon: Arc<dyn Daemon + Sync + Send>) {
         if !self.form_values_are_valid() || self.exists_duplicate() || self.recipients.is_empty() {
+            // The current form details are not valid to draft a spend, so remove any previously
+            // calculated amount as it will no longer be valid and could be misleading, e.g. if
+            // the user removes the amount from one of the recipients.
+            // We can leave any coins selected as they will either be automatically updated
+            // as soon as the form is valid or the user has selected these specific coins and
+            // so we should not touch them.
+            self.amount_left_to_select = None;
             return;
         }
 
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -1,6 +1,4 @@
-use std::collections::HashMap;
-use std::str::FromStr;
-use std::sync::Arc;
+use std::{cmp::Ordering, collections::HashMap, str::FromStr, sync::Arc};
 
 use iced::{Command, Subscription};
 use liana::{
@@ -65,6 +63,9 @@ pub trait Step {
 pub struct DefineSpend {
     balance_available: Amount,
     recipients: Vec<Recipient>,
+    /// If set, this is the index of a recipient that should
+    /// receive the max amount.
+    send_max_to_recipient: Option<usize>,
     /// Will be `true` if coins for spend were manually selected by user.
     /// Otherwise, will be `false` (including for self-send).
     is_user_coin_selection: bool,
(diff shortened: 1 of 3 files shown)


pythcoiner:
Love it!

What about greying out `Max` button if no fee rate/no coin selected?

pythcoiner:
i don't understand how it can be related to this PR, but i've one issue that i don't have on master:

image

i'm creating a new signet wallet, when i import key from coldcard, i got an `Xpub` instead a `Tpub`

edouardparis:
Working great, could it be possible to put the button it in a tooltip which on hover would display:
`Quickly calculate the remaining amount` or something like this.

jp1ac4:
In the end, it was simpler than I thought to add an exact condition on whether the max button for a given recipient should be clickable (it might be clickable for one recipient and not another).

I've also added a tooltip as suggested.

pythcoiner:
ACK  4e252d4

edouardparis:
>In the end, it was simpler than I thought to add an exact condition on whether the max button for a given recipient should be clickable (it might be clickable for one recipient and not another).

I cannot deactivate it to use it to another recipient.
If we lock the button, then user should be able to unlock it. 
Either:
1. we let it be a clickable button that redo calculation on click,
2. or we make it a "toggle" button: 
  - border became green to say it is active, user can deactivate it, only one recipient hast it.
  - and redraft takes it into account to redo calculation and modify the recipient amount at each modification of feerate.
  - The recipient amount form is greyed out as long as the Max button is active.

jp1ac4:
I've changed the button to a checkbox that remains selected and continues to recalculate the max for that recipient as changes are made elsewhere.

The checkbox can be selected at any time, but the max amount will only be calculated as long as the selected recipient's address is valid, all other recipients are valid and the feerate is valid. If the user has selected coins manually, then at least one coin must be selected and otherwise all coins will be selected for the calculation.

jp1ac4:
Thanks, @edouardparis, for your testing feedback.

I've made some further changes (some in separate commits as they are not directly related to the max button):
- Add alternative message to "Feerate needs to be set" in case something else is missing.
- If there is not enough information to redraft (incomplete form values), set the recipient's max amount to empty string and unset the amount left to select.
- if form values are valid and no coins are selected and it's user selection, set max amount to 0.
- if the amount left to select is 0, we only display "0 left to select" if the "Next" button is clickable, and otherwise display a message indicating what's still missing.

edouardparis:
It is working great ! Good work !

edouardparis:
Sorry a little nit, but I am ok to merge the PR.
It feels like the tooltip `Quickly calculate the remaining amount`does not provide a good explanation.
What do you think of something like: `Total amount of the selected coins less the other recipients amount and fees` ?

jp1ac4:
I think changing the wording makes sense given that now the checkbox remains selected.

Perhaps `Total amount remaining from selected coins after paying fee and any other recipients` ?

It might be better to omit "from selected coins" as it's not completely accurate in the case of auto-selection, so `Total amount remaining after paying fee and any other recipients`.

I've used "fee" rather than "fees" but I think both could work.

edouardparis:
> Total amount remaining after paying fee and any other recipients

Ok for this.

jp1ac4:
I've updated the tooltip wording and also changed its position to `Left` as otherwise it was hard to see the final letters.

Updated version:
image

edouardparis (approved):
ACK 46cd0f4923c5fe59a5ff08134c850d9a9ceb345a

PR #977, opened 2024-02-24
[GUI] Limit amount to 8 decimal digits

### LINKED ISSUE #798: Limit the send amout value field to 8 decimals
Can the input field format be enforced in Iced to avoid users entering precision under 1 sat??
image

pythcoiner:
fixes #798

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -521,7 +521,9 @@ impl Recipient {
                 }
             }
             view::CreateSpendMessage::RecipientEdited(_, "amount", amount) => {
-                self.amount.value = amount;
+                if let Some(a) = amount_checked(amount) {
+                    self.amount.value = a;
+                }
                 if !self.amount.value.is_empty() {
                     self.amount.valid = self.amount().is_ok();
                 } else {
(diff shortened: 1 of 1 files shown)


jp1ac4:
I'm wondering if this should rather be done at the form level, similarly to how we trim whitespaces:
https://github.com/wizardsardine/liana/blob/903328ffef8317b1f152633ac671ecd92071d8eb/gui/ui/src/component/form.rs#L53

E.g. we could add `new_amount_btc()` that trims any excess decimals before passing the value to the `on_change` function.

pythcoiner:
> I'm wondering if this should rather be done at the form level, similarly to how we trim whitespaces:
> 
> https://github.com/wizardsardine/liana/blob/903328ffef8317b1f152633ac671ecd92071d8eb/gui/ui/src/component/form.rs#L53
> 
> E.g. we could add `new_amount_btc()` that trims any excess decimals before passing the value to the `on_change` function.

i'll look into this, also i think we can filter out char other then "0123456789." and maybe automaticaly replace , by . ?

pythcoiner:
deprecated in favor of #992, following @jp1ac4 advice

PR #979, opened 2024-02-24
[GUI] fixed size/position for PSBT badges

### LINKED ISSUE #509: Spent label missaligned in PSBTs
While in the PSBTs section, the Spent label does not alight at the same point for "send-to-selfs" as it does to "normal" spends.

Seems the reason is because the reference point is the text shown on top of the miner fee.

Screenshot 2023-05-13 at 04 27 21

pythcoiner:
this PR try to fix #509
i refactored some badges , moving (optional) `batch` badge before mandatory ones and give fixed positions/size for mandatory badges

before:

image

after:

image

### DIFF
--- a/gui/src/app/view/psbts.rs
+++ b/gui/src/app/view/psbts.rs
@@ -139,9 +139,9 @@ fn spend_tx_list_view(i: usize, tx: &SpendTx) -> Element<'_, Message> {
                     None
                 })
                 .push_maybe(match tx.status {
-                    SpendStatus::Deprecated => Some(badge::deprecated_sized(120.0)),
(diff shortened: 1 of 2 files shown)


jp1ac4:
I like how this looks now. Should we use this new badge size also when viewing an individual PSBT?

image

pythcoiner:
i'm not sure anyone notice this change on psbt, as the badge in alone in the top right corner, but we can

darosior:
The before/after screenshots look amazing.

edouardparis (approved):
ACK e05e42136801e3168b99be378ae85e2852a8ccf6

I did a little refac to remove the need for `_sized` method. Thanks for the idea and the change.

PR #985, opened 2024-02-27
Taproot support

darosior:
This introduces Taproot support in the Liana daemon / core library.

We start by introducing support for `tr()` descriptors alongside `wsh()` descriptors in the Taproot modules. For Taproot-Liana descriptors whose primary spending path isn't a single key, we deterministically derive an unspendable internal key as per https://delvingbitcoin.org/t/unspendable-keys-in-descriptors/304/21. This is to allow signing devices to not display the internal key as a spendable key when verifying a descriptor. Currently signing device vendors signaled willingness to implement this scheme.

We then adapt the PSBT management logic to use Taproot fields when necessary and upgrade the hot signer to provide Schnorr signatures depending on the PSBT information.

Finally, the functional tests are adapted to be able to run the whole test suite under either P2WSH or Taproot. This is done in a somewhat hacky way as i bailed out of re-implementing a Taproot PSBT signer and finalizer in Python after implementing [TapMiniscript support in upstream python-bip380](https://github.com/darosior/python-bip380/pull/23). Instead we use a small Rust program to sign PSBTs for now and we skip a test which explicitly requires an external finalizer.


darosior:
I still need to:
- Cleanup the bugfixes to https://github.com/darosior/python-bip380/pull/23 i've got locally and release a new version to update it in our functional tests;
- Merge some preparatory PRs.

I should open a few issues for needed followups before i forget about it.

darosior:
Alright, i updated `python-bip380` and the CI to run the Taproot tests. Looks like one of them is a bit flaky, i'll have a look later. In the meantime it's ready for a first round of review!

darosior:
Rebased on master. I adapted the tests for the selection of unconfirmed coins from #873 in a separate commit, as it was less mechanic / obvious than the other tests.

What i still need to investigate:
- `test_coin_selection` is very slightly flaky. The assertion of the vsize of the ancestor might sometimes fail. This is unrelated to this PR, but can potentially make CI fail here;
- it seems the performance of `createspend` has been greatly reduced. We might see some timeout on functional tests making heavy use of this call. I need to profile it i suppose. But my hunch is rust-miniscript's way of populating a PSBT input is inefficient.

jp1ac4 (commented):
I went through most of the changes and found a couple of small typos in comments. Otherwise, the code made sense according to my limited understanding of Taproot :)

src/descriptors/mod.rs
```diff
@@ -49,6 +56,31 @@ impl From<LianaPolicyError> for LianaDescError {
...
+        // the policy. This is fine, because the keys in the policy are normalized (so
+        // the derivation path up to the wildcard is part of the origin).
+        if let Some((_, der_path_no_wildcard)) = der_path[..].split_last() {
+            if der_paths.contains(&der_path_no_wildcard.into()) {
+                return true;
+            }
+        }
+    }
+    false
+}
```
darosior, line 82: This could be part of a preparatory commit to reduce the diff of this one.

src/descriptors/mod.rs
```diff
@@ -540,26 +598,53 @@ impl DerivedSinglePathLianaDesc {
             .collect()
     }
 
+    // FIXME: update_with_descriptor() needs a Descriptor<DefiniteKey>. This is a temporary hack to
+    // avoid having to duplicate the cumbersome logic here. Could use translate_pk() instead in the
+    // future.
+    fn definite_desc(&self) -> descriptor::Descriptor<descriptor::DefiniteDescriptorKey> {
+        descriptor::Descriptor::<_>::from_str(&self.0.to_string()).expect("Must roundtrip")
```
darosior, line 610: :vomiting_face:

darosior (commented):
Just did a mechanical review. In addition to checking the correctness of what's here we need to think about what's *missing*.

src/signer.rs
```diff
@@ -419,14 +419,22 @@ mod tests {
             wildcard: Wildcard::Unhardened,
         });
         let recov_keys = descriptors::PathInfo::Single(recov_key);
-        let policy =
-            descriptors::LianaPolicy::new(prim_keys, [(46, recov_keys)].iter().cloned().collect())
-                .unwrap();
+        let policy = descriptors::LianaPolicy::new_legacy(
```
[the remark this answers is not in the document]
darosior: Fixed.

src/descriptors/mod.rs
```diff
@@ -539,6 +539,18 @@ impl DerivedSinglePathLianaDesc {
...
+
+    /// Update the PSBT input information with data from this derived descriptor.
+    pub fn update_psbt_in(&self, psbtin: &mut PsbtIn) {
+        psbtin.bip32_derivation = self.bip32_derivations();
+        psbtin.witness_script = Some(self.witness_script());
+    }
+
+    /// Update the info of a PSBT output for a change output with data from this derived
+    /// descriptor.
+    pub fn update_change_psbt_out(&self, psbtout: &mut PsbtOut) {
```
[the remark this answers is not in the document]
darosior: I think it's fine as is.

src/descriptors/mod.rs
darosior: Done.

tests/test_spend.py
```diff
@@ -308,42 +316,52 @@ def test_coin_selection(lianad, bitcoind):
...
     )
-    assert bitcoind.rpc.getmempoolentry(spend_txid_1)["ancestorsize"] == 161
-    assert bitcoind.rpc.getmempoolentry(spend_txid_1)["fees"]["ancestor"] * COIN == 339
-    # ancestor vsize at feerate 10 sat/vb = ancestor_fee / 10 = 339 / 10 = 33
-    # extra_weight <= (extra vsize * witness factor) = (161 - 33) * 4 = 512
-    # additional fee at 10 sat/vb (2.5 sat/wu) = 512 * 2.5 = 1280
+    anc_vsize = int(bitcoind.rpc.getmempoolentry(spend_txid_1)["ancestorsize"])
+    anc_fees = int(bitcoind.rpc.getmempoolentry(spend_txid_1)["fees"]["ancestor"] * COIN)
+    if not USE_TAPROOT:
+        assert anc_vsize == 161 and anc_fees == 339
```
jp1ac4, line 339: By removing these assertions about exact values, I'm no longer having any failures. I think it's OK to remove them as the assertions below with `additional_fee` will still be tested.

darosior: Indeed. I ran this test 200 times both on Taproot and Segwit v0 and didn't have a single failure. I'll drop those in a new commit and mark them as fixing #1000.

darosior:
Rebased on master. I found another crash using the fuzz tests which is now fixed. Unfortunately the compiler is insanely slow (like more than 1000x slower) for Taproot which makes the `descriptors` target useless.

src/descriptors/mod.rs
```diff
@@ -49,6 +53,31 @@ impl From<LianaPolicyError> for LianaDescError {
...
+        // Get the derivation path from the master fingerprint to the parent used to
+        // derive this key, in order to check whether it's part of the derivation paths
+        // used in this spending path (only checking the fingerprint isn't sufficient
+        // as a single signer may be used in more than one spending path).
+        // NOTE: this assumes there is only one derivation step after the key used in
+        // the policy. This is fine, because the keys in the policy are normalized (so
+        // the derivation path up to the wildcard is part of the origin).
+        if let Some((_, der_path_no_wildcard)) = der_path[..].split_last() {
+            if der_paths.contains(&der_path_no_wildcard.into()) {
+                return true;
```
edouardparis, line 77: `return der_paths.contains(&der_path_no_wildcard.into())`

src/descriptors/analysis.rs
```diff
@@ -357,6 +362,93 @@ impl PathInfo {
...
+        tap_tree
+            .iter()
+            .flat_map(|(_, ms)| ms.iter_pk())
+            .try_fold(Vec::new(), |mut acc, pk| {
+                let xkey = get_multi_xkey(&pk)?;
+                acc.extend_from_slice(&xkey.public_key.serialize());
+                Some(acc)
+            })?;
+    let chain_code: bip32::ChainCode = sha256::Hash::hash(&concat)[..]
+        .try_into()
```
edouardparis, line 420: `let chain_code = bip32::ChainCode::from(sha256::Hash::hash(&concat).as_ref());`

src/descriptors/mod.rs
darosior: Done.

src/descriptors/analysis.rs
darosior: Nice, done.

darosior:
ACK 4bb2372a63bb4b36ca9f0dd91d2373cff3e9e9a4 -- this underwent multiple rounds of review, Edouard tested it in his follow-up PR to the GUI and i wrote a couple fuzz targets exercising part of the logic introduced here

PR #986, opened 2024-02-27
Gui taproot

edouardparis:
based on #985

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -77,20 +77,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
 dependencies = [
  "cfg-if",
- "cipher 0.4.4",
+ "cipher",
  "cpufeatures",
-]
-
-[[package]]
-name = "aes-ctr"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7729c3cde54d67063be556aeac75a81330d802f0259500ca40cb52967f975763"
-dependencies = [
- "aes-soft",
- "aesni",
- "cipher 0.2.5",
- "ctr 0.6.0",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


darosior:
If not too complicated i think it's worth hiding the network and Taproot selectors under a collapsible "Advanced settings" options at the bottom right before the buttons.

As @kloaec rightly points out 99% people want mainnet, not signet or regtest, and the 1% can use advanced settings. Same goes for Taproot for now.

edouardparis:
20240311_18h20m46s_grim

darosior:
- Why are the advanced settings after the "Next" button?
- I get an error when signing with a simple "1 - 1" descriptor (signing using the primary key on the Ledger which is the internal Taproot key). Both for refresh and external payments.
- It seems change is not properly detected. Cf the screenshot. 
image


My descriptor: 
```
tr([b0822927/48'/1'/0'/2']tpubDEvZxV86Br8Knbm9tWcr5Hvmg5cYTYsg92vinqH6Bie6U8ix8CsoN9W11NQygdqVwmHUJpsHXxNsi5gXn36g4xNfLWkMqPuFhRZAmMQ7jjQ/<0;1>/*,and_v(v:pk([cafb6853/48'/1'/0'/2']tpubDF5GEQ4diRqBUs3hSq1CZ1eveX1yjHptGVEDH383xMJv52cwn2fzJcr3CDnDCRLrdBdRqA2jbggFgy7u9VBkmYW5qwVT8gzLQ4cYamqhMRr/<0;1>/*),older(65535)))#9w9v5fzt
```

darosior:
nit: i think the error message can just be "Taproot is not supported by this device". I think it would be clearer to a user.
image

darosior:
I can't click on next anymore after removing the Bitbox from the path though.

darosior:
When manually selecting coins, i can't click on "Next" when creating a spend.

darosior:
I don't think the bug above is related to this PR. I also found #1024 which i don't think is related to this PR.

darosior:
tACK b7f35c03300710c71f5e8a4af471260f0a35ddc0 -- i've lightly tested this a couple times. It's good enough to get in and get tested along with the other changes.

Issue #929, opened 2024-01-18
[GUI] Use `-rpcauth` option for Liana-managed bitcoind instead of cookie path


jp1ac4:
Following #890 and #898, we can now use the `-rpcauth` option when running the Liana-managed bitcoind and specify the user/password in the bitcoind section of Liana's config file.

This would avoid having to wait for the cookie file to appear when first starting bitcoind and also avoid problems in case the cookie file is deleted while bitcoind is running.

pythcoiner:
in which case `.cookie` can be deleted?

jp1ac4:
The only way I'm aware of is starting bitcoind twice in v26.0 (see https://github.com/bitcoin/bitcoin/pull/28784).

jp1ac4:
Regarding the cookie file being deleted while bitcoind is running, it might not be such a problem as we currently store the child process handle in the `Bitcoind` struct:
https://github.com/wizardsardine/liana/blob/3420c54793b1e132beebd2ea6add4d63fda446c1/gui/src/bitcoind.rs#L164

So maybe we could change the `stop` method to simply kill this process instead of doing it via RPC?
https://github.com/wizardsardine/liana/blob/3420c54793b1e132beebd2ea6add4d63fda446c1/gui/src/bitcoind.rs#L261

There may be other cases where we don't have the child process handle, such as in https://github.com/wizardsardine/liana/issues/915 and https://github.com/wizardsardine/liana/issues/924. In the latter case, we may not have the RPC user and password and so the cookie file may still be required.

pythcoiner:
> So maybe we could change the `stop` method to simply kill this process instead of doing it via RPC?

I guess the `stop` methop should be prefered when we can as incorrectly stop bitcoind can cause issues...
I especially don't know if there is a 'clean' way to kill a process on Windows

jp1ac4:
I tried changing it, but wasn't immediately sure how to deal with `Arc` and mutable references :thinking: In any case, I made a change in #988 to the `start()` function so that if there's some error creating the `BitcoinD` interface after starting bitcoind, it uses `process.kill()` before returning. So hopefully, if bitcoind is running, it means we're able to create an interface and will be able to stop it later on using `stop`.

darosior:
Note: #988 fixed this, but only for new wallets. Bitcoin Core 26.1 will have a fix for the most annoying cookie issue, but it won't be available to existing wallet too.

PR #987, opened 2024-02-29
lib: expose random module

jp1ac4:
This is motivated by #929. The GUI will need to generate a salt and password for RPC user/password authentication to the managed bitcoind, and can use this module to get random bytes.


darosior:
utACK daf60264d16a554f8877f28e11b5df6eab33d324

Issue #989, opened 2024-03-02
[Gui] accept args from the general section in bitcoin.conf


pythcoiner:
An user willing to use his own bitcoind hit this today:
![image](https://github.com/wizardsardine/liana/assets/124568858/1d895b49-6bb6-491e-ab0d-6fb5d334b645)

it seems to me his `bitcoin.conf` should miss `[main]` section, but why we do not use args if they are in the general section?

jp1ac4:
> why we do not use args if they are in the general section?

The bitcoin.conf file for managed bitcoind is auto-generated with an expected format. As we don't expect any user modifications, it can only parse the file correctly if it has this expected format. We should perhaps add something in the bitcoin.conf file to warn users that manual modifications may cause errors and be overwritten.

pythcoiner:
i guess this issue happend to an user that try to used its own bitcoind in place of the managed one:

![image](https://github.com/wizardsardine/liana/assets/124568858/c0c303d0-eb01-4948-b1b1-9fd9c3847457)

jp1ac4:
Hmm yeah, it might be they had selected the managed bitcoind option and then made some changes to the config file while importing the pruned node. We could in principle be flexible in the format the bitcoin.conf file takes, but it's simpler to assume a fixed format and I think if the user knows how to edit the file, then they might be better choosing the self-managed node option.

pythcoiner:
i guess the proper handling of this is document/example of bitcoin.conf

PR #990, opened 2024-03-02
wording

pythcoiner:
as pointed by #989 the error message was not explicit

### DIFF
--- a/gui/src/installer/step/bitcoind.rs
+++ b/gui/src/installer/step/bitcoind.rs
@@ -200,7 +200,7 @@ impl InternalBitcoindConfig {
                 );
             } else if !prop.is_empty() {
                 return Err(InternalBitcoindConfigError::UnexpectedSection(
-                    "General section should be empty".to_string(),
+                    "bitcoin.conf general section should be empty".to_string(),
                 ));
             }
         }


gui/src/installer/step/bitcoind.rs
```diff
@@ -200,7 +200,7 @@ impl InternalBitcoindConfig {
                 );
             } else if !prop.is_empty() {
                 return Err(InternalBitcoindConfigError::UnexpectedSection(
-                    "General section should be empty".to_string(),
+                    "bitcoin.conf general section should be empty".to_string(),
```
jp1ac4, line 203: It might be slightly better to mention the bitcoin.conf file in the `ReadingFile` error as returned by the `from_file` method below. You could also include the path to the file in this error variant's message. Then the "General section should be empty" message would be added at the end as additional info. For this, I think you'd need to wrap the error from `Self::from_ini(&conf_ini)` inside the `ReadingFile` error.

jp1ac4: Another option could be to mention the bitcoin.conf file in the `UnexpectedSection` error, which currently says "Unexpected section in file", and this message would also be shown for any other unexpected sections apart from the general section.

PR #991, opened 2024-03-03
[GUI] Move qr code to modal

### LINKED ISSUE #949: QRCode displayed only for last address
lurking kevin [here](https://youtu.be/7F-ibKh7fEU?t=4911), it appear that the QRCode cannot be dispayed on oldest address if we generate several, should be good to add 'dispay qr code' button i think.

pythcoiner:
fix #949:
- Remove Qr code from view
- Add a button `Show QR Code` that display the QRCode in a new modal
- Increased size of Qr Code as we now have more room

image

image

### DIFF
--- a/gui/src/app/state/receive.rs
+++ b/gui/src/app/state/receive.rs
@@ -27,6 +27,12 @@ use crate::daemon::{
     Daemon,
 };
 
+pub enum Modal {
+    VerifyAddress(VerifyAddressModal),
+    ShowQrCode(ShowQrCodeModal),
+    None,
+}
+
 #[derive(Debug, Default)]
 pub struct Addresses {
     list: Vec<Address>,
@@ -57,8 +63,7 @@ pub struct ReceivePanel {
     wallet: Arc<Wallet>,
     addresses: Addresses,
     labels_edited: LabelsEdited,
-    qr_code: Option<qr_code::State>,
-    modal: Option<VerifyAddressModal>,
+    modal: Modal,
     warning: Option<Error>,
 }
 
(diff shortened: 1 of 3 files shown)


gui/src/app/state/receive.rs
```diff
@@ -154,12 +155,20 @@ impl State for ReceivePanel {
                 ));
                 Command::none()
             }
+            Message::View(view::Message::ShowQrCode(i)) => {
+                let address = self.addresses.list.get(i).expect("Must be present").clone();
```
edouardparis, line 159: It is unexpected to happen, but it is safer to do the `if let Some(address) = self.address.get(i).as_ref()`
(I think `as_ref` will help us to not use `.clone()`)

gui/src/app/state/receive.rs
```diff
@@ -154,12 +155,20 @@ impl State for ReceivePanel {
...
+                let qr_code = qr_code::State::new(format!("bitcoin:{}?index={}", address, i)).ok();
```
edouardparis, line 160: I would move this code in the `ShowQrCodeModal::new(address, index) -> Option<ShowQrCodeModal>`
and do `self.modal = ShowQrCodeModal::new(address, i)`

edouardparis (commented):
I like the enum way of the change :+1:

jp1ac4:
Do you think it would be worth also showing the address below the QR code in the modal? It could be reassuring to the user in case several addresses have been generated on the screen and they forget which one they clicked on.

pythcoiner:
adressed comments, here how it looks now:

qr_modal

edouardparis:
Seems good, only need a rebase :+1:

pythcoiner:
rebased on master

edouardparis (approved):
ACK e38c5f123221ad33b9dc116a9c4d9bc269c518e2

PR #992, opened 2024-03-03
[GUI] Limit amount to 8 decimal digits

### LINKED ISSUE #798: Limit the send amout value field to 8 decimals

pythcoiner:
This PR replace #977 and fixes #798 
I've followed jp1ac4 advice and create a `Form::new_amount_btc()` method that act like `Form::new_trimmed()` ~w/ few more (filtering) features:~
~- allow only input of "0123456789,." characters~
~- "," is replaced by "."~
~- only one "." separator~
~- maximum 8 digit after separator~

~these features works for keyboard input or paste~

~i've tryied to add a "0" in case the String start with a "." but this make the cursor have a bad location and i do not find a way to control the cursor location w/ iced, so i revert this feature~

### DIFF
--- a/gui/ui/src/component/form.rs
+++ b/gui/ui/src/component/form.rs
@@ -1,3 +1,4 @@
+use bitcoin::Denomination;
 use iced::{widget::text_input, Length};
 
 use crate::{color, component::text, theme, util::Collection, widget::*};
@@ -57,6 +58,29 @@ where
         Self {
             input: text_input::TextInput::new(placeholder, &value.value)
                 .on_input(move |s| on_change(s.trim().to_string())),
+            warning: None,
+            valid: value.valid,
+        }
+    }
+
+    /// Creates a new [`Form`] that restrict input values to valid btc amount before applying the
+    /// `on_change` function.
+    /// It expects:
+    /// - a placeholder
+    /// - the current value
+    /// - a function that produces a message when the [`Form`] changes
+    pub fn new_amount_btc<F>(placeholder: &str, value: &'a Value<String>, on_change: F) -> Self
+    where
+        F: 'static + Fn(String) -> Message,
+    {
+        Self {
+            input: text_input::TextInput::new(placeholder, &value.value).on_input(move |s| {
+                if bitcoin::Amount::from_str_in(&s, Denomination::Bitcoin).is_ok() || s.is_empty() {
+                    on_change(s)
+                } else {
+                    on_change(value.value.clone())
+                }
+            }),
             warning: None,
             valid: value.valid,
         }
--- a/gui/src/app/view/spend/mod.rs
+++ b/gui/src/app/view/spend/mod.rs
@@ -319,9 +319,9 @@ pub fn create_spend_tx<'a>(
 
 pub fn recipient_view<'a>(
     index: usize,
-    address: &form::Value<String>,
-    amount: &form::Value<String>,
-    label: &form::Value<String>,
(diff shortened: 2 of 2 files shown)


pythcoiner:
not related to this PR: should we add `gui/ui/Cargo.lock` to `.gitignore`?

jp1ac4:
Instead of using regex, another option might be to use `bitcoin::Amount::from_str_in` to determine if the new string can be parsed as a BTC amount and then use this to decide whether to pass this updated value to the `on_change` function or keep the previous value. I see we already have `bitcoin` as a dependency in the ui project so wouldn't need to add anything.

darosior:
I agree with @jp1ac4, let's not reinvent the wheel down here if they already have parsing upstream.

pythcoiner:
i've done adviced changes, not yet tested because my crash issue

gui/ui/src/component/form.rs
```diff
@@ -62,6 +64,29 @@ where
...
+    /// - a placeholder
+    /// - the current value
+    /// - a function that produces a message when the [`Form`] changes
+    pub fn new_amount_btc<F>(placeholder: &str, value: &'a Value<String>, on_change: F) -> Self
+    where
+        F: 'static + Fn(String) -> Message,
+    {
+        Self {
+            input: text_input::TextInput::new(placeholder, &value.value).on_input(move |s| {
+                if bitcoin::Amount::from_str_in(&s, Denomination::Bitcoin).is_ok() {
```
jp1ac4, line 79: I think it would also be good to allow an empty string so that a user can delete the value completely and start again.

gui/ui/src/component/form.rs
```diff
@@ -1,4 +1,6 @@
+use bitcoin::Denomination;
 use iced::{widget::text_input, Length};
+use regex::Regex;
```
jp1ac4, line 3: This can be removed now.

jp1ac4 (commented):
Changes are working well. Just had a couple of comments.

pythcoiner:
adressed comments & squash

edouardparis:
Please remove the unused regex dep. Thanks

pythcoiner:
> Please remove the unused regex dep. Thanks

oops, done & rebased on master

gui/ui/src/component/form.rs
```diff
@@ -62,6 +63,31 @@ where
...
+    /// - the current value
+    /// - a function that produces a message when the [`Form`] changes
+    pub fn new_amount_btc<F>(placeholder: &str, value: &'a Value<String>, on_change: F) -> Self
+    where
+        F: 'static + Fn(String) -> Message,
+    {
+        Self {
+            input: text_input::TextInput::new(placeholder, &value.value).on_input(move |s| {
+                if bitcoin::Amount::from_str_in(&s, Denomination::Bitcoin).is_ok()
+                    || s == *"".to_string()
```
jp1ac4, line 79: I think you could write this as  `s.is_empty()`.

pythcoiner: done

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/9415dc939080caa5bf41e93c8717dd0b75bb6a0e.

edouardparis (approved):
ACK 9415dc939080caa5bf41e93c8717dd0b75bb6a0e

PR #993, opened 2024-03-04
[Installer] add back button into installer first step

### LINKED ISSUE #306: GUI installer: add a previous button
Currently, when a wallet is already set, clicking on "Install Liana on another network" opens the installer, with no way to go back.
If you miss click, you have to close and open the app, that's bad.

Adding a previous button, only when the user came from the launcher screen, would fix this.

I tried but didn't find a way to do it, while keeping the current UI, with the 3 buttons staying in the center of the screen.
So either :
 - Find a way to do it probably by using another layout than Row or Column
 - Add the previous button in the middle of the screen too
 - Put everything on the top of the screen

image

pythcoiner:
fixes #306 

i'm not really satisfied about the button position (any advice welcome), but at leat its functionnal.


gui/src/installer/mod.rs
```diff
@@ -33,6 +33,7 @@ pub struct Installer {
     steps: Vec<Box<dyn Step>>,
     hws: HardwareWallets,
     signer: Arc<Mutex<Signer>>,
+    back_button: bool,
```
jp1ac4, line 36: Is the idea that if the user passes the network in the command line, then it would not be possible to go back? Perhaps in that case, this field could be changed to `network: Option<Network>` so that it stores this pre-selected network, if any, and we could also use it on the `DefineDescriptor` step to select this same network.

gui/src/installer/step/mod.rs
```diff
@@ -52,11 +52,19 @@ pub trait Step {
 }
 
 #[derive(Default)]
-pub struct Welcome {}
+pub struct Welcome {
+    path: Option<PathBuf>,
```
jp1ac4, line 56: This might affect the logic, but perhaps there's no need for `Option` here as it should always have a value.

Also, it might also be worth renaming the field to be consistent with other steps and/or the context object, e.g. `data_dir`?

gui/src/installer/view.rs
```diff
@@ -90,6 +90,15 @@ pub fn welcome<'a>() -> Element<'a, Message> {
             .push(
                 Container::new(
                     Column::new()
+                        .push_maybe(path.as_ref().map(|p| {
+                            Row::new()
+                                .push(Space::with_width(Length::Fixed(30.0)))
+                                .push(Container::new(
+                                    button::secondary(None, " Back ")
```
jp1ac4, line 97: Perhaps the Back button could be labelled "Previous" to be consistent with other steps in the installer?

jp1ac4 (commented):
I've tested this locally and it's working nicely, with the Back button only shown if I don't pass the network by command line.

I've added a couple of comments, but I think @edouardparis will be able to judge better the approach.

gui/src/installer/view.rs
pythcoiner: Yes!

gui/src/installer/mod.rs
pythcoiner: i was not aware we can pass network as cli arg, i need to look at this,

gui/src/installer/step/mod.rs
pythcoiner: We need an option, as on a fresh install (or no other wallet created) there is no meaning to come back to the `Launcher`, as we directly launching the installer...

darosior:
> (any advice welcome)

Maybe the line below the three "Select" buttons, in the same style as the select buttons, centered?

pythcoiner:
> > (any advice welcome)
> 
> Maybe the line below the three "Select" buttons, in the same style as the select buttons, centered?

you mean move here?

image

darosior:
I meant below.

pythcoiner:
here the `Previous` button moved below slects button:

previous

gui/src/installer/mod.rs
pythcoiner: just look at this, to me it look that if the user select a network from cli, why would it go back to choose an other network?

The logic actually implemented is:
- if Installer is launch at startup, do not display the `Previous` button ( we land there in 2 cases: no wallet configured yet or network pass by cli)
- if we land to Installer from the Launcher, display the `Previous` button

i can change the logic if you think we should alway can go back to Launcher

jp1ac4: That's OK, I think it makes sense not to show the Previous button in those cases.

nondiremanuel:
This has been done in #1098 
I'm going to close this PR

Issue #791, opened 2023-11-03
GUI: Installer: Drop "backup descriptor" step when recovering from descriptor


pythcoiner:
During restore wallet flow, we ask user to backup descriptor again just after he supplied it. It's weird, should we assume that he can restore this descriptor because he previously backed up it?

darosior:
Good point. We can drop the "backup descriptor" step during recovery.

darosior:
@jp1ac4 while you are tackling #866 would you address this one as well? They are pretty related and seem they could be tackled in the same PR.

jp1ac4:
Looking at this again, do we need to change anything? There's currently only a "Register descriptor" step, but I think we still need this one as the user may no longer have the descriptor registered on any devices.

pythcoiner:
humm, i think we already fix it in a previous commit, i close.

jp1ac4:
Reopening as per @pythcoiner's findings in https://github.com/wizardsardine/liana_manual_testing/blob/master/tests/v5/v5-pre-rc1/pyth.md.

jp1ac4:
This affects the "Participate in new wallet" steps.

PR #996, opened 2024-03-06
gui: hint at rescan when importing wallet

### LINKED ISSUE #866: Hint at rescanning when user imports descriptor
@kloaec pointed out to me people are confused to not see their coins after importing a descriptor. I thought we told them they'll have to rescan to see them, but looks like we don't. We should.

jp1ac4:
This is to resolve #866.

I've added the message to the first step where the user enters the descriptor:

image

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -414,7 +414,19 @@ pub fn import_descriptor<'a>(
                     } else {
                         None
                     })
-                    .push(col_descriptor),
+                    .push(col_descriptor)
+                    .push_maybe(if change_network {
+                        // only show message when importing a descriptor
+                        Some(text(
+                            "After creating the wallet, \
+                            you will need to perform a rescan of \
+                            the blockchain in order to see your \
+                            coins and past transactions. This can \
+                            be done in Settings > Bitcoin Core.",
+                        ))
+                    } else {
+                        None
+                    }),
             )
             .push(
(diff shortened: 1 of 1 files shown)


jp1ac4:
Changed to draft as I will also do #791 as part of this PR.

jp1ac4:
> Changed to draft as I will also do #791 as part of this PR.

That issue has now been closed.

darosior:
ACK 09cf4f4c1f9da5ec8593ea022b1ec58f2dbd05f3 based on the screenshot in the OP. Haven't tested nor reviewed the code.

gui/src/installer/view.rs
```diff
@@ -414,7 +414,19 @@ pub fn import_descriptor<'a>(
                     } else {
                         None
                     })
-                    .push(col_descriptor),
+                    .push(col_descriptor)
+                    .push_maybe(if change_network {
+                        // only show message when importing a descriptor
+                        Some(Row::new().push(text(
```
edouardparis, line 420: You can remove useless Row::new()

edouardparis (commented):
Tested, looks good :+1:

gui/src/installer/view.rs
jp1ac4: Ah yes, thanks for spotting. Removed that now.

edouardparis (approved):
ACK 492baeaabaf181523d0aac3ac1da8ce2ca837ece

PR #997, opened 2024-03-06
gui: add transaction redirection

edouardparis:
based on #959 
This PR removes the payment section about the transaction inputs outputs for a redirection to the payment transaction (The `See transaction details` button). It makes it easy to fee bump a payment by clicking on it to be redirected to the transaction to rbf

20240305_18h13m51s_grim

Maybe the UX/UI of the payment should be changed to remove more transaction information, but I fear it makes the panel a little bit empty.

### DIFF
--- a/gui/src/app/menu.rs
+++ b/gui/src/app/menu.rs
@@ -5,6 +5,7 @@ pub enum Menu {
     Receive,
     PSBTs,
     Transactions,
+    TransactionPreSelected(Txid),
     Settings,
     Coins,
     CreateSpendTx,
--- a/gui/src/app/mod.rs
+++ b/gui/src/app/mod.rs
@@ -84,6 +84,7 @@ impl Panels {
             Menu::Receive => &self.receive,
             Menu::PSBTs => &self.psbts,
             Menu::Transactions => &self.transactions,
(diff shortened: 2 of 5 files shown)


jp1ac4:
> Maybe the UX/UI of the payment should be changed to remove more transaction information, but I fear it makes the panel a little bit empty.

Did you consider going directly to the transaction view after the first click? Then the view with the "See transaction details" button wouldn't be required.

jp1ac4:
Not directly related to this PR, but perhaps more noticeable after clicking that there's a difference in wording between the pages. The Home page uses "Outgoing payment" while the Transactions page uses  "Outgoing transaction".

edouardparis:
> Did you consider going directly to the transaction view after the first click? Then the view with the "See transaction details" button wouldn't be required.

Payment view is different than transaction view

> The Home page uses "Outgoing payment" while the Transactions page uses "Outgoing transaction".

This is normal, home lists the payments, a line per transaction output, while transactions panel lists the transactions, a line per batch.

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/997/commits/0c1a1b8d6d6ead20fdb1af659cace229f814d9fb.

PR #1001, opened 2024-03-11
commands: add comment about immature coins

jp1ac4:
As a follow-up to #873, this adds a comment to make clear that immature coins are not included as candidates for auto-selection.


darosior (commented):
Great catch. You missed the SQL constraint there is (basically "change XOR immature" too). I don't remember what i was thinking but there must be a reason i had both this assertion and the SQL constraint. It was introduced in https://github.com/wizardsardine/liana/pull/578. We should carefully make sure there is no piece of logic relying on this. This late in the release cycle i would suggest we make the poller actually respect the constraint (never try to create a coin which is both change and immature) instead of removing the constraint. We could then discuss in an issue whether it's safe to get rid of the constraint.

Here is a functional test which triggers the assertion on master and the SQL constraint failure on this branch:
```diff
diff --git a/tests/test_misc.py b/tests/test_misc.py
index f4a81c5..5c4b90a 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -239,6 +239,19 @@ def test_coinbase_deposit(lianad, bitcoind):
         and coin["spend_info"] is not None
     )
 
+    # We must also properly detect coinbase deposits to a change address. We used to have
+    # an assertion that a coin cannot both be change and a coinbase deposit. Since change
+    # is determined by the address... Technically we can.
+    change_desc = lianad.multi_desc.singlepath_descriptors()[1]
+    change_addr = bitcoind.rpc.deriveaddresses(str(change_desc), [0, 0])[0]
+    bitcoind.rpc.generatetoaddress(1, change_addr)
+    wait_for(lambda: any(c["is_immature"] for c in lianad.rpc.listcoins()["coins"]))
+    coin = next(c for c in lianad.rpc.listcoins()["coins"] if c["is_immature"])
+    bitcoind.generate_block(100)
+    wait_for_sync()
+    coin = next(c for c in lianad.rpc.listcoins()["coins"] if c["outpoint"] == coin["outpoint"])
+    assert not coin["is_immature"] and coin["block_height"] is not None
+
 
 @pytest.mark.skipif(
     OLD_LIANAD_PATH is None or USE_TAPROOT, reason="Need the old lianad binary to create the datadir."
```

jp1ac4:
Thanks for spotting the SQL constraint that I missed and providing the functional test.

As discussed, I've now removed this constraint so that a coin can be both change and immature.

I've also added a DB migration script. I could perhaps perform a check during the migration that `SELECT * FROM coins INTERSECT SELECT * FROM coins_new` has the same number of rows as `coins`.

darosior (commented):
Looks good to me. Could you create an exhaustive unit test for the migration? Have coins in various states, migrate the DB and make sure everything stayed the same. You can take inspiration from the `v0_to_v2_migration` test. If you make this a v3 to v4 you shouldn't even have to carve out helpers since we didn't change the columns.

src/database/sqlite/mod.rs
```diff
@@ -2130,7 +2216,167 @@ CREATE TABLE spend_transactions (
     }
 
     #[test]
-    fn v0_to_v3_migration() {
+    fn v3_to_v4_migration() {
+        let secp = secp256k1::Secp256k1::verification_only();
+
+        // Create a database with version 3, using the old schema.
+        let tmp_dir = tmp_dir();
+        eprintln!("{}", tmp_dir.as_path().to_string_lossy());
```
darosior, line 2224: Trailing debug print?

src/database/sqlite/mod.rs
```diff
@@ -2130,7 +2216,167 @@ CREATE TABLE spend_transactions (
...
+                )
+                .unwrap(),
+                is_immature: false,
+                block_info: None,
+                amount: bitcoin::Amount::from_sat(23145),
+                derivation_index: bip32::ChildNumber::from_normal_idx(10).unwrap(),
+                is_change: false,
+                spend_txid: None,
+                spend_block: None,
+            };
```
darosior, line 2267: Let's make the state of this one different from the previous one? In order to test more states. For instance coin a could stay like this (unconfirmed) and coin b could have `block_info` set (confirmed).

src/database/sqlite/mod.rs
```diff
@@ -2130,7 +2216,167 @@ CREATE TABLE spend_transactions (
...
+                )
+                .unwrap(),
+                is_immature: false,
+                block_info: None,
+                amount: bitcoin::Amount::from_sat(23200),
+                derivation_index: bip32::ChildNumber::from_normal_idx(3401).unwrap(),
+                is_change: true,
+                spend_txid: None,
+                spend_block: None,
+            };
```
darosior, line 2293: Same here. This one is the same as coin c. Maybe make it being unconfirmed but spent by an unconfirmed tx by setting `spend_txid`?

darosior (commented):
I think we can exercise more states when migrating from v3 to v4. Sorry to nit, i'm just always a bit wary to touch the database.

src/database/sqlite/mod.rs
jp1ac4: I've deleted this line now.

src/database/sqlite/mod.rs
jp1ac4: Ah sorry, it wasn't clear before, but I start with all coins as unconfirmed and then confirm/spend them subsequently using methods of `SqliteConn`.

jp1ac4:
I've updated the coins to use distinct values for each one (amount, vout, derivation index) and added a new `coin_e`.

I've also added some comments to clarify the approach taken and the final state of the coins.

darosior:
My bad, i just had a quick skim through the code and overlooked you confirmed them using the connection.

darosior:
ACK f9bae9cc0814d909a72a800d9f6d7ced2f99f06d -- this is very nice. Again, great catch. And thanks for adding a more extensive unit test.

PR #1010, opened 2024-03-15
Update our state upon broadcasting a transaction

darosior:
Fixes https://github.com/wizardsardine/liana/issues/887.

This takes a couple commits from #909 but takes the approach from there in another direction: we don't externalize the poller, since only a single instance must be ran. Instead we properly keep track of the (up to) two threads we manage in the `DaemonHandle` and provide a way for a user of the library to check for errors in any of the threads.

This approach allows us to 1) communicate with the poller thread from inside the Liana library/daemon (here we leverage this to tell it to poll) 2) eventually (#909) expose all internal errors from the library to the user instead of panic'ing internally.

See the commit messages for details.

### DIFF
--- a/src/bitcoin/poller/mod.rs
+++ b/src/bitcoin/poller/mod.rs
@@ -1,60 +1,95 @@
 mod looper;
 
-use crate::{
-    bitcoin::{poller::looper::looper, BitcoinInterface},
-    database::DatabaseInterface,
-    descriptors,
-};
+use crate::{bitcoin::BitcoinInterface, database::DatabaseInterface, descriptors};
 
 use std::{
     sync::{self, atomic},
     thread, time,
 };
 
+use miniscript::bitcoin::secp256k1;
+
 /// The Bitcoin poller handler.
 pub struct Poller {
-    handle: thread::JoinHandle<()>,
-    shutdown: sync::Arc<atomic::AtomicBool>,
+    bit: sync::Arc<sync::Mutex<dyn BitcoinInterface>>,
+    db: sync::Arc<sync::Mutex<dyn DatabaseInterface>>,
+    secp: secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+    // The receive and change descriptors (in this order).
+    descs: [descriptors::SinglePathLianaDesc; 2],
 }
 
 impl Poller {
-    pub fn start(
+    pub fn new(
         bit: sync::Arc<sync::Mutex<dyn BitcoinInterface>>,
         db: sync::Arc<sync::Mutex<dyn DatabaseInterface>>,
-        poll_interval: time::Duration,
         desc: descriptors::LianaDescriptor,
     ) -> Poller {
-        let shutdown = sync::Arc::from(atomic::AtomicBool::from(false));
-        let handle = thread::Builder::new()
-            .name("Bitcoin poller".to_string())
-            .spawn({
-                let shutdown = shutdown.clone();
-                move || looper(bit, db, shutdown, poll_interval, desc)
-            })
-            .expect("Must not fail");
+        let secp = secp256k1::Secp256k1::verification_only();
+        let descs = [
+            desc.receive_descriptor().clone(),
+            desc.change_descriptor().clone(),
+        ];
 
-        Poller { shutdown, handle }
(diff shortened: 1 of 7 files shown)


src/commands/mod.rs
```diff
@@ -688,7 +690,18 @@ impl DaemonControl {
         let final_tx = spend_psbt.extract_tx_unchecked_fee_rate();
         self.bitcoin
             .broadcast_tx(&final_tx)
-            .map_err(CommandError::TxBroadcast)
+            .map_err(CommandError::TxBroadcast)?;
+
+        // Finally, update our state with the changes from this transaction.
+        let (tx, rx) = mpsc::channel();
+        if let Err(e) = self.poller_sender.send(PollerMessage::PollNow(tx)) {
+            log::error!("Error requesting update from poller: {}", e);
```
edouardparis, line 698: I think it is not useful to make the broadcast command wait for the return of the poller, the broadcast did happen since the bitcoind received the tx to broadcast, and the notification to poll now makes the lapse of time of "unsync" db very very short, but If you prefer it, I am ok with it

darosior: I'm not sure to follow. Not waiting for the completion of the poll introduces a significant race condition (we'll probably return before the poll is complete), which makes the whole point of doing this moot?

src/commands/mod.rs
```diff
@@ -688,7 +690,18 @@ impl DaemonControl {
         let final_tx = spend_psbt.extract_tx_unchecked_fee_rate();
         self.bitcoin
             .broadcast_tx(&final_tx)
-            .map_err(CommandError::TxBroadcast)
+            .map_err(CommandError::TxBroadcast)?;
+
+        // Finally, update our state with the changes from this transaction.
+        let (tx, rx) = mpsc::channel();
```
edouardparis, line 696: GUI will use commands in different threads. It would require a `SyncSender`

```suggestion
        let (tx, rx) = mpsc::sync_channel();
```

darosior: Done.

src/bitcoin/poller/mod.rs
```diff
@@ -1,60 +1,131 @@
...
+                Err(mpsc::RecvTimeoutError::Disconnected) => {
+                    log::error!("Bitcoin poller communication channel got disconnected. Exiting.");
+                    return;
+                }
+            }
+
+            // Don't poll until the Bitcoin backend is fully synced.
+            if !synced {
+                let progress = self.bit.sync_progress();
+                log::info!(
```
edouardparis, line 116: Maybe pass this log in debug, in liana-gui console we get a lot of:
```

  2024-03-21T10:36:36.628035Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 99.90% (187699 blocks / 187728 headers)

  2024-03-21T10:36:36.664726Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 99.90% (187705 blocks / 187728 headers)

  2024-03-21T10:36:36.704515Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 99.90% (187711 blocks / 187728 headers)

  2024-03-21T10:36:36.736193Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 99.90% (187711 blocks / 187728 headers)

  2024-03-21T10:36:36.770596Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 99.90% (187717 blocks / 187728 headers)

  2024-03-21T10:36:36.807980Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 99.90% (187723 blocks / 187728 headers)

  2024-03-21T10:36:36.845011Z  INFO liana::bitcoin::poller:116: Block chain synchronization progress: 100.00% (187728 blocks / 187728 headers)

```
While the sync

darosior: I tried not to change behaviour in this PR.

darosior: Hmm but behaviour did change, looks like recv_timeout doesn't block, or not for long enough.

darosior:
ACK 58c71c794a050a2df460aec241c90efe89cf9ae0 -- did another pass and Edouard tested this in the GUI.

PR #1011, opened 2024-03-15
Add test update_spend_taproot

edouardparis:


edouardparis:
After talking with @darosior update_spend is already tested in multiple fonctionnal tests

PR #1014, opened 2024-03-18
fix recovery panel reload

edouardparis:
Once a psbt was generated with the recovery panel
it is impossible for user to start again a recovery process He is redirected to the previous generated psbt because of the panel state being not reset.

This commit reset the panel state on reload.


edouardparis (commented):
Self-ACK 56d9dbb72d5cb4c41355deeb9455c61edc403235

PR #1016, opened 2024-03-19
gui: clear RBF modal if selected transaction changes

jp1ac4:
As a follow-up to #959, this clears the RBF modal if the selected transaction changes. e.g. after going to the replacement transaction from the RBF modal and then returning to the Transactions panel and selecting a different transaction.

### DIFF
--- a/gui/src/app/state/transactions.rs
+++ b/gui/src/app/state/transactions.rs
@@ -127,6 +127,14 @@ impl State for TransactionsPanel {
                 } else {
                     self.txs.get(i - self.pending_txs.len()).cloned()
                 };
+                // Clear modal if it's for a different tx.
+                if let Some(modal) = &self.create_rbf_modal {
+                    if Some(modal.tx.tx.txid())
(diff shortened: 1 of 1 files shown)


gui/src/app/state/transactions.rs
```diff
@@ -127,6 +127,19 @@ impl State for TransactionsPanel {
                 } else {
                     self.txs.get(i - self.pending_txs.len()).cloned()
                 };
+                // Clear modal if it's for a different tx.
+                if let Some(modal) = &self.create_rbf_modal {
+                    if modal.tx.tx.txid()
```
edouardparis, line 132: I would prefer:
```rust
if Some(modal.tx.tx.txid()) != self.selected_tx.map(|selected| selected.tx.txid())
```

jp1ac4: I've updated this now.

edouardparis (approved):
ACK afd55299ed2fda5d2ee19232cb6c67397acd9b49

PR #1018, opened 2024-03-19
Gui handle poller error

edouardparis:
based on #986


darosior:
I tested on a busy wallet with dozens (if not hundreds) of coins and the poll when broadcasting is basically instant.

I tested crashing the poller, the GUI would only show an error as a banner when changing screen (except on the homepage). I think it's good enough for now.

darosior:
ACK a7f2bccb9a443c78a9788eedce06e2d753f7d11e

PR #1019, opened 2024-03-19
[GUI] tab on TextInputs

pythcoiner:
this PR  introduce handling of new events in all views:
- `Tab` will move cursor to the next TextInput
- `Tab` + `Shift` will move the cursor to previous TextInput

### DIFF
--- a/gui/src/main.rs
+++ b/gui/src/main.rs
@@ -2,7 +2,14 @@
 
 use std::{error::Error, io::Write, path::PathBuf, process, str::FromStr};
 
-use iced::{executor, Application, Command, Settings, Subscription};
+use iced::{
+    event::{self, Event},
+    executor,
+    keyboard::{self, KeyCode},
+    subscription,
+    widget::{focus_next, focus_previous},
+    Application, Command, Settings, Subscription,
+};
 use tracing::{error, info};
 use tracing_subscriber::filter::LevelFilter;
 extern crate serde;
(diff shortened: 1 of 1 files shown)


gui/src/main.rs
```diff
@@ -279,11 +317,13 @@ impl Application for GUI {
                 State::Launcher(v) => v.subscription().map(|msg| Message::Launch(Box::new(msg))),
             },
             iced_native::subscription::events().map(Self::Message::Event),
+            self.key_subscription(),
```
edouardparis, line 320: self.key_subscription is using iced_native::subscriptions::events_with, which is the same subscription that the one line 319. Please merge them together like this:

```rust
          iced_native::subscription::events_with(|event, status| match (&event, status) {
                 (
                    Event::Keyboard(keyboard::Event::KeyPressed {
                        key_code: KeyCode::Tab,
                        modifiers,
                        ..
                    }),
                    event::Status::Ignored,
                ) => Some(Message::KeyPressed(Key::Tab(modifiers.shift()))),
                (
                    iced::Event::Window(iced_native::window::Event::CloseRequested),
                    event::Status::Ignored,
                ) => Some(Message::Event(event)),
                _ => None,
            }),
```

edouardparis (commented):
Very interesting :eyes:  Thanks for looking for cool feature like this in the iced examples

gui/src/main.rs
pythcoiner: done

edouardparis (approved):
ACK 158bdc8a70bcd1a5dae34a842e7ac0bb39c20127

PR #1020, opened 2024-03-20
descriptors: change detection under Taproot

darosior:
I remember thinking about this helper when working on #985, but for some reason i didn't address it there. Here it is.


edouardparis (approved):
utACK 52c3613568736a1f2b4ab57bffcaee31d3d85b35

PR #1022, opened 2024-03-20
gui: sort events and txs after push in the list

edouardparis:
when a pending tx is confirmed, it is pushed in the HistoryTransaction list at the end. It should be first in the list. We sort now the list in order to make sure it will be always the case.


edouardparis (commented):
Self-ACK ff7e30794b7c79c3aa237dde33f52e716703c434

PR #1023, opened 2024-03-20
gui: record updates to aliases

jp1ac4:
This is to fix https://github.com/wizardsardine/liana/issues/1017.

It adds the `wallet` to the `reload` trait method so that changes are recorded.

As part of this change, the `TransactionsPanel` now has a `wallet` field to be used when calling its own `reload` method.


edouardparis:
Note for later: one possible fix would have been to pass Arc<Wallet> in the `State::update` method and remove it from each state internal fields, but it would require a more invasive change than this current PR.

edouardparis (approved):
ACK 67ebde1432c5cea5afe34f4fa79e2e9a119875a6

PR #1031, opened 2024-03-25
gui: fix condition to create draft spend

jp1ac4:
This issue was encountered by @pythcoiner in https://github.com/wizardsardine/liana_manual_testing/blob/master/tests/v5/v5-pre-rc1/pyth.md#comments.

The `balance_available` value was not being refreshed when new coins were received and so it was sometimes not possible to click Next.

It's enough, however, to check the amount left to select is 0, and so both the `balance_available` and `total_amount` fields can be removed.

### DIFF
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -68,7 +68,6 @@ pub trait Step {
 }
 
 pub struct DefineSpend {
-    balance_available: Amount,
     recipients: Vec<Recipient>,
     /// If set, this is the index of a recipient that should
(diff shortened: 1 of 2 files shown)


edouardparis:
It was legacy code used before our usage of the daemon method response that gives the amount left to select.
Cool that your fix simplifies the code. Will test ASAP.

darosior:
Neat.

edouardparis (approved):
ACK 68f8dbc43ba00de04b1ee33e0e8f2bd77242b127

Issue #1030, opened 2024-03-25
api: add address usage status in the list_addresses response


edouardparis:

pythcoiner:
You mean `listaddresses retun something like?:
```
{
  "result": {
    "addresses": [
      {
        "receive": {
           "address": "tb1pxvpvz8pqx2a5qkskdczv24jzqw795zgf4amqgz56kmqcj9yfecwsxlzra7",
           "used: true,
        },
        "index": 0,
        "change": {
           "addresse": "tb1pnymemz7mxdclhdgg6cmwxmnafymurpj7z0cyaydtuzdz4npkanzqxf5shu"'
           "used": false,
        },
      },
     
    ]
  }
}
```

PR #1032, opened 2024-03-25
gui: remove used addresses from receive panel

### LINKED ISSUE #1026: [GUI] v5 master branch - addresses don't go away despite being used.
image

https://mempool.space/signet/address/tb1qtjv7ruwy57249j5g4g4jcwhctvwcnatc332jcl50yvtnzlch8cjs2vcxje

edouardparis:
If the receive panel is the current panel, then we check every 10s if  a coin exists that is a deposit to one of the generated address, if it is the case we remove the address from the list. We do also the same when reloading the receive panel (when user come back from an other panel).

It is possible to make it more efficient by using the `list_addresses` method if its response include the address usage status (See #1030).

close #1026

### DIFF
--- a/gui/src/app/message.rs
+++ b/gui/src/app/message.rs
@@ -26,6 +26,7 @@ pub enum Message {
     LoadWallet,
     WalletLoaded(Result<Arc<Wallet>, Error>),
     Info(Result<GetInfoResult, Error>),
+    CheckUsedAddresses,
     ReceiveAddress(Result<(Address, ChildNumber), Error>),
     Coins(Result<Vec<Coin>, Error>),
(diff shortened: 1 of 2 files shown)


darosior:
I disagree with the approach here. It seems like a bad hack to be honest. An address which was shared doesn't mean it was necessarily used onchain.

edouardparis:
Ok, I will need some guidance to fix #1026, should I just reset the list when user leave and come back to the receive panel ? We can continue discussion on the issue page.

darosior:
I guess yes? That's how it was before and i think it was fine.

darosior:
I think this can now be closed?

PR #1051, opened 2024-03-26
increase QrCode size

### LINKED ISSUE #1024: Truncated address when showing a receive address as a QR code
image

pythcoiner:
fixes #1024, i increased the qr code size in order  to let more room for address

image


pythcoiner:
superseded by #2155

PR #1054, opened 2024-03-26
gui: remove descriptor backup in installer when participating in new wallet

### LINKED ISSUE #791: GUI: Installer: Drop "backup descriptor" step when recovering from descriptor

jp1ac4:
This fixes #791 which is regarding the "Participate in new wallet" steps.

The previous step requires the descriptor and so no need to back it up.

I also fixed an issue where it was possible to click Next with an invalid network selected.


edouardparis (approved):
ACK 52cbc3ee386973c12d002da3a688266994261312

PR #1055, opened 2024-03-26
gui: make max tooltip more legible

jp1ac4:
This resolves an issue found by @kloaec. I've changed the tooltip position to `Bottom` instead of `Left` and added some blank spaces so that text is padded at screen edge.

image

Full screen:
image


darosior:
utACK 617845234c4e64e02b47d4e29971b094f38f5281

PR #1061, opened 2024-03-28
spend: reword warning messages

jp1ac4:
As discussed with @kloaec, this PR rewords the warning messages when creating a spend.

I didn't include the selected feerate value in the updated messages as that would have required more code changes, albeit fairly simple ones, beyond the message wording itself. I can add those here if required or otherwise in a follow-up.

Here's a screenshot showing the new warnings:

image

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1571,7 +1571,14 @@ mod tests {
         );
         assert_eq!(tx.output[0].value.to_sat(), 95_000);
         // change = 100_000 - 95_000 - /* fee without change */ 127 - /* extra fee for change output */ 43 = 4830
-        assert_eq!(warnings, vec!["Change amount of 4830 sats added to fee as it was too small to create a transaction output."]);
+        assert_eq!(
+            warnings,
+            vec![
+                "Dust UTXO. The minimal change output allowed by Liana is 5000 sats. \
+                Instead of creating a change of 4830 sats, it was added to the \
+                transaction fee. Select a larger input to avoid this from happening."
+            ]
+        );
 
         // Increase the target value by the change amount and the warning will disappear.
         *destinations.get_mut(&dummy_addr).unwrap() = 95_000 + 4_830;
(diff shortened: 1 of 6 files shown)


darosior:
I think it's good to not include the selected feerate value in the backport. Can you do that in a follow-up which we won't backport?

tests/test_spend.py
```diff
@@ -9,6 +9,15 @@
 )
 
 
+def additional_fees(anc_vsize, anc_fee, target_feerate):
+    """The additional fee which must have been computed by lianad."""
+    computed_anc_vsize = int(anc_fee / target_feerate)
+    print(f"c  ", computed_anc_vsize)
```
darosior, line 15: trailing debugging trace :)

EDIT: oh, it was in the original. We should still drop those prints IMO.

jp1ac4:
Sure, I'll create a separate issue to keep track of that.

src/commands/mod.rs
```diff
@@ -1571,7 +1571,14 @@ mod tests {
         );
         assert_eq!(tx.output[0].value.to_sat(), 95_000);
         // change = 100_000 - 95_000 - /* fee without change */ 127 - /* extra fee for change output */ 43 = 4830
-        assert_eq!(warnings, vec!["Change amount of 4830 sats added to fee as it was too small to create a transaction output."]);
+        assert_eq!(
+            warnings,
+            vec![
+                "Dust UTXO. The minimal change output allowed by Liana is 5000 sats. \
+                Instead of creating a change of 4830 sats, it was added to the \
+                transaction fee. Select a larger input to avoid this from happening."
```
darosior, line 1579: Not asking to change it. But do you find this new message clearer? I'm not convinced.

src/spend.rs
```diff
@@ -551,15 +551,20 @@ impl fmt::Display for CreateSpendWarning {
...
                 amt,
-                if *amt > 1 {"s"} else {""},
+                if *amt > 1 { "s" } else { "" },
             ),
             CreateSpendWarning::AdditionalFeeForAncestors(amt) => write!(
                 f,
-                "An additional fee of {} sat{} has been added to pay for ancestors at the target feerate.",
+                "CPFP: an unconfirmed input was selected. The current transaction fee \
+                was increased by {} sat{} to make the average feerate of both the input \
+                and current transaction equal to the selected feerate.",
```
darosior, line 565: This one is much clearer to a normie indeed. :)

darosior (commented):
Nice!

src/commands/mod.rs
jp1ac4: It's good that it now mentions the 5000 threshold, but otherwise it's quite similar to before I think. Using "dust UTXO" could be confusing, but like "CPFP" in the other message, it may help the user investigate further.

tests/test_spend.py
jp1ac4: Have dropped these now.

darosior:
ACK ed86696daf556533592b3285c70a2386a84ece71

PR #1064, opened 2024-04-01
gui: bump bitcoind to 26.1

darosior:


jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/1064/commits/3084808f0ce852040061b5ce392c53618367df64.

Compared values with those in https://bitcoincore.org/bin/bitcoin-core-26.1/SHA256SUMS. Tested with a new wallet installation and then using existing wallet with the new version.

PR #1070, opened 2024-04-01
doc/USAGE: point to our website for download

darosior:
No need to update this document for every single version anymore. Also, Github doesn't let us access how many people download the binaries.


darosior:
self-ACK 744d8f856cdf0f3e3aa94f26e85e49202516244e -- only doc

PR #1076, opened 2024-04-05
Bump bdk_coin_select to 0.3.0

### LINKED ISSUE #923: spend: bump bdk_coin_select dependency to 0.3.0
This will enable us to revert the temporary fix from https://github.com/wizardsardine/liana/pull/867.

Note that the min change value is no longer inclusive (see https://github.com/bitcoindevkit/coin-select/pull/14#discussion_r1439665103) and so the min value we set in the change policy will need to be decreased by 1.

jp1ac4:
This is to resolve #923.

The `score` method of the `LowestFee` metric has been fixed upstream and so our temporary partial fix from #867 is no longer required.

The `min_fee` parameter of `select_coins_for_spend`, if positive, now ensures that RBF rule 4 is satisfied. Accordingly, it has been renamed to `replaced_fee` and made an `Option`. We could potentially have used the `SpendTxFees` enum as a parameter directly instead of `feerate_vb` and `replaced_fee`, but `feerate_vb` is currently `f32` rather than `u64` and so I kept them as separate parameters.

Thanks to how the `replaced_fee` parameter works, the fee iteration approach used in `rbf_psbt` to ensure the replacement satisfies [RBF rule 4](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy) is no longer required.

`base_weight` is no longer stored in `CoinSelector` and instead the output weights are stored in `Target`. This means that the `output_weight` of `DrainWeights` no longer needs to take into account a potential change in output count varint as this is now handled by bdk_coin_select.

The min value for change no longer includes the threshold itself and so we have to subtract 1 from our change policy's min value (see https://github.com/bitcoindevkit/coin-select/pull/14#discussion_r1439665103). We already have a [test](https://github.com/wizardsardine/liana/blob/master/src/commands/mod.rs#L1653-#L1654) that fails without this subtraction as it expects a change output of 5000 sats.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,9 +64,9 @@ checksum = "35636a1494ede3b646cc98f74f8e62c773a38a659ebc777a2cf26b9b74171df9"
 
 [[package]]
 name = "bdk_coin_select"
-version = "0.1.1"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c0320167c3655e83f0415d52f39618902e449186ffc7dfb090f922f79675c316"
+checksum = "3c084bf76f0f67546fc814ffa82044144be1bb4618183a15016c162f8b087ad4"
 
 [[package]]
 name = "bech32"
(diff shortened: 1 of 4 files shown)


darosior (approved):
ACK 43ecd94a46a8abf82a6882e741aa1db2fd0abd7c

Great work as usual.

PR #1078, opened 2024-04-07
[WIP] Baaca - Your ledger Bitcoin companion

pythcoiner:
This PR Introduce a new binary: 'Baaca - Your ledger Bitcoin companion'
baaca is a tool that let user install firmware on a ledger device (S/S+/X).
 i've only test against S/S+ as i don't have a x in hand.

image

this application is separated in 2 parts:
- The Iced(12.1) front-end that run in the main thread:
    - The GUI implementation is contained in `baaca/src/gui.rs`
    - The GUI used the Liana theme that is at the moment in `baaca/src/theme.rs` and `baaca/src/color.rs` but should be linked to the Liana GUI `theme.rs` and `color.rs` after #597 will be merged
- Ledger backend that run in its own separated thread, this back-end largely reuse code from darosior [ledger_installer](https://github.com/darosior/ledger_installer):
    - This part is still messy, contained in 2 files (`baaca/src/ledger.rs` / `baaca/src/ledger_lib.rs` ), i need to rework some functions in order to have a more clean/clear API (maybe we should do that in a separate crate containing only the lib +  a CLI in order to be more reusable outside of Liana?)

they communicate via mesage through  `async-channels`

Fews comments about this implementation (that is still alpha):
- Update of firmware is deactivated (uncommenting this [line](https://github.com/pythcoiner/ledger_installer_gui/blob/304b7cb2edb8cb6bc3c2e8e97a0bd972948f538b/src/gui.rs#L252) will allow used to click on a `Try update` button that will barely overwrite the actal app on device by installing the latest one).
- The ledger thread try to connect to ledger devices every 2 seconds and stop when it get connected ( i let it poll continuously while idle before, but have sometimes weird behaviour )
- I do not try the behaviour if 2 ledger devices connected
- We rely on `minreq` as dependency, patched by dorosior because I hit an issue about IPV6 w/ his PoC.
- For both fetching the firmware and know the app version installed(w/ app not opened) we have to connect to the Ledger API, that can be at the moment very slow (i sometime experience 40 seconds wait, not sure if it is about ledger API side or a `minreq` issue)
- Wording and sequence  and sequence of info messages passed to user can be improved a lot (i'm very bad at wording :smile: )
- The binary have been tested only under linux at  the moment


darosior:
I think this should be left as a separate project. Probably at all, but at least for now for sure.

pythcoiner:
closing here, move [there](https://github.com/darosior/ledger_installer/pull/3)

PR #1080, opened 2024-04-18
guix: use Rust 1.70 for release builds

darosior:
This is prep work for #597, which bumps the MSRV of the GUI to 1.70.

We are being pulled over in two different directions when it comes to our reproducible builds. On the one hand we need to target reasonably old glibc versions in order to be compatible with older systems. On the other hand the immaturity of the Rust ecosystem makes us require bleeding edge versions of the compiler. With Guix to get the newer versions of the compiler we need to also bump the glibc version.

This was not a sustainable situation. I was planning for a long time to cleanup our reproducible builds. To create a proper Guix package for both the daemon and the GUI using the build system they provide. I had envisioned this way i could rewrite the inputs of the Guix package to use an older glibc, while being able to bump the Guix time-machine. It would even have allowed us to perform Windows builds inside Guix! And who knows i could even have attempted to perform Apple ones too.

Unfortunately it turned out to be more complicated than that. I couldn't manage to get my package to compile using an older glibc. Some details about some of my failed attempts can be found there: https://lists.gnu.org/archive/html/help-guix/2024-04/msg00056.html.

Instead of wasting more time on this, backport the newer Rust declarations from up-to-date Guix to Guix-of-our-time-machine.

### DIFF
--- a/contrib/reproducible/guix/guix-build.sh
+++ b/contrib/reproducible/guix/guix-build.sh
@@ -60,7 +60,7 @@ for project_folder in "" "gui"; do
     PROJECT_ROOT="$PWD/$project_folder"
     PROJECT_VENDOR_DIR="$VENDOR_DIR/$project_folder"
     PROJECT_OUT_DIR="$OUT_DIR/$project_folder"
-    PROJECT_PATCHES_ROOT="$PWD/contrib/reproducible/guix/patches/$project_folder"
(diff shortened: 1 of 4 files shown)


edouardparis (approved):
utACK 860a1ea07467788473f0a83edc4d7a02d8ad3c1a

darosior:
This is waiting on Edouard to test this.

darosior:
Our hashes match. :rocket:

PR #1081, opened 2024-04-19
GUI: change daemon trait to handle async methods

edouardparis:
This a preparatory work in order to have a lianalite backend connection to satisfy the Daemon trait.
This PR also allows us to use tokio::sync::Mutex for the embedded daemon which goes great with the iced tokio runtime.

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -2643,6 +2643,7 @@ name = "liana_gui"
 version = "5.0.0"
 dependencies = [
  "async-hwi",
+ "async-trait",
  "backtrace",
  "base64 0.21.6",
  "bitcoin_hashes 0.12.0",
--- a/gui/Cargo.toml
+++ b/gui/Cargo.toml
@@ -14,6 +14,7 @@ name = "liana-gui"
 path = "src/main.rs"
 
 [dependencies]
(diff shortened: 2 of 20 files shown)


gui/src/daemon/client/jsonrpc.rs
```diff
@@ -97,12 +88,14 @@ impl JsonRPCClient {
 
         debug!("Sending to lianad: {:#?}", request);
 
-        stream.write_all(&[serde_json::to_string(&request).unwrap().as_bytes(), b"\n"].concat())?;
+        stream
+            .write_all(&[serde_json::to_string(&request).unwrap().as_bytes(), b"\n"].concat())
+            .await?;
+
+        let mut buffer = Vec::<u8>::new();
+        stream.read_to_end(&mut buffer).await?;
```
jp1ac4, line 96: When testing with an external daemon, I found it was hanging on this step (calling `getinfo`) while loading the GUI.

I did a quick test adding `stream.shutdown().await?;` just before reading and that seemed to fix it, but not sure if that's the correct way.

jp1ac4 (commented):
I encountered an issue using an external daemon where it hangs on loading. Otherwise, the changes look good.

PR #1083, opened 2024-04-23
Add nixGL to gui/shell.nix

edouardparis:
Nixos 23.11 comes with libc 2.38, this version of libc may not be compatible with some drivers. For now the hack found is to use a community wrapper that detects the requirements and do the link (https://github.com/guibou/nixGL). usage:
nixGL cargo run


edouardparis (commented):
Self-ACK 9aa0b990a36716954ff30149982d1b5928cc3e57

PR #1084, opened 2024-04-25
Remove legacy harware_wallets field from gui config

edouardparis:
Ledger hmacs, wallet name and fingerprint labels where moved one year ago in the settings.json file.
Here the commit of the new settings module:
https://github.com/wizardsardine/liana/commit/bf1e9e4b808a7c275766975467e637b883524236

A second commit in v4 checked that settings.json is present otherwise it does the migration from the gui configuration file to the settings file:
https://github.com/wizardsardine/liana/commit/3eeba082483c06318ca7a1f018ce2c5da47f5f32

We remove the legacy field from the config field so the wallet module does not depend on the config module anymore to load hardware wallet information.

Tested on gui.toml config with legacy field hardware_wallets and it does not fail to launch.

### DIFF
--- a/gui/src/app/config.rs
+++ b/gui/src/app/config.rs
@@ -1,4 +1,3 @@
-use crate::hw::HardwareWalletConfig;
 use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
 use tracing_subscriber::filter;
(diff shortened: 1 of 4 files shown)


jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/pull/1084/commits/45c7b850a64318152e245b269ffef25d63033384.

I didn't do any specific testing, but checked I could open an existing wallet and create a new one.

PR #1092, opened 2024-05-21
spend: randomized anti-fee sniping

### LINKED ISSUE #44: Basic anti fee sniping for Spend transactions
https://github.com/revault/minisafe/blob/ac6e0443ea70207ad25aa38e142ca30c25b70d66/src/commands/mod.rs#L298

jp1ac4:
This is to resolve #44.

In the first commit, I combine the destinations and change parameters in `spend::create_spend` in order to avoid "too many arguments" error when adding a new parameter. I think these two parameters combine naturally so that `destinations` includes both recipients and change address.

The second commit sets locktime following the same approach as Bitcoin Core:
- locktime is set to current tip height, but randomly (about 10% of cases) value is set up to 100 blocks earlier
- if tip is more than 8 hours old, locktime is set to 0

For randomness, I'm currently using the current time's milliseconds in order not to add another dependency, which I think is good enough for this use case.

For consistency, I decided to use "locktime" everywhere instead of "lock time" or "lock-time".

### DIFF
--- a/src/spend.rs
+++ b/src/spend.rs
@@ -4,6 +4,7 @@ use std::{
     collections::{BTreeMap, HashMap},
     convert::TryInto,
     fmt,
+    time::Duration,
 };
 
 pub use bdk_coin_select::InsufficientFunds;
@@ -33,6 +34,10 @@ pub const MAX_FEE: bitcoin::Amount = bitcoin::Amount::ONE_BTC;
 
 /// Assume that paying more than 1000sat/vb in feerate is a bug.
 pub const MAX_FEERATE: u64 = 1_000;
+
+/// Do not set locktime if tip age in seconds is older than this.
+// See also https://github.com/bitcoin/bitcoin/blob/ecd23656db174adef61d3bd753d02698c3528192/src/wallet/spend.cpp#L906.
+pub const MAX_ANTI_FEE_SNIPING_TIP_AGE_SECS: u64 = 8 * 60 * 60; // 8 hours
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum InsaneFeeInfo {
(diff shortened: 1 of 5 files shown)


src/spend.rs
```diff
@@ -497,6 +502,37 @@ fn derived_desc(
...
+    now: Duration,
+    tip_height: u32,
+    tip_time_secs: Option<u32>,
+) -> LockTime {
+    tip_time_secs
+        .map(|tip_time| now.as_secs().saturating_sub(tip_time.into()))
+        .filter(|tip_age| *tip_age <= MAX_ANTI_FEE_SNIPING_TIP_AGE_SECS)
+        .map(|_| {
+            // Randomly set locktime further back using current time as source of randomness.
+            let millis = now.subsec_millis();
```
darosior, line 523: Would there be an advantage to use more precision? Probably truncation, which would be significant for instance between one thousand and one billion (nano).

A quick test
```rust
println!("{} - {} - {}", a.subsec_nanos(), a.subsec_micros(), a.subsec_millis());
```

Gives me:
```
140062106 - 140062 - 140
```

So it's probably use the most precision available?

darosior:
> in order to avoid "too many arguments" error when adding a new parameter

This is not a good reason. Maybe we should refactor this code, for instance using the common "builder pattern", but using a less-readable tuple to quiet the warning is worse than disabling the warning. I think it's fine to use `#![allow(clippy::too_many_arguments)]` for now.

pythcoiner:
> * locktime is set to current tip height, but randomly (about 10% of cases) value is set up to 100 blocks earlier

the main reason for this is against fingerprinting?

src/spend.rs
jp1ac4: I initially considered using micros, but thought millis might be better as most systems should have at least this level of precision. If using a higher precision, I could make sure the logic is such that in case of truncation, we would set the locktime to the current height.

jp1ac4:
> This is not a good reason. Maybe we should refactor this code, for instance using the common "builder pattern", but using a less-readable tuple to quiet the warning is worse than disabling the warning. I think it's fine to use `#![allow(clippy::too_many_arguments)]` for now.

That's fair enough :) I'll drop that commit and disable the warning.

> the main reason for this is against fingerprinting?

This is to improve the privacy of other services that may be delayed in broadcasting. See also:
https://github.com/bitcoin/bitcoin/blob/413844f1c2a3d8f7cfef822f348f26df488b03c7/src/wallet/spend.cpp#L947-L950

jp1ac4:
I've now reverted the change to `create_spend` parameters and used nano precision to calculate the height delta.

src/spend.rs
```diff
@@ -497,6 +502,40 @@ fn derived_desc(
     desc.derive(coin.deriv_index, secp)
 }
 
+/// Get value to use for transaction nLockTime in order to
+/// discourage fee sniping.
+///
+/// The approach follows that taken by Bitcoin Core:
+/// - most of the time, the value returned will be the current
+/// block height, but will randomly be up to 100 blocks earlier.
```
darosior, line 510: If anyone's wondering why we set the current block height and not the next block height (which is the one we want to be the first valid), it's because by Bitcoin consensus `nLockTime` is not the first height at which this transaction is valid but the last height at which it is invalid.

src/spend.rs
```diff
@@ -497,6 +502,40 @@ fn derived_desc(
...
+        .map(|tip_time| now.as_secs().saturating_sub(tip_time.into()))
+        .filter(|tip_age| *tip_age <= MAX_ANTI_FEE_SNIPING_TIP_AGE_SECS)
+        .map(|_| {
+            // Randomly (approx 10% of cases) set locktime further back
+            // using current time as source of randomness.
+            let nanos = now.subsec_nanos();
+            // Note this condition will fail if nano precision is not available
+            // and so nothing will be subtracted.
+            let delta = if nanos % 10 == 1 {
+                (nanos % 1000) / 10 + 1 // a number in [1, 100]
```
darosior, line 528: Why `% 1000 / 10` and not `% 100`?

darosior (commented):
ACK f2791744d8ad3068d12a39bb07f3745012785a5f

src/spend.rs
jp1ac4: This ensures we can get all values in `[1, 100]`. Given the condition `nanos % 10 == 1`, using `% 100` would limit the set of possible values.

src/spend.rs
jp1ac4: I was wondering this when first starting on the issue :slightly_smiling_face: I should have added this comment in the code.

src/spend.rs
darosior: Ugh, of course! Thanks.

PR #1097, opened 2024-05-28
gui: fix rbf feerate input form

### LINKED ISSUE #860: GUI: RBF: improve feerate text input
> It might be better not to trim the feerate value as it's difficult to edit a single-digit value. It's not possible to delete the existing value and enter a new one and we have to instead add a second digit and delete the first.

_Originally posted by @jp1ac4 in https://github.com/wizardsardine/liana/pull/852#discussion_r1419802693_

jp1ac4:
This fixes #860.

It makes it possible to edit a single-digit feerate, for example, by deleting the existing value and then entering a new one.


edouardparis (approved):
ACK 3b604093ce13a6727b5201c59b9af014f28d329f

PR #1098, opened 2024-05-29
Set network in launcher

### LINKED ISSUE #306: GUI installer: add a previous button

edouardparis:
This Pull requests introduce three changes:

1. Network can only be chosen with the launcher panel. A drop down showing what network user can install is shown to the user when clicking on install a new network.
2. A previous button is added to the first installer page, so user can be redirected to launcher and the network choices if he has chosen the wrong network. close #306 
3. In order to make it clear for the user which network he is actually using or he is currently installing a blue banner is added on top of the window if the network is different than the bitcoin mainnet.

20240529_13h01m15s_grim
20240529_13h01m27s_grim
20240529_13h01m02s_grim
20240529_13h00m35s_grim

### DIFF
--- a/gui/src/app/mod.rs
+++ b/gui/src/app/mod.rs
@@ -20,7 +20,10 @@ use tokio::runtime::Handle;
 use tracing::{error, info, warn};
 
 pub use liana::{commands::CoinStatus, config::Config as DaemonConfig, miniscript::bitcoin};
-use liana_ui::widget::Element;
+use liana_ui::{
+    component::network_banner,
+    widget::{Column, Element},
+};
 
 pub use config::Config;
 pub use message::Message;
@@ -318,6 +321,11 @@ impl App {
     }
 
     pub fn view(&self) -> Element<Message> {
-        self.panels.current().view(&self.cache).map(Message::View)
+        let content = self.panels.current().view(&self.cache).map(Message::View);
+        if self.cache.network != bitcoin::Network::Bitcoin {
+            Column::with_children(vec![network_banner(self.cache.network).into(), content]).into()
+        } else {
+            content
+        }
     }
 }
--- a/gui/src/installer/message.rs
+++ b/gui/src/installer/message.rs
@@ -1,7 +1,4 @@
-use liana::miniscript::{
-    bitcoin::{bip32::Fingerprint, Network},
-    DescriptorPublicKey,
-};
+use liana::miniscript::{bitcoin::bip32::Fingerprint, DescriptorPublicKey};
 use std::path::PathBuf;
 
 use super::Error;
@@ -29,7 +26,6 @@ pub enum Message {
     Select(usize),
     UseHotSigner,
     Installed(Result<PathBuf, Error>),
-    Network(Network),
     CreateTaprootDescriptor(bool),
     SelectBitcoindType(SelectBitcoindTypeMsg),
     InternalBitcoind(InternalBitcoindMsg),
--- a/gui/src/installer/message.rs
+++ b/gui/src/installer/message.rs
@@ -20,6 +20,7 @@ pub enum Message {
     Next,
     Skip,
     Previous,
+    BackToLauncher,
     Install,
     Close,
     Reload,
--- a/gui/src/installer/mod.rs
+++ b/gui/src/installer/mod.rs
@@ -73,6 +73,10 @@ impl Installer {
         )
     }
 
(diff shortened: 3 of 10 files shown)


darosior:
Here: 
image

nit: i think it would be clearer if the two sections were:
- Open an existing wallet
- Create a new wallet

It's already obvious from the options the user has to select a network. I think it's more helpful to tell the user what the selection would achieve.

jp1ac4:
It looks very nice :heart_eyes:

After deleting an existing wallet, the list of networks for a new install does not refresh to include the deleted network.

It would be nice to refresh this list even while the modal is open:
image

Regarding the "Change network" button, this is also covered as part of #993. There are some other special cases to consider, e.g. if the network has been specified in the command line, it may not make sense to show the button.

The only other thing I found is that the default window size may make the list not display nicely, e.g.
image

This is resolved if the window is made bigger.

darosior:
This is the screen i get on the first launch on a fresh datadir:
image

We shouldn't tell "Welcome back". Only "Welcome" works i guess?

We should not have the "Select network" section, since there is no existing wallet. We should have the "Install on mainnet" by default, with a small line inviting to install on another network (by selecting which one from a dropdown). As per the mockups:
image

Also had the same issue as Michael above:
image

In addition, again "Select network" should not be there if there is no installed wallet.

nondiremanuel:
I would propose adding a back symbol ("<" or something equivalent) in the "Change Network" button to indicate that, by clicking it, the user would be redirected to the previous step.

jp1ac4:
Deletion of existing wallet works nicely now and the list of networks is displaying well on the default window size.

I found that if I use a new data directory, I get an error in the logs when selecting the network to use:
```
ERROR liana_gui::logger: Failed to change logger settings: Io(
    Os {
        code: 2,
        kind: NotFound,
        message: "No such file or directory",
    },
)
```
and then after completing the installer:
```
ERROR liana_gui::logger:102: Failed to remove installer log file /home/michael/test/liana/gui/abcdef/installer.log error:Os {
    code: 2,
    kind: NotFound,
    message: "No such file or directory",
}
```

darosior:
Sorry for nitpicking, but now it's not very clear what's happening.
image

Could this read "Install Liana on Bitcoin mainnet"? Or better yet not mention mainnet at all, some would not know what "mainnet" is (i can already see from here the question "is that a shitcoin?").

Ideally i think we would have something like a button which reads something like "Create a Liana wallet" with like a right-facing arrow. And then below, in a smaller font, some text which reads "Select a different network". Clicking on this text would expand the list of networks. Selecting one of the network would collapse the list and change the Bitcoin logo in the main "Create a Liana wallet" button. What do you think? Am i making sense?

Alternatively, we can try going for something simpler which wouldn't have you re-implement multiple components. But i'm not sure what. Right now it's very confusing. The user can't understand what clicking on the button is going to achieve for them.

edouardparis:
> Ideally i think we would have something like a button which reads something like "Create a Liana wallet" with like a right-facing arrow. And then below, in a smaller font, some text which reads "Select a different network". Clicking on this text would expand the list of networks. Selecting one of the network would collapse the list and change the Bitcoin logo in the main "Create a Liana wallet" button. What do you think? Am i making sense?

Only in the case of it is a fresh install and no other wallet is created ?
I think it adds one useless clicks, once the list of network is expanded, user should just click on it to start the install.

darosior:
Maybe. Do you have another suggestion? I don't think users should see "mainnet". And they should know what clicking on the button is going to do.

I'm also trying to think through an alternative which would be less time consuming to implement. Do you have an idea? Maybe just a label before the "Bitcoin mainnet" button which reads "Create a new Liana wallet:"? This is ugly, we keep mainnet, but it's simple.

jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/pull/1098/commits/c0e7b63809fd60f4fc6f5e9cd5223c138e25cfde.

PR #1096, opened 2024-05-26
Add a warning in installer if mainnet wallet already exists

pythcoiner:
this PR fixes #1039, it add a warning in the installer if an wallet already exists on mainnet & grey-out `Next` button if the network  is not valid.

image

PR #1105, opened 2024-06-06
fix ui: wrong usage of widget state

edouardparis:
In a previous PR #1096, we made usage
of the widget state instead of the component state to set the collapsed value when the widget is created. But the widget is reset at each refresh of the interface, instead the component state does not change.

This commit makes the distinction between the two by having the widget state as an initial state and the component one as a capture of the user clicks.


jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/pull/1105/commits/a9cfe45c0a010e39cb2fdd0b59926be5c13f2a4e.

PR #1115, opened 2024-06-11
Rename remote watchonly wallet

### LINKED ISSUE #1059: [Remote bitcoind] wallet name is create in /home/<user>/.liana
liana

here my username on laptop is `pythliana`, bitcoind on remote running under `pyth` user, so we ask bitcoind to create the wallet at `/home/pythliana/.liana` on remote, but it's out of remote userspace so it fails

pythcoiner:
This PR fixes #1059 
The watchonly wallet name was the local path, now its `liana-<checksum>` (only on remote backend).
No migration is handled in case a legacy wallet already on the remote backend, and it should be backward compatible as we do not remove the legacy wallet.
Nothing change if we connect to local backend.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -230,7 +230,24 @@ fn setup_bitcoind(
         .bitcoind_config
         .as_ref()
         .ok_or(StartupError::MissingBitcoindConfig)?;
-    let bitcoind = BitcoinD::new(bitcoind_config, wo_path_str)?;
+
+    let is_remote = !bitcoind_config.addr.ip().is_loopback();
+
+    // Extract wallet fingerprint and append it to the wallet name
+    let descriptor_str = config.main_descriptor.clone().to_string();
+    let descriptor_fingerprint = descriptor_str
+        .split('#')
+        .last()
+        .expect("Descriptor always has a fingerprint");
+    let new_wo_path_str = format!("liana-{}", descriptor_fingerprint);
+
+    let new_wo_path = if is_remote {
+        new_wo_path_str.clone()
+    } else {
+        wo_path_str.clone()
+    };
+
+    let mut bitcoind = BitcoinD::new(bitcoind_config, new_wo_path)?;
     bitcoind.node_sanity_checks(
         config.bitcoin_config.network,
         config.main_descriptor.is_taproot(),
@@ -239,14 +256,27 @@ fn setup_bitcoind(
         log::info!("Creating a new watchonly wallet on bitcoind.");
         bitcoind.create_watchonly_wallet(&config.main_descriptor)?;
         log::info!("Watchonly wallet created.");
-    } else {
+    } else if !is_remote {
         #[cfg(windows)]
         if !cfg!(test) && !wo_path.exists() {
             return Err(StartupError::NoWatchonlyInDatadir);
         }
     }
+
     log::info!("Loading our watchonly wallet on bitcoind.");
-    bitcoind.maybe_load_watchonly_wallet()?;
+    if let Err(e) = bitcoind.maybe_load_watchonly_wallet() {
+        // here if is_remote and errored we blindly fallback to legacy naming in order
+        // to handle unexpected edge cases like user setup w/ 127.0.0.1
+        // and later change it to its local 192.168.x.x ....
+        if is_remote {
+            log::info!("Loading legacy wallet.");
+            bitcoind = BitcoinD::new(bitcoind_config, wo_path_str)?;
+            bitcoind.maybe_load_watchonly_wallet()?;
+        } else {
+            Err(e)?
+        }
+    }
+
     bitcoind.wallet_sanity_checks(&config.main_descriptor)?;
     log::info!("Watchonly wallet loaded on bitcoind and sanity checked.");
 


pythcoiner:
I' actually testing this PR:
- Linux Wallet + remote backend (linux):
  - [x] create single wo wallet
  - [x] can create several wo wallets 
  - [ ] migration
- Linux Wallet + managed bitcoind:
  - [fail] create new wallet:
image
  - [ ] migration
- Win 10 Wallet + remote backend (linux):
  - [fail] create single wo wallet
  - [ ] create several wo wallets
  - [ ] migration
- Win10 Wallet + managed bitcoind:
  - [ ] create new wallet
  - [ ] migration
- MacOS Sonoma + remote backend (linux):
  - [ ] create single wo wallet
  - [ ] create several wo wallets
  - [ ] migration
- MacOS Sonoma Wallet + managed bitcoind:
  - [ ] create new wallet
  - [ ] migration

pythcoiner:
migration on regtest fail both for linux / macOS:
on `regtest`:
image

pythcoiner:
In case of migration, it trigger a rescan, i'm not sure if we should communicate this to the GUI?

pythcoiner:
Tested on:
- [x] Linux
- [x] macOS (VM)
- [x] Win 10 (VM)

jp1ac4:
> The watchonly wallet name was the local path, now its `liana-<checksum>` (only on remote backend).

I think this makes sense in case the same remote node is used for multiple Liana wallets.

> The migration is handled in case a legacy wallet already on the remote backend, and it should be backward compatible as we do not remove the legacy wallet.

I wonder if there could be problems in case the remote node is pruned, for example. Would it be possible to continue using the legacy wallet if it exists on the remote node? Perhaps for an existing Liana wallet with remote node, you could first try to load a wallet with the new name and then if it fails, a wallet with the legacy name? For new Liana wallets, you can then use the new name only.

pythcoiner:
> Perhaps for an existing Liana wallet with remote node, you could first try to load a wallet with the new name and then if it fails, a wallet with the legacy name

atm, if it fails to load the 'new' name and if there is a legacy wallet, we create one w/ new name + rescan, but you got a good point w/ a pruned node, we can ends up w/ the new wallet 'lost' coins older then prune point...

It's looks more safe to me to not migrate at all:
- if legacy wallet on the node we keep using it
- if no legacy wallet we use the new name

jp1ac4:
Yes, I think it would be safer not to migrate existing wallets for now.

pythcoiner:
> Yes, I think it would be safer not to migrate existing wallets for now.

done.

src/lib.rs
```diff
@@ -239,14 +258,28 @@ fn setup_bitcoind(
...
     }
     log::info!("Loading our watchonly wallet on bitcoind.");
-    bitcoind.maybe_load_watchonly_wallet()?;
+    if is_remote {
+        if let Err(e) = bitcoind.maybe_load_watchonly_wallet() {
+            let wallets = bitcoind.list_wallets();
+            // if a wallet w/ legacy name exist
+            if wallets.contains(&wo_path_str) && !wallets.contains(&new_wo_path_str) {
+                log::info!("Loading legacy wallet.");
+                bitcoind.change_watchonly_wallet_path(wo_path_str);
```
jp1ac4, line 274: I think it would be safer to use `BitcoinD::new()` when changing the wallet name as it also performs connection checks.

src/lib.rs
```diff
@@ -230,7 +232,24 @@ fn setup_bitcoind(
...
-    let bitcoind = BitcoinD::new(bitcoind_config, wo_path_str)?;
+
+    let is_remote = bitcoind_config.addr.ip() != IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+
+    // Extract wallet fingerprint and append it to the wallet name
+    let descriptor_str = config.main_descriptor.clone().to_string();
+    let descriptor_fingerprint = descriptor_str
+        .split('#')
+        .last()
+        .expect("Descriptor have always a fingerprint");
```
jp1ac4, line 243: nit: I would rephrase to "Descriptor always has a fingerprint"

src/lib.rs
```diff
@@ -239,14 +258,28 @@ fn setup_bitcoind(
...
             return Err(StartupError::NoWatchonlyInDatadir);
         }
     }
     log::info!("Loading our watchonly wallet on bitcoind.");
-    bitcoind.maybe_load_watchonly_wallet()?;
+    if is_remote {
+        if let Err(e) = bitcoind.maybe_load_watchonly_wallet() {
+            let wallets = bitcoind.list_wallets();
+            // if a wallet w/ legacy name exist
+            if wallets.contains(&wo_path_str) && !wallets.contains(&new_wo_path_str) {
```
jp1ac4, line 272: To avoid repeating logic, could you call `maybe_load_watchonly_wallet()` directly without performing this check here?

Perhaps you can check instead if the first attempt failed because the wallet doesn't exist by checking the error code, which I think would be `-18`: https://github.com/bitcoin/bitcoin/blob/41544b8f96dbc9c6b8998acd6522200d67cdc16d/src/rpc/protocol.h#L81.

jp1ac4 (commented):
I think the approach looks good. I've added a few comments to consider.

pythcoiner:
comments adressed.

pythcoiner:
I'm testing this:
- [x] w/ a local bitcoind nothing change for both linux/win10/macOS(sonoma)
- [x] w/ linux client + remote bitcoind (linux):
  - [x] if a legacy wallet on remote we use it (no migration)
  - [x] if no legacy wallet on remote we create w/ new naming
- [x] w/ win10 client + remote bitcoind (linux)
  - [?] user can create a legacy wallet w/ master (b36220504fd895d69d59325eae6527708736b272) but liana fail at startup w/ this error:
 
image

then running this PR it will create a wallet w/ the new naming and works well, it is not what firtly expected here as we have a legacy wo wallet (but we can be sure user never used it) and creatig a new one w/ new naming (but i think it's safe as the legacy wallet had not been used so there is no need for migration)
  - [x] if no legacy wallet it create a new one w/ new naming as intended.
- [ ] w/ macOS(sonoma) client + remote bitcoind (linux)
  - [ ] legacy walet
  - [ ] new wallet name

src/lib.rs
```diff
@@ -239,14 +258,26 @@ fn setup_bitcoind(
...
-    } else {
+    } else if !is_remote {
         #[cfg(windows)]
         if !cfg!(test) && !wo_path.exists() {
             return Err(StartupError::NoWatchonlyInDatadir);
         }
     }
     log::info!("Loading our watchonly wallet on bitcoind.");
-    bitcoind.maybe_load_watchonly_wallet()?;
+    if is_remote {
```
jp1ac4, line 268: Could you add this `if is_remote {}` as an `else {}` to the `if fresh_data_dir {} else if !is_remote {}` above? I think we only need to check for a legacy wallet if it's not a fresh data dir, as otherwise we will have just created a wallet with the new name.

src/lib.rs
```diff
@@ -239,14 +258,26 @@ fn setup_bitcoind(
...
+            if e.is_wallet_not_found() {
+                log::info!("Loading legacy wallet.");
+                bitcoind = BitcoinD::new(bitcoind_config, wo_path_str)?;
+                bitcoind.maybe_load_watchonly_wallet()?;
+            } else {
+                Err(e)?
+            }
+        }
+    } else {
+        bitcoind.maybe_load_watchonly_wallet()?;
```
jp1ac4, line 279: nit: I think you could have a single call to `maybe_load_watchonly_wallet()` after the `if {} else {}` block.

jp1ac4 (commented):
Thanks for the updates. I just had a couple more comments, but it's looking good.

src/lib.rs
```diff
@@ -230,7 +232,24 @@ fn setup_bitcoind(
         .bitcoind_config
         .as_ref()
         .ok_or(StartupError::MissingBitcoindConfig)?;
-    let bitcoind = BitcoinD::new(bitcoind_config, wo_path_str)?;
+
+    let is_remote = bitcoind_config.addr.ip() != IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
```
darosior, line 236: This detection seems brittle. Can't we just query the system to know whether something is running at the configured (ip, port)?

src/lib.rs
darosior: Yes. Even clearer:
- If fresh data dir
- Else if remote
- Else (left untouched)

src/lib.rs
darosior: +1. I think you should `match` on the call to `maybe_load_watchonly_wallet` and only check `is_remote` in case of failure.

src/lib.rs
pythcoiner: does `create_watchonly_wallet()` also load the wallet in bitcoind? if so why we were calling `maybe_load_watchonly_wallet()` prior to this change?

jp1ac4: Yep, it loads the wallet also. I suppose in that case the call to `maybe_load_watchonly_wallet()` could have been in the `else` block only, but calling it on an already-loaded wallet is also fine as it returns immediately.

pythcoiner:
comments adressed w/ few notes:
- i replace IPV4 check against 127.0.0.1 by .is_loopback() that matches both IPV4(127.0.0.1) IPV6(::1)
- remove the error check if `maybe_load_watchonly_wallet()` fail in order to have a 'blind' fallback to the legacy naming in the case we got a false positive at detecting `is_remote`
- i did not `match` on the error handling as we have nothing to do if `maybe_load_watchonly_wallet()` success

jp1ac4 (commented):
Added a couple of comments.

src/lib.rs
```diff
@@ -239,14 +256,26 @@ fn setup_bitcoind(
...
+    } else if is_remote {
+        log::info!("Loading our watchonly wallet on bitcoind.");
+        if let Err(e) = bitcoind.maybe_load_watchonly_wallet() {
+            // here if is_remote and errored we blindly fallback to legacy naming in order
+            // to handle unexpected edge cases like user setup w/ 127.0.0.1
+            // and later change it to its local 192.168.x.x ....
+            if is_remote {
+                log::info!("Loading legacy wallet.");
+                bitcoind = BitcoinD::new(bitcoind_config, wo_path_str)?;
+                bitcoind.maybe_load_watchonly_wallet()?;
```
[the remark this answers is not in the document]
pythcoiner: not definitively should be separate fron the if fresh_datadir statement

pythcoiner:
i'll get back on this tmr

pythcoiner:
Closing for now, i've overlooked too much critical points, will work on a different approach.

jp1ac4:
Is the main issue how to determine if it's remote?

pythcoiner:
> Is the main issue how to determine if it's remote?

i think the approach i've described [here](https://github.com/wizardsardine/liana/pull/1115#issuecomment-2179976109) is better

PR #1116, opened 2024-06-11
Set minimum window size

### LINKED ISSUE #506: Add minimum resize limit.
When I drag the corner of that window to make it smaller, it has no limit so I can make it so small that you don't even see text at all.

Not sure if screenshot makes sense :)

Screenshot 2023-05-12 at 20 32 01

### LINKED ISSUE #507: Scroll bar for receive address
When the window is small(or made small) the full address is not visible and as such introduces a horizontal scroll bar.

It looks kinda weird, and also makes life harder when double-checking addrs.

I guess if https://github.com/wizardsardine/liana/issues/506 gets fixed will solve this problem.

Maybe use the size of the address as a reference for the minimum OX size of the window?
Screenshot 2023-05-12 at 20 35 47

pythcoiner:
fixes #506 
i put ~~800x800~~ as default size, feel free to told me if you think something else more appropriate.
edit: i change to 1000x800 so we can also close #507 (i do not have strong opinion, 1000 looks a bit large to me, but if we decide less than 1000 i think we shouls close #507 also)

### DIFF
--- a/gui/src/main.rs
+++ b/gui/src/main.rs
@@ -4,11 +4,10 @@ use std::{error::Error, io::Write, path::PathBuf, process, str::FromStr};
 
 use iced::{
     event::{self, Event},
-    executor,
-    keyboard::{self},
+    executor, keyboard,
     widget::{focus_next, focus_previous},
     window::settings::PlatformSpecific,
-    Application, Command, Settings, Subscription,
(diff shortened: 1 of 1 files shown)


edouardparis (commented):
Let's go for you choice of size.
ACK 3bca6faffe6f1802e08471af05548f355804767c

PR #1124, opened 2024-06-19
replace the descriptor by template in descriptor registration step

### LINKED ISSUE #727: GUI: display descriptor also in its template form
`wsh(and_or(@0/<0;1>/*, ...`

pythcoiner:
This PR replace the descriptor displayed in the installer descriptor registration step, by the wallet policies template + its keys, as almost (all?) signing devices displays the discriptor in this way for registration.

image

closes #727

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -1,3 +1,4 @@
+use async_hwi::utils::extract_keys_and_template;
 use iced::widget::{
     checkbox, container, pick_list, radio, scrollable, scrollable::Properties, slider, Space,
     TextInput,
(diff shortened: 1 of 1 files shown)


jp1ac4:
I tested it and it's looking nice, except the card containing each key can have a different length. I think you could use `.width(Length::Fill)` so that the cards containing keys all fill the width of the page, the same as the card at the bottom for selecting devices, to give a more uniform look.

pythcoiner:
done:
image

jp1ac4 (approved):
Tested ACK 40f9dd9bc43bac0c966585ac2770a02aa4c92765.

I tested both wsh and tr descriptors. E.g.

image

Issue #853, opened 2023-12-07
RBF follow-ups


darosior:
Here are a couple things which didn't warrant holding up the PR but would be nice to have:
- Expand the explicit-RBF-signaling sanity check to all conflicting transactions (https://github.com/wizardsardine/liana/pull/816#discussion_r1417618042)
- The "feerate too low" error could include the minimum feerate to be more helpful (https://github.com/wizardsardine/liana/pull/816#discussion_r1416268844)

@jp1ac4 feel free to suggest more.

jp1ac4:
We can also check if any directly conflicting transactions are spending unconfirmed change and include these unconfirmed coins as candidates for the replacement transaction (https://github.com/wizardsardine/liana/issues/826#issuecomment-1853058734).

jp1ac4:
We should add ancestor info to coin selection candidates when creating a replacement to ensure we pay extra for any parent transactions that have a lower feerate (https://github.com/wizardsardine/liana/pull/873#discussion_r1518611295).

jp1ac4:
We should support RBF on a recovery transaction. Currently, the RBF candidate coins all have their sequence set to `None`.

nondiremanuel:
I updated the issue description to include Michael's comments. I also divided it in different task and marked the second one as completed as per @aaroncarlucci's PR #1186 which was merged (thanks!)

jp1ac4:
> We should add ancestor info to coin selection candidates when creating a replacement to ensure we pay extra for any parent transactions that have a lower feerate ([#873 (comment)](https://github.com/wizardsardine/liana/pull/873#discussion_r1518611295)).

See https://github.com/wizardsardine/liana/issues/1255.

PR #1125, opened 2024-06-19
descriptors: improve satisfaction size estimate for single key

### LINKED ISSUE #1118: Size estimation in the taproot single key primary path case
Splits from #1006 

This is the low hanging fruit for the size estimation under taproot:

> But the low hanging fruit really is when the primary path is a single key which becomes the Taproot internal key. We could detect this case even without using the planning module.

We agreed with @jp1ac4 that he will have a look at this one and try to see if it's feasible to include it in v6.

jp1ac4:
This is to resolve #1118.

It improves the satisfaction size estimate in the case of a primary path spend involving a single key.

Tests have been updated to account for transaction fees now being lower in some cases due to better estimates.

It was already known that we sometimes end up with a transaction feerate (in sat/vb) lower than our target value, since our fee calculation from coin selection is based on sat/weight, which can give a sat/vb value lower than our target due to rounding. This will probably happen more often now given that we'll pay lower fees due to better size estimates.


darosior:
I think we should avoid hardcoding sizes ourselves as much as we can. Also, we should consider any primary path, not only those with a single key. Here is a diff which achieves this using rust-miniscript's planning module:

<details>

<summary>Click to expand diff</summary>

```diff
diff --git a/src/descriptors/mod.rs b/src/descriptors/mod.rs
index 605b13d..3cefe9f 100644
--- a/src/descriptors/mod.rs
+++ b/src/descriptors/mod.rs
@@ -6,12 +6,13 @@ use miniscript::{
         secp256k1,
     },
     descriptor,
+    plan::{Assets, CanSign},
     psbt::{PsbtInputExt, PsbtOutputExt},
     translate_hash_clone, ForEachKey, TranslatePk, Translator,
 };
 
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     convert::TryInto,
     error, fmt,
     str::{self, FromStr},
@@ -56,6 +57,10 @@ impl From<LianaPolicyError> for LianaDescError {
     }
 }
 
+fn varint_len(n: usize) -> usize {
+    bitcoin::VarInt(n as u64).size()
+}
+
 // Whether the key identified by its fingerprint+derivation path was derived from one of the xpubs
 // for this spending path.
 fn key_is_for_path(
@@ -233,40 +238,59 @@ impl LianaDescriptor {
     /// Callers are expected to account for the Segwit marker (2 WU). This takes into account the
     /// size of the witness stack length varint.
     pub fn max_sat_weight(&self, use_primary_path: bool) -> usize {
-        let max_weight_to_satisfy =
-            if use_primary_path && matches!(self.policy().primary_path, PathInfo::Single(..)) {
-                match self.multi_desc {
-                    descriptor::Descriptor::Wsh(ref wsh) => {
-                        let redeem_script_size = match wsh.as_inner() {
-                            descriptor::WshInner::Ms(ms) => ms.script_size(),
-                            descriptor::WshInner::SortedMulti(..) => {
-                                unreachable!("None of our descriptors is a sorted multi")
-                            }
-                        };
-                        // script size + script + ECDSA signature
-                        // Assumes all ECDSA signatures are 73 bytes, including push opcode and
-                        // sighash suffix.
-                        1 + redeem_script_size + 73
-                    }
-                    descriptor::Descriptor::Tr(..) => {
-                        // Assumes all Schnorr signatures are 66 bytes, including push opcode and
-                        // sighash suffix.
-                        66
+        if use_primary_path {
+            // Get the keys from the primary path, to get a satisfaction size estimation only
+            // considering those.
+            let keys = self
+                .policy()
+                .primary_path
+                .thresh_origins()
+                .1
+                .into_iter()
+                .fold(BTreeSet::new(), |mut keys, (fg, der_paths)| {
+                    for der_path in der_paths {
+                        keys.insert(((fg, der_path), CanSign::default()));
                     }
-                    _ => unreachable!("We only allow wsh and tr descriptors"),
-                }
-            } else {
-                self.multi_desc
-                    .max_weight_to_satisfy()
-                    .expect("Always satisfiable")
+                    keys
+                });
+            let assets = Assets {
+                keys,
+                ..Default::default()
             };
-        // We add one to account for the witness stack size, as the values above give the
-        // difference in size for a satisfied input that was *already* in a transaction
-        // that spent one or more Segwit coins (and thus already have 1 WU accounted for the
-        // emtpy witness). But this method is used to account between a completely "nude"
-        // transaction (and therefore no Segwit marker nor empty witness in inputs) and a
-        // satisfied transaction.
-        max_weight_to_satisfy + 1
+
+            // Unfortunately rust-miniscript satisfaction size estimation is inconsistent. For
+            // Taproot it considers the whole witness (including the control block size + the
+            // script size) but under P2WSH it does not consider the witscript! Therefore we
+            // manually add the size of the witscript, but only under P2WSH by the mean of the
+            // `explicit_script()` helper.
+            let der_desc = self
+                .receive_desc
+                .0
+                .at_derivation_index(0)
+                .expect("unhardened index");
+            let witscript_size = der_desc
+                .explicit_script()
+                .map(|s| varint_len(s.len()) + s.len())
+                .unwrap_or(0);
+
+            // Finally, compute the satisfaction template for the primary path and get its size.
+            der_desc
+                .plan(&assets)
+                .expect("Always satisfiable")
+                .witness_size()
+                + witscript_size
+        } else {
+            // We add one to account for the witness stack size, as the values above give the
+            // difference in size for a satisfied input that was *already* in a transaction
+            // that spent one or more Segwit coins (and thus already have 1 WU accounted for the
+            // emtpy witness). But this method is used to account between a completely "nude"
+            // transaction (and therefore no Segwit marker nor empty witness in inputs) and a
+            // satisfied transaction.
+            self.multi_desc
+                .max_weight_to_satisfy()
+                .expect("Always satisfiable")
+                + 1
+        }
     }
 
     /// Get the maximum size difference of a transaction input spending a Script derived from this
@@ -1112,6 +1136,48 @@ mod tests {
             desc.spender_input_size(false),
             32 + 4 + 1 + 4 + wu_to_vb(witness_size),
         );
+
+        // Now perform the sanity checks under Taproot.
+        let owner_key = PathInfo::Single(descriptor::DescriptorPublicKey::from_str("[abcdef01]xpub6Eze7yAT3Y1wGrnzedCNVYDXUqa9NmHVWck5emBaTbXtURbe1NWZbK9bsz1TiVE7Cz341PMTfYgFw1KdLWdzcM1UMFTcdQfCYhhXZ2HJvTW/<0;1>/*").unwrap());
+        let heir_key = PathInfo::Single(descriptor::DescriptorPublicKey::from_str("[abcdef01]xpub688Hn4wScQAAiYJLPg9yH27hUpfZAUnmJejRQBCiwfP5PEDzjWMNW1wChcninxr5gyavFqbbDjdV1aK5USJz8NDVjUy7FRQaaqqXHh5SbXe/<0;1>/*").unwrap());
+        let timelock = 52560;
+        let desc = LianaDescriptor::new(
+            LianaPolicy::new(
+                owner_key.clone(),
+                [(timelock, heir_key.clone())].iter().cloned().collect(),
+            )
+            .unwrap(),
+        );
+
+        // If using the primary path, it's a keypath spend.
+        assert_eq!(desc.max_sat_vbytes(true), (1 + 65 + 3) / 4);
+        // If using the recovery path, it's a script path spend. The script is 40 bytes long. The
+        // control block is just the internal key and parity, so 33 bytes long.
+        assert_eq!(
+            desc.max_sat_vbytes(false),
+            (1 + 65 + 1 + 40 + 1 + 33 + 3) / 4
+        );
+
+        // The same against the spender_input_size() helper, adding the size of the txin and
+        // checking against a dummy witness stack.
+        fn wit_size(stack: &[Vec<u8>]) -> usize {
+            varint_len(stack.len())
+                + stack
+                    .iter()
+                    .map(|item| varint_len(item.len()) + item.len())
+                    .sum::<usize>()
+        }
+        let txin_boilerplate = 32 + 4 + 1 + 4;
+        let stack = vec![vec![0; 64]];
+        assert_eq!(
+            desc.spender_input_size(true),
+            txin_boilerplate + wu_to_vb(wit_size(&stack)),
+        );
+        let stack = vec![vec![0; 33], vec![0; 40], vec![0; 64]];
+        assert_eq!(
+            desc.spender_input_size(false),
+            txin_boilerplate + wu_to_vb(wit_size(&stack)),
+        );
     }
 
     #[test]
```

</details>

It's not as straightforward as it can get since their planning module is a bit inconsistent, still i think it's much cleaner. I've also expanded the inheritance_sat_size unit test with a Taproot descriptor which sanity checks the behaviour with a primary-path-as-internal-key.

src/spend.rs
```diff
@@ -575,6 +576,8 @@ pub struct CreateSpendRes {
 /// * `main_descriptor`: the multipath Liana descriptor, used to derive the addresses of the
 /// candidate coins.
 /// * `secp`: necessary to derive data from the descriptor.
+/// * `use_primary_path`: whether the primary spending path should be used when estimating
+/// satisfaction size.
 /// * `tx_getter`: an interface to get the wallet transaction for the prevouts of the transaction.
 /// Wouldn't be necessary if we only spent Taproot coins.
```
pythcoiner, line 582:
```suggestion
/// Wouldn't be necessary if we only spent non-taproot coins.
```
Not related to this PR but i notice this few weeks ago

darosior: No, the original sentence is correct. :)

src/commands/mod.rs
```diff
@@ -906,6 +907,7 @@ impl DaemonControl {
             match create_spend(
                 &self.config.main_descriptor,
                 &self.secp,
+                true, // TODO: allow RBF of recovery transaction
```
darosior, line 910: Heh, good point.

src/spend.rs
```diff
@@ -575,6 +576,8 @@ pub struct CreateSpendRes {
 /// * `main_descriptor`: the multipath Liana descriptor, used to derive the addresses of the
 /// candidate coins.
 /// * `secp`: necessary to derive data from the descriptor.
+/// * `use_primary_path`: whether the primary spending path should be used when estimating
+/// satisfaction size.
```
darosior, line 580: Technically we already have this information from the nSequence to use. But maybe it's clearer by using a boolean, i don't feel strongly.

src/spend.rs
pythcoiner: my bad, i misunderstood when you told me about that! :man_facepalming:

tests/test_rpc.py
```diff
@@ -1118,13 +1118,19 @@ def test_rbfpsbt_bump_fee(lianad, bitcoind):
...
+    # feerate to set the min feerate, instead of 1 sat/vb of first
     # transaction:
-    with pytest.raises(RpcError, match=f"Feerate too low: 10."):
-        lianad.rpc.rbfpsbt(first_txid, False, 10)
-    # Using 11 for feerate works for P2WSH. For Taproot we need 12.
-    feerate = 12 if USE_TAPROOT else 11
+    with pytest.raises(RpcError, match=f"Feerate too low: {int(rbf_1_feerate)}."):
+        lianad.rpc.rbfpsbt(first_txid, False, int(rbf_1_feerate))
+    # Using 1 more for feerate works.
+    feerate = int(rbf_1_feerate) + 1
```
darosior, line 1133: Neat.

tests/test_spend.py
```diff
@@ -252,17 +252,17 @@ def test_send_to_self(lianad, bitcoind):
...
-    if not USE_TAPROOT:
-        assert specified_feerate <= spend_feerate <= int(specified_feerate * 115 / 100)
+    spend_feerate = res["fees"]["base"] * COIN / res["vsize"]  # keep as decimal
+    if USE_TAPROOT:
+        assert specified_feerate <= spend_feerate < specified_feerate + 0.5
     else:
-        # FIXME: under Taproot we should not consider the max feerate of all leaves if there
-        # is a spendable internal key.
-        assert specified_feerate <= spend_feerate <= int(specified_feerate * 125 / 100)
+        assert specified_feerate - 0.5 < spend_feerate < specified_feerate + 0.5
```
darosior, line 265: Why is there still this discrepancy between Taproot and P2WSH? That we could now get a spend feerate lower than the specified feerate is also concerning.

tests/test_spend.py
```diff
@@ -364,24 +371,39 @@ def test_coin_selection(lianad, bitcoind):
...
+    feerate = 2
+    spend_res_2 = lianad.rpc.createspend({dest_addr_2: 10_000}, [], feerate)
     assert "psbt" in spend_res_2
-    assert len(spend_res_2["warnings"]) == 0
     spend_psbt_2 = PSBT.from_base64(spend_res_2["psbt"])
+    spend_txid_2 = spend_psbt_2.tx.txid().hex()
+    if USE_TAPROOT:
+        assert len(spend_res_2["warnings"]) == 0
+    else:
+        # We still get a warning in the non-taproot case.
```
darosior, line 394: Just checking my understanding of the rationale for updating this: that's because the size of the child is lower under Taproot, and we now correctly pre-estimate its size, so its feerate is slightly larger and therefore it triggers the CPFP logic?

darosior (commented):
ACK modulo my concern about how we can end up with a smaller feerate than estimated.

src/spend.rs
jp1ac4: OK great, I think it would be good to avoid adding a parameter in that case. I'll try changing it to get the info from nSequence instead.

tests/test_spend.py
jp1ac4: I found the actual P2WSH feerate for the same test can be either above or below the target feerate. Perhaps the higher values occur for low-r ECDSA signatures that are smaller than the max satisfaction size.

In this particular instance, I found the Taproot feerate was always above the target feerate and so kept two separate assertions, but I think in general it can also be smaller than the target feerate.

darosior: How comes? I thought we were always overestimating? Underestimating is more serious and should never happen.

> Perhaps the higher values occur for low-r ECDSA signatures that are smaller than the max satisfaction size.

This could explain the higher value (since we assume signature are always high-r in the estimation) but not the lower one.

jp1ac4: > Underestimating is more serious and should never happen.

It's due to rounding up when converting from weight to vb. For example, if we have a target feerate of 8 sat/vb, the coin selection will use 2 sat/wu as the target. If it finds a transaction with weight, say, 1003 wu, then it will pay a fee of at least 2006 sats. The weight in vb is 251, which gives a feerate of at least 7.99 sat/vb.

Note that if the target feerate is 1 sat/vb, the resulting feerate will not be smaller. For 1 sat/vb, this gives a target of 0.25 sat/wu, and the fee is at least `ceil(0.25 * weight)` sats, but this is also the size of the transaction in vb.

We might be able to modify how coin selection is done to ensure we get the target feerate in sat/vb, e.g. by iteratively increasing the target sat/wu until we reach the target sat/vb.

src/spend.rs
jp1ac4: I've now removed the parameter and instead check the candidates' nSequence values.

src/commands/mod.rs
jp1ac4: The explicit parameter has now been removed, but I've added the TODO here: https://github.com/wizardsardine/liana/issues/853#issuecomment-2181096159.

tests/test_spend.py
jp1ac4: > We might be able to modify how coin selection is done to ensure we get the target feerate in sat/vb, e.g. by iteratively increasing the target sat/wu until we reach the target sat/vb.

It might be possible to do this via the `score()` method of our selection metric so that instead of returning the fee based on the sat/wu target, it returns the fee based on the sat/vb target, which would avoid the need to iterate over sat/wu values.

jp1ac4: But even better if there were an option in bdk_coin_select to target the feerate in sat/vb instead of internally using sat/wu.

darosior: Alright, so this is orthogonal to this PR. Can we somehow also round up when converting from sat/vb to sat/wu to counteract this effect? I'd rather we always very slightly overpay than sometimes create transaction which can't propagate..

jp1ac4: I don't think we can a priori because how the value is rounded up will depend on the weight of the final transaction.

From a quick look, it should be fairly straightforward to add this option to the coin-select crate so that if you specify the feerate in sat/vb, it calculates the fee based on the transaction's vsize instead of weight. In the shorter term, we could add this logic on our side only to the modified metric we're using.

jp1ac4: I see that mempool.space calculates the vsize as weight/4, without rounding up, and uses this for their sat/vb feerate calculation. This would be another option, but I don't think it's technically correct.

src/spend.rs
```diff
@@ -663,6 +664,13 @@ pub fn create_spend(
         value: bitcoin::Amount::MAX,
         script_pubkey: change_addr.addr.script_pubkey(),
     };
+    // If no candidates have relative locktime, then we should use the primary spending path.
+    // Note we set this value before actually selecting the coins, but we expect either all
+    // candidates or none to have relative locktime sequence so this is fine.
+    let use_primary_path = !candidate_coins
+        .iter()
+        .filter_map(|cand| cand.sequence)
+        .any(|seq| seq.is_relative_lock_time());
```
darosior, line 673: Nice. Would have been good to update the documentation of this function to mention the meaning given to the presence of a specified locktime though.

darosior (approved):
ACK 0eda557fdf219b16af8b2d831a755148f1286669

I ran the test suite a few dozens of times in parallel to trigger any potential flakiness in the hardcoded fee values and didn't get any failure.

tests/test_spend.py
darosior: We should track the potential underestimation in a separate issue and try to come up with a fix for v7.

jp1ac4: For reference, https://github.com/wizardsardine/liana/issues/1132 was created.

PR #1131, opened 2024-06-21
Use rustc 1.70 required by gui for docker reproducible build

edouardparis:


darosior:
Looks correct to me, will test.

Issue #1046, opened 2024-03-26
[Remote bitcoind] Remove watchonly wallet when removing local wallet


pythcoiner:
If we remove a wallet that have a remote backend, we do not remove it on bitcoind side, it can lead to issue hit in #1045

darosior:
I think we should rather fix the buggy support for non-local bitcoind nodes instead. (https://github.com/wizardsardine/liana/issues/1059)

jp1ac4:
Even if we fix where the watchonly wallet is created, I think we'll still have this issue if we try to create a new wallet unless the watchonly wallet has a different name.

darosior:
True. So in addition to correctly creating the wo wallet (#1059), we should also delete it if `watchonly_wallet_in_bitcoind_datadir` is set when removing a local datadir.

pythcoiner:
> Even if we fix where the watchonly wallet is created, I think we'll still have this issue if we try to create a new wallet unless the watchonly wallet has a different name.

i think we should have a unique name for each wallet like liana<hash>

pythcoiner:
i'm currently working on #1059, and i don't see how we can remove the watchonly wallet on a remote mode, as bitcoin-core does not have a command (or i miss it) for this and we do not have access to the remote filesystem.

jp1ac4:
Hmm I don't think it's possible. The most we can probably do is unload the wallet via RPC and display a message to the user that they need to delete the bitcoind wallet file before creating a new Liana wallet.

> i think we should have a unique name for each wallet like liana

This would also help if we want to support multiple wallets on a remote node.

pythcoiner:
It looks to me #1115 closes this issue, as new wallet names are in the form `liana-<fingerprint>`, tested today:
- create a legacy named wallet on remote from master (b36220504fd895d69d59325eae6527708736b272)
- remove this wallet client-side
- try to create a new one (still from master): it fail like #1045
- try to create a new one from #1115 (549b8f2b7c567861b2817af9f82a11b50fc35b25) works fine, create a new wo wallet w/ new naming

pythcoiner:
now #1141 is merged i think we can close this => 
- we technicaly cannot remove the remote wo wallet from liana
- we avoid user to create new ones

I close it for now, anyone feel free to reopen if we should do something else.

PR #1137, opened 2024-06-25
installer: add a warning modal if remote address is not loopback

### LINKED ISSUE #1059: [Remote bitcoind] wallet name is create in /home/<user>/.liana

pythcoiner:
This PR add a warning modal if user try to connect to a non-loopback address & use RPCAuth:

image

It also add a check on address + RPC user/password: 
`Check connection` button is greyed out until address is valid + user and password are filled (if RPCAuth selected)

Closes #1059 

 - [x] better check of valid address, atm `::1:<port>` fail but should not (IPV6 loopback)

### DIFF
--- a/gui/src/installer/message.rs
+++ b/gui/src/installer/message.rs
@@ -45,6 +45,8 @@ pub enum DefineBitcoind {
     RpcAuthTypeSelected(RpcAuthType),
     PingBitcoindResult(Result<(), Error>),
     PingBitcoind,
+    DisplayModal(bool),
+    PingAndCloseModal,
 }
 
 #[derive(Debug, Clone)]
--- a/gui/src/installer/step/bitcoind.rs
+++ b/gui/src/installer/step/bitcoind.rs
@@ -1,6 +1,6 @@
 #[cfg(target_os = "windows")]
 use std::io::{self, Cursor};
-use std::net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, TcpListener};
 use std::path::PathBuf;
 use std::str::FromStr;
 
@@ -336,6 +336,7 @@ pub struct DefineBitcoind {
     selected_auth_type: RpcAuthType,
     address: form::Value<String>,
     is_running: Option<Result<(), Error>>,
+    modal_displayed: bool,
 
     // Internal cache to detect network change.
     network: Option<Network>,
@@ -348,6 +349,7 @@ impl DefineBitcoind {
             selected_auth_type: RpcAuthType::CookieFile,
             address: form::Value::default(),
             is_running: None,
+            modal_displayed: false,
             network: None,
         }
     }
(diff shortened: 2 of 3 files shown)


nondiremanuel:
Does this address also #1046 or will it need another PR?

nondiremanuel:
Thanks for taking this PR! A couple of questions/comments more on the "UI" side:

- Why is the font green? It doesn't give me the idea of a warning at first glance (or is it maybe a standard that we already use for warnings? Do we have one?)
- Wording: "if you are confident in your actions" doesn't say much in my opinion. If the remote bitcoind is not supported as we say in the first sentence, how could confidence fix it? Either we support it but requires expertise or not support it. Am I missing something?
In general, the second sentence doesn't seem very useful to me. If the purpose of the warning is to warn the user about remote bitcoind while not blocking them, I think the first sentence already accomplishes it. We could just change the "Continue" button to "Proceed anyway" or something like this. But maybe I'm missing something more that we want to communicate to the user.

pythcoiner:
> * Why is the font green? It doesn't give me the idea of a warning at first glance (or is it maybe a standard that we already use for warnings? Do we have one?)
because i'm not good at design stuff :grin:. Which color should we use? (i think red is more for error)

> * Wording: "if you are confident in your actions" doesn't say much in my opinion. If the remote bitcoind is not supported as we say in the first sentence, how could confidence fix it? Either we support it but requires expertise or not support it. Am I missing something?
>   In general, the second sentence doesn't seem very useful to me. If the purpose of the warning is to warn the user about remote bitcoind while not blocking them, I think the first sentence already accomplishes it. We could just change the "Continue" button to "Proceed anyway" or something like this. But maybe I'm missing something more that we want to communicate to the user.

I'm also not good at wording :grin:! i ask gpt to reword me, but it seems he's not better than me...

The mean is there is some special case where the user have to connect it's local machine w/ a local IP (nixOS?, VM, Docker, etc...), so it looks like an external IP, but only the (advanced) user know it, we cannot determine it. I dont think we should put much more details here, but maybe we should document it (or not).

Feel free to advise me wording, i'll be happy to replace mine!

pythcoiner:
> Does this address also #1046 or will it need another PR?

I think we can also close it if this PR merged, as technicaly we cannot remove the wo wallet on the remote (we do not have access to the remote filesystem and bitcoind does not have a command for this) and the only thing we can do is avaoid the user to create it (it's what we do in this PR) (and maybe document it)

nondiremanuel:
> >   because i'm not good at design stuff 😁. Which color should we use? (i think red is more for error)

If we have other warning modals in the application (if so, I don't remember how they are formatted), I would align it to those. Usually a warning symbol close to the "Warning" title of the modal, leaving the font white could be enough. Alternatively, I would use yellow.

> I'm also not good at wording 😁! i ask gpt to reword me, but it seems he's not better than me...
> 
> The mean is there is some special case where the user have to connect it's local machine w/ a local IP (nixOS?, VM, Docker, etc...), so it looks like an external IP, but only the (advanced) user know it, we cannot determine it. I dont think we should put much more details here, but maybe we should document it (or not).
> 
> Feel free to advise me wording, i'll be happy to replace mine!

In order to understand which could be the warning, what happens if the user inserts a remote node IP and still decides to go on ignoring the warning? Do we throw an error or something?

darosior:
For the warning i was thinking of a simple line (in yellow for instance, but i'm not sure about the color) displayed next to the text input for the ip.

pythcoiner:
> In order to understand which could be the warning, what happens if the user inserts a remote node IP and still decides to go on ignoring the warning? Do we throw an error or something?

4 issues are described at start of this [comment](https://github.com/wizardsardine/liana/issues/1059#issuecomment-2179757148)

pythcoiner:
> For the warning i was thinking of a simple line (in yellow for instance, but i'm not sure about the color) displayed next to the text input for the ip.
> […](#)

My 2 points for the modal are:
- they cannot miss the warning
- if they choose cancel we automatically reload default ip:port values in the address field

pythcoiner:
there is no yellow color defined in liana [theme](https://github.com/wizardsardine/liana/blob/master/gui/ui/src/color.rs), anyone to pick one? (i'm color blind, so not easy for me :grin:)

nondiremanuel:
I would personally go for the yellow sentence (as in the #1141) since it is less intrusive for the user and at the same time accomplishes to communicate the message. Regarding the wording, I would use something like:

"Connection to a remote bitcoin node is not supported. Insert an IP address bound to the same machine running Liana (ignore this warning if that's already the case)"

or some variant of it (I'm not sure about the verb "bind" for an IP address. "Belonging to"? "Used by"? Else?)

nondiremanuel:
> there is no yellow color defined in liana [theme](https://github.com/wizardsardine/liana/blob/master/gui/ui/src/color.rs), anyone to pick one? (i'm color blind, so not easy for me 😁)

The one in the other PR seems good to me. Maybe I would make the font a little bigger

nondiremanuel:
The alternative solution (#1141) was merged, so I close this PR

PR #1139, opened 2024-06-25
add specter support for tap-miniscript

pythcoiner:
This pr enable specter support for tap-miniscript.
tested:

descriptor registration
spend trough internal key path
spent trough taptree path


darosior:
Note we are in feature freeze, so this will have to wait for v7.

PR #1141, opened 2024-06-26
installer: add a warning if remote adress is not loopback

pythcoiner:
This PR is a variant of #1137, it add a warning under text input field if user try to connect to a non-loopback address & use RPCAuth:

image

It also add a check on address + RPC user/password/cookie:
Check connection button is greyed out until address is valid + user and password are filled (if RPCAuth selected)

Closes https://github.com/wizardsardine/liana/issues/1059

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -1,12 +1,13 @@
 use async_hwi::utils::extract_keys_and_template;
 use iced::widget::{
-    checkbox, container, pick_list, radio, scrollable, scrollable::Properties, slider, Space,
-    TextInput,
+    checkbox, container, pick_list, radio, scrollable, scrollable::Properties, slider, Button,
+    Space, TextInput,
 };
 use iced::{alignment, widget::progress_bar, Alignment, Length};
 
 use async_hwi::DeviceKind;
 use liana_ui::component::text;
+use std::net::{Ipv4Addr, Ipv6Addr};
 use std::path::PathBuf;
 use std::{collections::HashSet, str::FromStr};
 
@@ -839,6 +840,17 @@ pub fn define_bitcoin<'a>(
     selected_auth_type: &RpcAuthType,
     is_running: Option<&Result<(), Error>>,
 ) -> Element<'a, Message> {
+    let is_loopback = if let Some((ip, _port)) = address.value.clone().rsplit_once(':') {
+        let (ipv4, ipv6) = (Ipv4Addr::from_str(ip), Ipv6Addr::from_str(ip));
+        match (ipv4, ipv6) {
+            (_, Ok(ip)) => ip.is_loopback(),
+            (Ok(ip), _) => ip.is_loopback(),
+            _ => false,
+        }
+    } else {
+        false
+    };
+
     let col_address = Column::new()
         .push(text("Address:").bold())
         .push(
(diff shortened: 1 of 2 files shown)


gui/src/installer/view.rs
```diff
@@ -839,6 +840,21 @@ pub fn define_bitcoin<'a>(
...
+            match (ipv4, ipv6) {
+                (_, Ok(ip)) => ip.is_loopback(),
+                (Ok(ip), _) => ip.is_loopback(),
+                _ => false,
+            }
+        } else {
+            false
+        }
+    } else {
+        true
```
jp1ac4, line 855: Is it worth applying the IP check also here in case a remote cookie path has been entered somehow?

gui/src/installer/view.rs
```diff
@@ -951,9 +983,13 @@ pub fn define_bitcoin<'a>(
                     .spacing(10)
                     .push(Container::new(
                         button::secondary(None, "Check connection")
-                            .on_press(Message::DefineBitcoind(
-                                message::DefineBitcoind::PingBitcoind,
-                            ))
+                            .on_press_maybe(if check_connect_enable {
```
jp1ac4, line 986: I like this `on_press_maybe`! We'll be able to use that in many places :)

jp1ac4 (commented):
Added a few comments. I didn't comment on the wording of the warning as I understand this is already being discussed separately.

gui/src/installer/view.rs
darosior: +1, we might as well show the warning in both cases.

gui/src/installer/view.rs
```diff
@@ -917,6 +942,13 @@ pub fn define_bitcoin<'a>(
         })
         .spacing(10);
 
+    let check_connect_enable = if let RpcAuthType::UserPass = selected_auth_type {
+        address.valid
+            && !rpc_auth_vals.password.value.is_empty()
+            && !rpc_auth_vals.user.value.is_empty()
+    } else {
+        address.valid
```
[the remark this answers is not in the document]
pythcoiner: done

gui/src/installer/view.rs
pythcoiner: done

pythcoiner:
i've update the sentence to what @nondiremanuel offer in the other PR, about the size i've kept the same as the error displayed if the ip is not correct for consistency, but can increase if you guys think is better.

gui/src/installer/view.rs
pythcoiner: i just noticed i have never been use in Liana! :smile:

jp1ac4 (commented):
Thanks for the updates. Sorry, I missed a couple of small things in my previous review, but otherwise I've tested it and it's looking good.

gui/src/installer/view.rs
```diff
@@ -852,6 +864,15 @@ pub fn define_bitcoin<'a>(
             .size(text::P1_SIZE)
             .padding(10),
         )
+        .push_maybe(
+            if !is_loopback  && address.valid{
+                Some(iced::widget::Text::new("Connection to a remote bitcoin node is not supported. Insert an IP address bound to the same machine running Liana (ignore this warning if that's already the case)")
```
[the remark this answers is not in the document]
jp1ac4: Note, an example of `cargo fmt` not working with the long string is that there's an extra space in `!is_loopback  && address.valid`.

nondiremanuel:
> I didn't comment on the wording of the warning as I understand this is already being discussed separately.

Feel free to if you see something could be improved (especially since we are not native English speakers)!

gui/src/installer/view.rs
jp1ac4: Apparently it was only added in iced 0.10.0 :monocle_face:

jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/pull/1141/commits/3d55616b66e8a0995e2ef9e4ad6d294bf23fe6a7.

PR #1144, opened 2024-06-27
Do not let user try recovery if one block left

### LINKED ISSUE #1102: Grey out `next` button if no recovery path availables
image
If no recovery path available we still can select coins and click `next`, then an error is triggered , i think we should at least grey-out the next button.

nit: we should named the error, its not unknown as we have the right description.

pythcoiner:
Closes #1102 and resolve @kloaec issue where 1 block recovery path was displayed recoverable even if tx not confirmed.

### DIFF
--- a/gui/src/app/state/recovery.rs
+++ b/gui/src/app/state/recovery.rs
@@ -247,7 +247,7 @@ fn recovery_paths(wallet: &Wallet, coins: &[Coin], blockheight: i32) -> Vec<Reco
                 .iter()
                 .filter(|coin| {
                     coin.spend_info.is_none()
-                        && remaining_sequence(coin, blockheight as u32, sequence) <= 1
+                        && remaining_sequence(coin, blockheight as u32, sequence, true) <= 1
                 })
                 .fold(
                     (0, Amount::from_sat(0)),
--- a/gui/src/app/state/mod.rs
+++ b/gui/src/app/state/mod.rs
@@ -158,8 +158,12 @@ impl State for Home {
                             if coin.block_height.is_some() {
                                 self.balance += coin.amount;
                                 let timelock = self.wallet.main_descriptor.first_timelock_value();
-                                let seq =
-                                    remaining_sequence(&coin, cache.blockheight as u32, timelock);
+                                let seq = remaining_sequence(
+                                    &coin,
+                                    cache.blockheight as u32,
+                                    timelock,
+                                    false,
+                                );
                                 // Warn user for coins that are expiring in less than 10 percent of
                                 // the timelock.
                                 if seq <= timelock as u32 * 10 / 100 {
--- a/gui/src/app/state/spend/step.rs
+++ b/gui/src/app/state/spend/step.rs
@@ -147,8 +147,8 @@ impl DefineSpend {
         self.coins.sort_by(|(a, a_selected), (b, b_selected)| {
             if *a_selected && !b_selected || !a_selected && *b_selected {
                 b_selected.cmp(a_selected)
-            } else if remaining_sequence(a, blockheight, timelock)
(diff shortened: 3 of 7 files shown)


nondiremanuel:
For the moment I assigned it to v7 since as we were discussing yesterday it is not strictly required to have it immediately (it is a behavior that should not be "regular" Liana wallet usage)

pythcoiner:
updated wording as offered by @jp1ac4 on discord:
image

nondiremanuel:
Some minor comments:
- Can we make "path" singular when there is only one recovery path available?
- For the same reason, when there is 1 recovery path available, "select one" is not very precise. Maybe we should use a more general "please select"
- Is "will be" correct in this case? Why not "x recovery path(s) is/are available?"? 

Also, was the green bar below the sentence already present? I didn't get it last time I tried iirc.

pythcoiner:
image

image

nondiremanuel:
Great thanks! The only nit that I notice remaining (not part of this change but I just noticed) is the fact that "can recover X coins" should be singular (_coin_) when the coin is 1. We could maybe use this PR to correct also this nit.

pythcoiner:
image

jp1ac4:
Checking again the original issue, I'm wondering if the problem was that the coin that had a recovery path available at the next block was still unconfirmed (see https://github.com/wizardsardine/liana/issues/1102#issuecomment-2255475005). The `createrecovery` command only considers confirmed coins.

pythcoiner:
> Checking again the original issue, I'm wondering if the problem was that the coin that had a recovery path available at the next block was still unconfirmed (see [#1102 (comment)](https://github.com/wizardsardine/liana/issues/1102#issuecomment-2255475005)). The `createrecovery` command only considers confirmed coins.

You right i guess we were both (Kevin and I) testing w/ a 1 block recovery timelock, i'll look for a way to handle this edge case.

pythcoiner:
i've added a new `recovery` arg to `remaining_sequence()` instead in order to handle the edge case

darosior:
A 1-block relative locktimed transactions cannot be included in the same block as its parent. Only transactions valid in the next block can be broadcast on the network. That's the reason why `createrecovery` only considers confirmed coins.

nondiremanuel:

gui/src/app/state/recovery.rs
```diff
@@ -247,7 +247,7 @@ fn recovery_paths(wallet: &Wallet, coins: &[Coin], blockheight: i32) -> Vec<Reco
                 .iter()
                 .filter(|coin| {
                     coin.spend_info.is_none()
-                        && remaining_sequence(coin, blockheight as u32, sequence) <= 1
+                        && remaining_sequence(coin, blockheight as u32, sequence, true) <= 1
```
jp1ac4, line 250: I think it might be enough to include an additional filter here that `coin.block_height.is_some()` and then the number of recoverable coins will match those found by the `createrecovery` command. Would this avoid the need to modify the `remaining_sequence` function?

pythcoiner: you mean as a replacement for the whole PR?

jp1ac4: Possibly. There seems to be an inconsistency with `createrecovery` in terms of which coins are considered, which can lead to an error because the GUI thinks there are coins to recover and then the `createrecovery` command ignores them.

I think we should first align the GUI with lianad and then we can see if this error with one block still occurs.

nondiremanuel:
@jp1ac4 can we close this one after we merged #1643 or is it still needed?

jp1ac4:
Thanks for the reminder.

As far as I can tell, the issue addressed by this PR has been resolved by #1144 and therefore this PR is no longer required.

PR #1145, opened 2024-06-27
Add -h/--help args to the GUI

pythcoiner:
```shell
$ liana-gui -h

Usage: liana-gui [OPTIONS]

Options:
    --conf <PATH>       Path of configuration file (gui.toml)
    --datadir <PATH>    Path of liana datadir
    -v, --version       Display liana-gui version
    -h, --help          Print help
    --bitcoin           Use bitcoin network (default)
    --testnet           Use testnet network
    --signet            Use signet network
    --regtest           Use regtest network

```


jp1ac4:
I wonder if there might be a simple way to automate this, but perhaps it wouldn't be worth doing as the options don't change often.

pythcoiner:
> I wonder if there might be a simple way to automate this, but perhaps it wouldn't be worth doing as the options don't change often.

'clap' crate is a good solution , but it's adding one more dependency, i think it not worth it

jp1ac4 (approved):
Tested ACK 243f4eac0aaf3def57b3a1107ff8db67749609e7.

PR #1149, opened 2024-06-28
Only Coldcard MK4 supported as of now

pythcoiner:


darosior:
@scgbckbone told me the Edge firmware for the Q should be available by next week.

PR #1150, opened 2024-06-28
gui: bump version to v6

darosior:
In preparation of the rc1 release.

### DIFF
--- a/gui/src/daemon/model.rs
+++ b/gui/src/daemon/model.rs
@@ -64,7 +64,7 @@ impl SpendTx {
         secp: &secp256k1::Secp256k1<impl secp256k1::Verification>,
         network: Network,
     ) -> Self {
-        let max_vbytes = desc.unsigned_tx_max_vbytes(&psbt.unsigned_tx);
+        let max_vbytes = desc.unsigned_tx_max_vbytes(&psbt.unsigned_tx, false);
         let change_indexes: Vec<usize> = desc
             .change_indexes(&psbt, secp)
             .into_iter()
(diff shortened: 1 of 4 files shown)


gui/src/daemon/model.rs
```diff
@@ -64,7 +64,7 @@ impl SpendTx {
         secp: &secp256k1::Secp256k1<impl secp256k1::Verification>,
         network: Network,
     ) -> Self {
-        let max_vbytes = desc.unsigned_tx_max_vbytes(&psbt.unsigned_tx);
+        let max_vbytes = desc.unsigned_tx_max_vbytes(&psbt.unsigned_tx, false);
```
darosior, line 67: @edouardparis is `SpendTx::new` used for both the recovery and the primary-path spends? If only for primary path i could use `true` here instead to avoid overestimating.

edouardparis: Yes used by both: https://github.com/wizardsardine/liana/blob/master/gui/src/app/state/recovery.rs#L173
but I can create a SpendTx::new_recovery or add an arg to the function

darosior:
self-ACK 973938fb20cfba75536623f21d4277c5407bb3cc

gui/src/daemon/model.rs
darosior: If you think it's worth doing, let's open an issue to track this.

jp1ac4: Could we check for any transaction inputs with the nSequence being a recovery locktime to determine whether the primary path should be used? (I borrowed from https://github.com/wizardsardine/liana/pull/1125#discussion_r1647565509 :slightly_smiling_face:)

PR #1165, opened 2024-07-01
Add interrupt method to state trait

### LINKED ISSUE #1162: Cannot register descriptor from `settings` w/ Jade
Jade detected here

image

but not here

image

then if i unplug and re-plug it, it's detected:

image

edouardparis:
Before changing panel, the previous state may need to release a hardware connection for example.
In order to achieve this we introduce an interrupt method that is used to drop action state.

close #1162


pythcoiner:
tACK 5c59c4f5deb5576b7935191d4ed11a8916c974de

PR #1166, opened 2024-07-01
v6: Add interrupt method to state trait

edouardparis:
Before changing panel, the previous state
may need to release a harware connection
for example.
In order to achieve this we introduce an interrupt method that is used to drop action state.

backport of #1165


edouardparis (commented):
ACK c4e281438c2f4d0770df643a43ba5481356b7a4e

PR #1171, opened 2024-07-04
gui: fix displayed feerate estimate

### LINKED ISSUE #1168: [GUI] Displayed feerate is underestimated for primary path spends
This issue arises from the improvement in size estimation from #1125. It only affects the value displayed and not how the underlying transaction is created.

The displayed feerate for new spends & PSBTs is underestimated:

image

Note that when viewing the transaction itself, the exact feerate is shown:

image

See https://github.com/wizardsardine/liana/pull/1150#discussion_r1658663088 and subsequent discussion for possible fixes.

jp1ac4:
This is to fix #1168.

When estimating transaction size, check if any inputs are using a relative locktime.

I think this logic could later be moved to the `unsigned_tx_max_vbytes` method itself.

### DIFF
--- a/gui/src/daemon/model.rs
+++ b/gui/src/daemon/model.rs
@@ -64,7 +64,14 @@ impl SpendTx {
         secp: &secp256k1::Secp256k1<impl secp256k1::Verification>,
         network: Network,
     ) -> Self {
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
ACK 6c5f83ce15805997632cc6f698d40b4c2eb4017e

PR #1178, opened 2024-07-08
gui: update Liana to latest

darosior:
To get the fixes since rc1 and drop the rc1 suffix in the version.


darosior:
self-ACK 108290373655ced2e9ff56e2c9e8fdce22a81678

Issue #56, opened 2022-10-07
Alternative Bitcoin backend(s)


darosior:
The only Bitcoin backend the first release will have, and that we'll support in the near future, is Bitcoin Core using a watchonly wallet there. Other software tend to (ab)use the Electrum protocol and other kinds of backends (explorer APIs for instance). We might want to support other Bitcoin backends than Bitcoin Core.

For instance i can think of those as being desirable:
- Electrum (but we need to introduce Tor, and/or TLS... Meh)
- Maybe a Bitcoin Core backend that does not require a watchonly wallet (`scantxoutset`?)

Thoughts? Is there interest from users for prioritising this feature?

darosior:
Looks like https://github.com/bitcoindevkit/rust-electrum-client could be a reasonable choice for an Electrum backend. Depends on their supported MSRV of course.

Or maybe even BDK? It looks like a lot of their features have been made optional, so we could just use it as the Electrum backend without pulling more dependencies than their own code. That sounds reasonable to both not reinvent the wheel and concentrate efforts there should we need improvements or have bug fixes. Provided there is a decent MSRV too.
See https://github.com/bitcoindevkit/bdk.

darosior:
Another alternative. Sexier but that requires more work, is supporting BIP158 using Nakamoto. https://github.com/cloudhead/nakamoto

darosior:
Looks like BDK's [MSRV is 1.46](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md#contribution-workflow). Which works for us. But i think they are going for 1.48 now, which is still fine.

vincenzopalazzo:
>Another alternative. Sexier but that requires more work, is supporting BIP158 using Nakamoto. https://github.com/cloudhead/nakamoto

I like to have this alternative, and I will put some more love on it in the near future, so if you think to reconsidered it we may want to talk some day

darosior:
Awesome! Feel free to share design decisions or a WIP branch here if you'd like to have feedback before posting the PR.

-------- Original Message --------

vincenzopalazzo:
I am in touch with @cloudhead, so I plan to work upstream all my changes, for now, are all inside different PR https://github.com/cloudhead/nakamoto/pulls and I arrived at a proof of concept with a core lightning plugin.

Would it be cool to have a list of things that you need, one of my problems currently is estimating the fee like bitcoind, but I need some free time to think about it harder

darosior:
[...]

PR #1180, opened 2024-07-09
Save transactions in database

jp1ac4:
This is the first step of https://github.com/wizardsardine/liana/issues/56#issuecomment-2183063784.

The poller will now save transactions in our own database. These transactions are selected based on the deposit and spend transactions of coins. Only the txid and transaction itself are saved, with the corresponding block height and time taken from the coins table.

In a couple of follow-up commits, I've replaced some RPC calls to bitcoind with DB queries.

### DIFF
--- a/src/database/sqlite/mod.rs
+++ b/src/database/sqlite/mod.rs
@@ -2384,7 +2384,7 @@ CREATE TABLE labels (
     }
 
     #[test]
-    fn v0_to_v2_migration() {
+    fn v0_to_v5_migration() {
         let secp = secp256k1::Secp256k1::verification_only();
 
         // Create a database with version 0, using the old schema.
@@ -2486,6 +2486,12 @@ CREATE TABLE labels (
         maybe_apply_migration(&db_path, &[]).unwrap();
         let db = SqliteDb::new(db_path, None, &secp).unwrap();
 
+        // The DB version has been updated.
+        {
+            let mut conn = db.connection().unwrap();
+            let version = conn.db_version();
+            assert_eq!(version, 5);
+        }
         // We should now be able to insert another PSBT, to query both, and the first PSBT must
         // have no associated timestamp.
         {
@@ -2528,6 +2534,22 @@ CREATE TABLE labels (
             let coins = conn.coins(&[], &[]);
             assert_eq!(coins.len(), 3);
             assert_eq!(coins.iter().filter(|c| !c.is_immature).count(), 2);
+        }
+
+        // We can insert labels.
+        {
+            let mut conn = db.connection().unwrap();
+
+            let txid_str = "0c62a990d20d54429e70859292e82374ba6b1b951a3ab60f26bb65fee5724ff7";
+            let txid = LabelItem::from_str(txid_str, bitcoin::Network::Bitcoin).unwrap();
+            let mut txids_labels = HashMap::new();
+            txids_labels.insert(txid.clone(), Some("hello".to_string()));
+            conn.update_labels(&txids_labels);
+
+            let mut items = HashSet::new();
+            items.insert(txid);
+            let db_labels = conn.db_labels(&items);
+            assert_eq!(db_labels[0].value, "hello");
         }
 
         fs::remove_dir_all(tmp_dir).unwrap();
@@ -2857,44 +2879,6 @@ CREATE TABLE labels (
             let bitcoin_txs: HashSet<_> = bitcoin_txs.into_iter().collect();
             assert_eq!(bitcoin_txs.len(), bitcoin_txs_in_db.len());
             assert_eq!(bitcoin_txs, bitcoin_txs_in_db);
-        }
-
-        fs::remove_dir_all(tmp_dir).unwrap();
-    }
-
-    #[test]
-    fn v0_to_v5_migration() {
-        let secp = secp256k1::Secp256k1::verification_only();
-
-        // Create a database with version 0, using the old schema.
-        let tmp_dir = tmp_dir();
-        fs::create_dir_all(&tmp_dir).unwrap();
-        let db_path: path::PathBuf = [tmp_dir.as_path(), path::Path::new("lianad_v0.sqlite3")]
-            .iter()
-            .collect();
-        let mut options = dummy_options();
-        options.schema = V0_SCHEMA;
-        options.version = 0;
-        create_fresh_db(&db_path, options, &secp).unwrap();
-
-        let db = SqliteDb::new(db_path, None, &secp).unwrap();
-        db.maybe_apply_migrations(&[]).unwrap();
-
-        {
-            let mut conn = db.connection().unwrap();
-            let version = conn.db_version();
-            assert_eq!(version, 5);
-
-            let txid_str = "0c62a990d20d54429e70859292e82374ba6b1b951a3ab60f26bb65fee5724ff7";
-            let txid = LabelItem::from_str(txid_str, bitcoin::Network::Bitcoin).unwrap();
-            let mut txids_labels = HashMap::new();
-            txids_labels.insert(txid.clone(), Some("hello".to_string()));
-            conn.update_labels(&txids_labels);
-
-            let mut items = HashSet::new();
-            items.insert(txid);
-            let db_labels = conn.db_labels(&items);
-            assert_eq!(db_labels[0].value, "hello");
         }
 
         fs::remove_dir_all(tmp_dir).unwrap();


src/commands/mod.rs
```diff
@@ -991,38 +991,22 @@ impl DaemonControl {
...
-                        height: block.map(|b| b.height),
-                        time: block.map(|b| b.time),
-                    })
-            })
+        let transactions = db_conn
+            .list_wallet_transactions(&txids)
+            .into_iter()
+            .map(|(tx, height, time)| TransactionInfo { tx, height, time })
             .collect();
         ListTransactionsResult { transactions }
```
pythcoiner, line 999: i think we can call directly `list_transaction()` here?
```rust
        let txids = self.db.connection().list_txids(start, end, limit);
        self.list_transactions(&txids)
```

jp1ac4: Ah yes, you're right. I'll change that.

src/database/sqlite/mod.rs
```diff
@@ -682,6 +682,101 @@ impl SqliteConn {
...
+        // The UNION will remove duplicates.
+        // We assume that a transaction's block info is the same in every coins row
+        // it appears in.
+        let query = format!(
+            "SELECT t.tx, c.blockheight, c.blocktime \
+            FROM transactions t \
+            INNER JOIN ( \
+                SELECT txid, blockheight, blocktime \
+                FROM coins \
+                WHERE wallet_id = {WALLET_ID} \
```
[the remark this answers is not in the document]
jp1ac4: This is due to the `WHERE txid IN (..)` clause in the query. It seems not straightforward to insert a list using the ? approach and so the query string is constructed using `format!()` instead, which is also the approach used by the `db_labels` method.

src/testutils.rs
```diff
@@ -437,6 +439,60 @@ impl DatabaseConnection for DummyDatabase {
...
+        let txs = self.db.read().unwrap().txs.clone();
+        let coins = self.coins(&[], &[]);
+        // Get deposit and spend txids of all coins and keep only those not in txs.
+        coins
+            .keys()
+            .map(|op| op.txid)
+            .chain(coins.values().filter_map(|c: &Coin| c.spend_txid))
+            .filter(|txid| !txs.contains_key(txid))
+            .collect()
+    }
```
jp1ac4, line 453: Note this may currently return duplicate txids. Perhaps I should rather leave this as `todo!()` as the code is not currently used anywhere.

edouardparis:
I think list_missing_txids may fetch a too big batch of txids. I was thinking maybe it is better to loop over list_txids: https://github.com/wizardsardine/liana/blob/master/src/database/mod.rs#L148 from current time to genesis block timestamp, and for unconfirmed transactions use the coins(&[Status::Unconfirmed], ...)

jp1ac4:
Perhaps yes, although in that case, we should probably also do the same for the `coins()` method, as the max number of missing txids should be of a similar order to the total number of coins in the DB.

src/testutils.rs
jp1ac4: This code is indeed used by the tests so instead I changed it to remove duplicate txids.

jp1ac4:
I used a loop to get the missing txids using `ROW_NUMBER()` as it seemed the simplest way. Most of the time, we don't expect there to be many missing values, only when the transactions table is initially populated for an existing wallet or perhaps if the poller has not been run for a long time.

edouardparis:
Sorry If I was unclear, my point was to fetch missing txids and store transactions by batch in order to not keep a whole set in memory.

jp1ac4:
Ah I see, yep that makes sense, will change it.

jp1ac4:
> In fact, i am thinking that we do not need to paginate txids to do that, just fetch 20 txids from the coins table where the txid is not in the txid field of the table transactions, fetch their raw tx with bitcoind then store them in transactions table and repeat until no txids are fund.

I couldn't quite do it like this in the end as there's a chance we won't be able to retrieve the transaction from bitcoind, and so the loop might have continued forever. Instead, I get all missing txids in one query still, but then retrieve chunks of transactions at a time so that we don't store them all in memory at once. As the transactions are larger than txids, and we already store the same order of coins in memory as there are txids, I felt it was fine to do it this way.

src/commands/mod.rs
```diff
@@ -150,27 +150,27 @@ impl fmt::Display for RbfErrorInfo {
     }
 }
 
-/// A wallet transaction getter which fetches the transaction from our Bitcoin backend with a cache
+/// A wallet transaction getter which fetches the transaction from our database backend with a cache
```
jp1ac4, line 153: Perhaps I could keep the Bitcoin backend as a fallback in case we can't find the tx in the DB.

src/commands/mod.rs
```diff
@@ -990,39 +990,19 @@ impl DaemonControl {
...
-                        time: block.map(|b| b.time),
-                    })
-            })
-            .collect();
-        ListTransactionsResult { transactions }
+        self.list_transactions(&txids)
     }
 
     /// list_transactions retrieves the transactions with the given txids.
     pub fn list_transactions(&self, txids: &[bitcoin::Txid]) -> ListTransactionsResult {
```
jp1ac4, line 999: Perhaps for any txs not found in the DB, I could use the Bitcoin backend as a fallback.

src/bitcoin/poller/looper.rs
```diff
@@ -156,6 +156,23 @@ fn update_coins(
     }
 }
 
+// Add missing transactions to the database.
+fn add_missing_txs(bit: &impl BitcoinInterface, db_conn: &mut Box<dyn DatabaseConnection>) {
+    let txids = db_conn.list_missing_txids();
+    log::debug!("Missing txids: {:?}", txids);
+    // Take chunks to avoid storing too many transactions in memory.
+    for chunk in txids.chunks(100) {
```
darosior, line 164: What you're iterating over is already stored in memory, no matter the chunks you take.

darosior (commented):
I think we should do something closer to what we do for the backend and its poller:
- Have the coins table's txid row reference a transaction in the transactions table (thereby introducing a constraint that a transaction must exist in database for any coin);
- When inserting new coins to the database, first insert the corresponding transaction.

Of course this means you need access to transactions at the time of the migration. I'm still thinking about ways to reduce ugliness of doing that.

src/bitcoin/poller/looper.rs
jp1ac4: All the txids are in memory, but using chunks here is in order not to store all txs themselves in memory.

darosior: Oh my bad i read too quickly when writing the main review comment.

jp1ac4:
> Of course this means you need access to transactions at the time of the migration. I'm still thinking about ways to reduce ugliness of doing that.

Perhaps we could add a migration step in `Poller::new()` around the same place we call `looper::maybe_initialize_tip(&bit, &db)`.

darosior:
Alright, so here is how i think we should proceed:
1. Separate migration from the SQLite database constructor
2. When setting up SQLite at init, first poll the database version before proceeding with any migration
3. If the database version is one for which Bitcoin transactions weren't stored in DB, query them from bitcoind
4. Only then proceed with the SQLite migration and pass the Bitcoin transactions which need to be stored

At the expense of making the code less encapsulated and a bit harder to follow, this gives us two nice robustness properties:
- If any Bitcoin transaction is missing (which should never happen :tm:) the migration will fail.
- For any utxo in DB its transaction is always in DB too.

I've implemented this on this branch: https://github.com/darosior/liana/commits/pr_1180. Feel free to take anything from there, of course.

jp1ac4:
Thanks for the updated approach and code.

I wonder whether for such migrations that need to query the DB (to get the txids in the present case), we should first migrate the DB to the most recent version, then query the DB, and then apply the final migration. Although I don't think it would cause any problems in the present case, the DB query used in the current code base may not always be compatible with the user's DB version.

Perhaps this could be done by passing the bitcoind connection to `maybe_apply_migrations` instead of the transactions themselves?

darosior:
I think we should strive to make migrations atomics: your database is either fully compatible with the previous version or the next version. For instance with what you suggest a user could use Liana version N+1 with a database created with Liana version N, apply the first migration, for whatever reason encounter an error a crash or just stop the application before applying the second migration, and be stuck on a database which can neither be used by Liana version N nor by Liana version N+1.

If the existing query methods on the SQLite connection struct were not appropriate to an old database we could just write a custom one for this purpose.

> Perhaps this could be done by passing the bitcoind connection to maybe_apply_migrations instead of the transactions themselves?

I precisely tried to avoid entangling the Bitcoin and Database modules.

jp1ac4:
Sorry, I wasn't clear in my description. I didn't mean that the migration from DB versions 4 to 5 would be split in two separate steps. Each migration from one DB version to the following would still be atomic.

So the function `migrate_v4_to_v5` would still perform all changes on the DB. But instead of receiving the txs, it would instead get these from bitcoind as an initial step within the function.

But I agree that keeping in mind the potential need to modify DB queries for previous DB versions should be sufficient. Indeed, that could still be required using my approach if version N - 1 is different.

I'll proceed using your code and start adding it into this PR.

jp1ac4:
Does bitcoind's watchonly wallet keep transactions that are no longer in the best chain or were never confirmed and are no longer in the mempool?

darosior:
Yes.

jp1ac4:
I've now updated it to use foreign key constraints from the coins to transactions tables and to retrieve transactions as part of the migration.

src/lib.rs
```diff
@@ -197,8 +203,34 @@ fn setup_sqlite(
...
+                .into_iter()
+                .map(|txid| bit.get_transaction(&txid).map(|res| res.tx))
+                .collect::<Option<Vec<_>>>()
+                .ok_or(StartupError::DbMigrateBitcoinTxs(
+                    "missing transaction in Bitcoin backend",
+                ))?
+        } else {
+            Vec::new()
+        };
+        sqlite.maybe_apply_migrations(&wallet_txs)?;
```
jp1ac4, line 231: I now only call this if it's not a fresh data dir.

src/lib.rs
```diff
@@ -116,6 +117,10 @@ impl fmt::Display for StartupError {
                 f,
                 "Our Bitcoin interface is bitcoind but we have no 'bitcoind_config' entry in the configuration."
             ),
+            Self::DbMigrateBitcoinTxs(msg) => write!(
+                f,
+                "Error when migrating Bitcoin transaction from Bitcoin backend to database: {}.", msg
```
jp1ac4, line 122: I removed reference to bitcoind so that message applies to Electrum and other backends when available.

src/database/sqlite/mod.rs
```diff
@@ -2101,6 +2050,14 @@ CREATE TABLE labels (
...
 
+        let bitcoin_txs: Vec<_> = (0..2)
```
jp1ac4, line 2261: This `v0_to_v2_migration()` test should either be renamed or merged with the `v0_to_v5_migration()` test.

darosior: True, it's clearly not migrating from v0 to v2.

src/database/sqlite/utils.rs
```diff
@@ -262,6 +333,11 @@ pub fn maybe_apply_migration(db_path: &path::Path) -> Result<(), SqliteDbError>
                 migrate_v3_to_v4(&mut conn)?;
                 log::warn!("Migration from database version 3 to version 4 successful.");
             }
+            4 => {
+                log::warn!("Upgrading database from version 4 to version 5.");
+                migrate_v4_to_v5(&mut conn, bitcoin_txs)?;
```
darosior, line 338: nit: we could log the number of Bitcoin transactions to be inserted in DB at the same time?

src/lib.rs
darosior: There will never be a version of Liana shipped with both Electrum and a database version <= 4. Also why i made the code below where this error may be triggered bitcoind-specific.

(no need to change it either way, it's not a big deal)

src/lib.rs
```diff
@@ -197,8 +203,34 @@ fn setup_sqlite(
...
 
+    // If opening an existing wallet whose database does not yet store the wallet transactions,
+    // query them from the Bitcoin backend before proceeding to the migration.
     let sqlite = SqliteDb::new(db_path, options, secp)?;
-    sqlite.maybe_apply_migrations()?;
+    if !fresh_data_dir {
+        let mut conn = sqlite.connection()?;
+        let wallet_txs = if conn.db_version() <= MAX_DB_VERSION_NO_TX_DB {
+            let bit = bitcoind.as_ref().ok_or(StartupError::DbMigrateBitcoinTxs(
+                "a connection to a Bitcoin backend is required",
```
darosior, line 213: nit: "to bitcoind". This code is bitcoind-specific (on purpose).

src/database/mod.rs
```diff
@@ -339,6 +342,10 @@ impl DatabaseConnection for SqliteConn {
             })
             .collect()
     }
+
+    fn get_tx(&mut self, txid: &bitcoin::Txid) -> Option<bitcoin::Transaction> {
+        self.get_tx(txid)
```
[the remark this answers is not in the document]
jp1ac4: Ah yes, I'll change it.

src/lib.rs
jp1ac4: Ah that's true, although a user could in principle change their existing daemon config file to specify an Electrum connection before starting the new version of Liana.

src/bitcoin/poller/looper.rs
```diff
@@ -196,7 +196,13 @@ fn add_txs_to_db(
...
     let txs: Vec<_> = txids
         .into_iter()
-        .map(|txid| bit.wallet_transaction(&txid).map(|(tx, _)| tx))
+        .filter_map(|txid| {
+            if db_conn.get_tx(&txid).is_none() {
+                Some(bit.wallet_transaction(&txid).map(|(tx, _)| tx))
+            } else {
+                None
+            }
+        })
```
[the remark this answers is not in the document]
jp1ac4: In order to avoid adding a new DB method for this, I'll just get the existing txids from the current DB coins. There might be additional entries in the transactions table that are not referenced by the coins table, but this shouldn't happen often.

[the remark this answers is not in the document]
darosior: both sound good to me!

src/database/sqlite/mod.rs
```diff
@@ -2423,6 +2432,7 @@ CREATE TABLE labels (
 
         // SqliteDb new is doing the migration.
         let db = SqliteDb::new(db_path, None, &secp).unwrap();
+        db.maybe_apply_migrations().unwrap();
```
darosior: The comment just above is wrong "SqliteDb::new is doing the migration". There is a similar comment at the top of the previous test which needs to be updated too.

src/database/sqlite/utils.rs
```diff
@@ -229,9 +229,80 @@ fn migrate_v3_to_v4(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError
...
+) -> Result<(), SqliteDbError> {
+    db_exec(conn, |db_tx| {
+        db_tx.execute(
+            "
+            CREATE TABLE transactions (
+                id INTEGER PRIMARY KEY NOT NULL,
+                txid BLOB UNIQUE NOT NULL,
+                tx BLOB UNIQUE NOT NULL
+            );",
+            rusqlite::params![],
```
darosior, line 244: This is missing to actually insert the transactions in DB and no single test caught it..

src/database/mod.rs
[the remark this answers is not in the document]
jp1ac4: I think we could also remove the `Option` from the return type for `TxGetter::get_tx()` as we are now guaranteed by the foreign key constraint to have a wallet transaction in the DB.

src/database/sqlite/mod.rs
```diff
@@ -682,6 +696,71 @@ impl SqliteConn {
...
+            }
+            Ok(())
+        })
+        .expect("Database must be available")
+    }
+
+    pub fn list_wallet_transactions(
+        &mut self,
+        txids: &[bitcoin::Txid],
+    ) -> Vec<DbWalletTransaction> {
```
darosior, line 719: This needs a unit test.

src/database/sqlite/mod.rs
```diff
@@ -682,6 +696,71 @@ impl SqliteConn {
...
+        // The UNION will remove duplicates.
```
darosior, line 720: I'm writing a unit test on the migration from v4 to v5 and it seems like it doesn't.

darosior:
I've written a unit to specifically exercise the migration from v4 to v5, including the failure case: https://github.com/darosior/liana/commit/c9c82f2b06183b149825da8a9a861f20bc214dc8.

It only trips on the last check which compares the transactions from DB with the transactions created in the test because i think your `list_wallet_transactions` returns duplicates. See https://github.com/wizardsardine/liana/pull/1180#discussion_r1700312663.

jp1ac4:
The failing test may be due to the spend_txid having multiple different block infos, whereas `list_wallet_transactions` assumes that a given txid can't have inconsistent block info in the coins table.

darosior:
> The failing test may be due to the spend_txid having multiple different block infos, whereas `list_wallet_transactions` assumes that a given txid can't have inconsistent block info in the coins table.

Ah, good catch. Fixing it now.

src/database/sqlite/mod.rs
jp1ac4: I've now merged it with the `v0_to_v5_migration()` test.

src/database/mod.rs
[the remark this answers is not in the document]
jp1ac4: > I think we could also remove the `Option` from the return type for `TxGetter::get_tx()` as we are now guaranteed by the foreign key constraint to have a wallet transaction in the DB.

That's as long as we only pass expected `txid` values, so best to leave it as `Option` I think.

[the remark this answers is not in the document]
jp1ac4: I've removed the `get_tx` database method now.

src/bitcoin/poller/looper.rs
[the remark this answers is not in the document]
jp1ac4: I've added a `list_saved_txids()` database method now and use this as a filter.

src/bitcoin/poller/looper.rs
```diff
@@ -156,6 +159,51 @@ fn update_coins(
...
 }
 
+// Add new deposit and spend transactions to the database.
+fn add_txs_to_db(
+    bit: &impl BitcoinInterface,
+    db_conn: &mut Box<dyn DatabaseConnection>,
+    updated_coins: &UpdatedCoins,
+) {
+    // Get the outpoints of new/updated coins and their spend txids, if any.
+    let mut outpoints = HashMap::<bitcoin::OutPoint, Option<bitcoin::Txid>>::new();
```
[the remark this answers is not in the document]
jp1ac4: That's true, have changed it to just a set of txids.

src/database/sqlite/mod.rs
```diff
@@ -160,9 +164,14 @@ impl SqliteDb {
         Ok(SqliteDb { db_path })
     }
 
-    /// If the database version is older than expected, migrate it to the current version.
-    pub fn maybe_apply_migrations(&self) -> Result<(), SqliteDbError> {
-        maybe_apply_migration(&self.db_path)
+    /// If the database version is older than expected, migrate it to the current version. If
+    /// migrating from a database version 4 or earlier, all the wallet Bitcoin transactions must be
+    /// passed through the `bitcoin_txs` parameter otherwise the transaction will fail.
```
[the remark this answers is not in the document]
jp1ac4: I've changed this now.

src/database/sqlite/utils.rs
jp1ac4: I added another `log::warn` message here.

src/lib.rs
jp1ac4: I've left these as they are for now but I see your point.

src/database/sqlite/mod.rs
jp1ac4: Thanks, I think I've removed them all now. There was one place where the test avoided using `SqliteDb::new()` specifically not to do the migration, and so I changed it now to use `SqliteDb::new()` and removed the comment.

src/database/sqlite/mod.rs
jp1ac4: There was a simple unit test, but I've updated it with your migration test in order for there to be multiple outpoints for a given txid.

src/database/sqlite/mod.rs
jp1ac4: Thanks, I've cherry-picked your commit and made some small changes.

jp1ac4:
I think I've now made all the requested changes.

src/database/sqlite/mod.rs
```diff
@@ -729,6 +729,50 @@ impl SqliteConn {
...
+                .map(|txid| format!("x'{}'", FrontwardHexTxid(*txid)))
+                .collect::<Vec<_>>()
+                .join(",")
+        );
+        let w_txs: Vec<DbWalletTransaction> =
+            db_query(&mut self.conn, &query, rusqlite::params![], |row| {
+                row.try_into()
+            })
+            .expect("Db must not fail");
+        assert_eq!(
```
jp1ac4, line 764: I added this assertion.

darosior: Ok. Can we make it a debug_assert? It's not ideal to perform an allocation just for an assertion.

darosior:
Tested migrating from v4.0, v5.0 and v6.0 using the migration functional test.

darosior:
Tested by upgrading a mainnet wallet. No issue.

darosior:
ACK 3953a0b29e4d8dba45074a7d351be6c8cea6ecfc

src/database/sqlite/mod.rs
jp1ac4: Sure, I'll change it.

jp1ac4: Have changed it now.

darosior:
re-ACK bf1e90e0fa7f357dc5fb20c12e454f58433dd898

jp1ac4:
Thanks for the review and additional commits :smiley:

darosior:
Thanks for bearing through the many iterations!

Issue #962, opened 2024-02-07
[GUI] no separator for thousands


kloaec:
The balance display is really cool for the separations of the decimals and making sats visible.
We do not have a separation of the thousands though!

There are different ways to approach this, either the American way XX,XXX.XX or just spacing XX XXX.XX.
I prefer the space as a more "global" approach, given that some countries use the comma and other the dot for the decimals, I don't want them to be confused.

The same should be done for the millions.

![image](https://github.com/wizardsardine/liana/assets/15522963/4c1f34e0-03d6-4aea-8f00-5d1350261b36)

pythcoiner:
I also in favor of space. (never test w/ such amount tbh :grinning:

Nonnyjoe:
I'll Like to handle this issue

darosior:
@Nonnyjoe welcome to the project. Sounds good, i'm assigning this issue to you.

aaroncarlucci:
I looked into this a bit recently while attempting to onboard to the project. Unfortunately, haven't come up yet with a fix, but decided to leave some initial, possibly ignorant feedback here from someone picking up both Rust and the Liana stack.
* FYI - there were some very recent changes WRT to `bitcoin::Amount` formatting committed recently to the [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/commit/014c4931be9edb530b1db15e14feeca5506cad6b) crate. I think that crate might support passing in a custom formatter to render amounts?
* That crate also makes use of overriding `fmt::Display` to define custom formatting. I played around locally with this, but it was over my head at the moment. Perhaps there is an opportunity to subclass `bitcoin::Amount` and define a custom formatter? Could be a good opportunity to refactor what appears to be a growing if/else statement in amount.rs?
* Another option might be to consider the [num_format](https://docs.rs/num-format/latest/num_format) crate and use "French" formatting, or define a custom format. Another dependency, though... and doesn't support floats, so render of BTC values might need to be split on the decimal and rendered separately?

Maybe I'll be able to come up with some code eventually, but hoping that sharing my research thus far helps somehow.

pythcoiner:
Hi @aaroncarlucci welcome on board! 
I think we should not add any dependency for this.

In my mind the most straigtforward is to rework the logic in `amount.rs::split_digits()`:
- splitting `s` w/ '.'
- then chunk & join both sides (integer part should be done in reverse)
```rust
    let decimal = decimal
        .chars()
[...]

PR #1182, opened 2024-07-09
GUI: Add space separator to integer portion of BTC amount rendering

aaroncarlucci:
This PR attempts to address #962, refactoring the amount component to support an internationalized display format and fixing a bug where the integer portion of the decimal didn't support the space separator for BTC values.

I'm attempting to learn Rust, Bitcoin programming, the `iced` GUI library and the Liana stack simultaneously, so this took altogether waaaaay too long, but it ended up being a great onboarding task to start wrapping my head around Rust strings, Strings, slices, vectors, iterators, and rummaging around in the documentation. That said, there are probably plenty of small details or oversights, so pile on the feedback. Here are a few review topics of interest:

* It didn't make sense to me at first that `split_digits()` was responsible for adding GUI rows and formatting details like `bold()` and text size, so I ripped that out in favor of a function that focused on formatting the decimal amount, returning a tuple containing strings for the integer and fraction portions of the number. This resulted in probably removing some embedded `Row` objects. I didn't see a problem with this, but please advise if it was too aggressive.
* I wrote a quick test for the formatting in `amount.rs` which can be run from `gui/ui` using `cargo test --release`. Maybe there is a better place for this? Other test cases that should be added?
* The only thing I could observe between the `amount_with_size()` and `unconfirmed_amount_with_size()` functions was that the former was applying bold formatting to the integer part of the number. I removed a lot of duplicate code between these functions and reduced them to focusing on building the GUI elements, rather than number formatting logic. I hope this is right, but could be missing something as it was rather hard to read.

There is probably a lot of room for improvement. Thanks :)

### DIFF
--- a/gui/ui/src/component/amount.rs
+++ b/gui/ui/src/component/amount.rs
@@ -95,6 +95,22 @@ fn split_digits<'a, T: 'a>(mut s: String, size: u16, bold: bool) -> impl Into<El
                 });
         }
     }
+
+    // Reformat the integer portion of the amount with space separation.
+    let (integer, fraction) = match s.split_once('.') {
+        Some((i, f)) => (i, f),
+        None => (s.as_str(), "00 000 000"),
+    };
+    let mut integer = integer
+        .chars()
+        .collect::<Vec<_>>()
+        .rchunks(3)
+        .map(|c| c.iter().collect::<String>())
+        .collect::<Vec<_>>();
+    integer.reverse();
+    let integer = integer.join(" ");
+    s = format!("{integer}.{fraction}");
+
     if bold {
         Row::new().push(text(s).bold().size(size))
     } else {


aaroncarlucci:
As an afterthought, the `fraction` portion could also be expressed by padding 8 zeros and then formatting with slices, like:

```
    let fraction = format!("{:0<8}", fraction);
    let fraction = format!(
        "{} {} {}",
        &fraction[0..2],
        &fraction[2..5],
        &fraction[5..8]
    );
```

pythcoiner:
not yet review the code but formatting in this PR changed from the original one:

original:

image

this PR:

image

aaroncarlucci:
@pythcoiner Thanks for the screenshots. Really helps to understand the intent of the original code -- looks like bold formatting should be applied after the occurrence of a non-zero digit? It seems my changes were indeed too aggressive.

I'm going to take another shot at this preserving more of the original logic, and then perhaps try another iteration of refactoring.

Also -- I'd like to hit you up on Discord sometime for some advice on setting up the local development environment with test data.

pythcoiner:
the bold stuff and also if amount < 1BTC , zeros before _sat_ part are greyed out

aaroncarlucci:
I force-pushed another, less intrusive change that I think should work and preserve the existing formatting rules earlier. Seems good on my end, but I would be interested to know if it looks acceptable on your data set as well, if you find some time to run the build.

Still might take a shot at refactoring, but I think the original display requirements are back intact.

pythcoiner:
looks ok now:
image

ACK c6b8133286a4009cc1e72935e30af36d43705237

thanks for the PR!

edouardparis:
Hello, thank you for the change. It seems that something is not good on the last @pythcoiner screenshot:
the first two lines:
+ 0.00 123.00 000 0000

aaroncarlucci:
@edouardparis would you please share the correct integer amount that is supposed to be represented? My guess here is `0.00799800` sats.

edouardparis:
20240715_16h18m55s_grim

gui/ui/src/component/amount.rs
```diff
@@ -95,6 +95,22 @@ fn split_digits<'a, T: 'a>(mut s: String, size: u16, bold: bool) -> impl Into<El
                 });
         }
     }
+
```
edouardparis, line 98: Suggestion: maybe do
```
let is_more_than_one_bitcoin = false; 
for prefix in prefixes {
    if s.starts.with(prefix) {
          is_more_than_one_bitcoin = true;
    ...}
}

if is_more_than_one_bitcoin {
  /// your code
}

aaroncarlucci:
FYI - I'm making an attempt to rewrite the amount component rendering, as an exercise to improve my understanding, and hopefully, the readability / organization. Will do it in a separate PR, and if it doesn't work or make sense we can discard it.

That said, can you please clarify the difference in stylistic requirements between `amount_with_size()` and `unconfirmed_amount_with_size()`? While the functions look largely identical, my read is that the display of unconfirmed amounts don't render a portion of the text in bold. Is this accurate?

aaroncarlucci:
Refactored approach submitted in #1198.

PR #1189, opened 2024-07-11
installer: reset signing device registration if descriptor change

### LINKED ISSUE #1157: If we register descriptor then edit policy, we cannot register desciptor anymore
On colcard, but should also happen w/ sprecter/bitbox/jade

pythcoiner:
closes #1157

### DIFF
--- a/gui/src/installer/step/descriptor.rs
+++ b/gui/src/installer/step/descriptor.rs
@@ -1394,6 +1394,11 @@ impl RegisterDescriptor {
 
 impl Step for RegisterDescriptor {
     fn load_context(&mut self, ctx: &Context) {
+        // we reset device registered set if the descriptor have changed.
+        if self.descriptor != ctx.descriptor {
+            self.registered = Default::default();
+            self.done = false;
+        }
         self.descriptor.clone_from(&ctx.descriptor);
         let mut map = HashMap::new();
         for key in ctx.keys.iter().filter(|k| !k.name.is_empty()) {


gui/src/installer/step/descriptor.rs
```diff
@@ -1399,6 +1401,11 @@ impl Step for RegisterDescriptor {
         for key in ctx.keys.iter().filter(|k| !k.name.is_empty()) {
             map.insert(key.master_fingerprint, key.name.clone());
         }
+        // we reset device registered set if the descriptor have changed
+        if self.registered_descriptor != self.descriptor {
+            self.registered = Default::default();
+            self.registered_descriptor.clone_from(&self.descriptor);
+        }
```
edouardparis, line 1408: I think we do not need to have a new field. example:
```rust
fn load_context(&mut self, ctx: &Context) {
        // we reset device registered set if the descriptor have changed
        if self.descriptor != ctx.descriptor {
            self.registered = Default::default();
        }
        self.descriptor.clone_from(&ctx.descriptor);
        let mut map = HashMap::new();
        for key in ctx.keys.iter().filter(|k| !k.name.is_empty()) {
            map.insert(key.master_fingerprint, key.name.clone());
        }
     }
```

pythcoiner: adressed

jp1ac4 (commented):
Tested this successfully.

I think we should also clear the checkbox "I have registered the descriptor on my device(s)".

pythcoiner:
i'll do

edouardparis (approved):
ACK fe595b2448f105b45413218b5c1319678d4d5860

PR #1190, opened 2024-07-11
disable previous button if device processing

### LINKED ISSUE #1158: [Installer] if we move screen back during descriptor registration we get stuck
Step to reproduce:
- go up to registration process
- start to register descriptor on ledger
- go back previous screen
- cancel registration on ledger
- go to registration screen

the state of device is still at 

```
Processing...
Please check your device
```

Note: this state is tied to the device index in the list, not to the device itself, for instance, if when we get back to the registration screen the device order have changed (one unplugged) the step will be at the same index, not following device.

pythcoiner:
This PR diasable the previous button  at the descriptor registration step if a device is still processing, avoiding issues like #1158.
The other way around can be to reset the state of `self.processing` if user go back, but look less clean to me.

closes #1158

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -750,7 +750,11 @@ pub fn register_descriptor<'a>(
             })
             .spacing(50),
         true,
-        Some(Message::Previous),
+        if !processing {
+        Some(Message::Previous)
+        } else {
+            None
+        }
     )
 }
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
utACK f457dd447cd4af89aace9f56e085073590dde8e5

PR #1198, opened 2024-07-16
GUI: Refactor amount formatting

aaroncarlucci:
This PR is based on #1182 and aims to actually refactor the amount component while adding a space separator to the integer portion of BTC amount displays as per #962.

The code organization is reworked to try and separate the conversion of `bitcoin::Amount` type to a `iced` renderable `Row` elements in steps via smaller functions by:

1. Converting the `Amount` type to a string with the integer and fraction portions formatted in space-separated three digit chunks.
2. Detecting where in that string the non-zero BTC amount occurs
3. Converting the string into parts and rendering the preceding zeros and spaces with normal styling while applying a bold `Row` render element for the significant amount.

To me, this approach makes the code easier to read and is a small step toward separating the string formatting functionality from the render element building. If we find more bugs with this approach, or the team simply doesn't like the reorganization, it's fine to scrap it, as it was a decent learning exercise anyway. Looking forward to feedback.

### DIFF
--- a/gui/ui/src/component/amount.rs
+++ b/gui/ui/src/component/amount.rs
@@ -95,6 +95,22 @@ fn split_digits<'a, T: 'a>(mut s: String, size: u16, bold: bool) -> impl Into<El
                 });
         }
     }
+
+    // Reformat the integer portion of the amount with space separation.
+    let (integer, fraction) = match s.split_once('.') {
+        Some((i, f)) => (i, f),
+        None => (s.as_str(), "00 000 000"),
+    };
+    let mut integer = integer
+        .chars()
+        .collect::<Vec<_>>()
+        .rchunks(3)
+        .map(|c| c.iter().collect::<String>())
+        .collect::<Vec<_>>();
+    integer.reverse();
+    let integer = integer.join(" ");
+    s = format!("{integer}.{fraction}");
+
     if bold {
         Row::new().push(text(s).bold().size(size))
     } else {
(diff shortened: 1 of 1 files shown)


gui/ui/src/component/amount.rs
```diff
@@ -3,38 +3,82 @@ pub use bitcoin::Amount;
...
+// Build the rendering elements for displaying a Bitcoin amount.
+// The text should be bolded beginning where the BTC amount is non-zero.
+fn render_amount<'a, T: 'a>(amount: String, size: u16) -> Row<'a, T> {
     let spacing = if size > P1_SIZE { 10 } else { 5 };
-    let sats = format!("{:.8}", a.to_btc());
-    assert!(sats.len() >= 9);
+
+    let (before, after) = match split_at_first_non_zero(amount) {
+        Some((b, a)) => (b, a),
+        None => panic!("Non-zero integer character not found."),
```
edouardparis, line 75: 0 amount output transaction are not relayed by Core but are possible at the consensus level. I would prefer that you remove the panic and simply return two empty string.

edouardparis (commented):
Thank you for this PR, the change looks good, I just have a nit.

gui/ui/src/component/amount.rs
aaroncarlucci: Np -- thanks. I wasn't quite sure what to do there anyway...

edouardparis (approved):
ACK 59035060e9b2b911809fdc225bfa862244f5f6d6

pythcoiner:
it looks to me we introduce #1218 w/ this PR

PR #1200, opened 2024-07-18
gui: update `SettingState.wallet` on `WalletUpdated()` message

### LINKED ISSUE #1170: [GUI] Updated wallet settings sometimes not displayed
This happens when clicking on "Wallet" in "Settings -> Wallet" after updating a value.

To reproduce, go to Settings -> Wallet:

image

Change a fingerprint alias and click "Update":

image

Then click on "Wallet" in "Settings -> Wallet" and the previous value is shown:

image

The updated value is shown again after navigating away from this page and returning.

pythcoiner:
This PR fixes #1170.

result Message from `update_key_aliases()` emited [here](https://github.com/wizardsardine/liana/blob/0a7ff2b0ea1b6f4e98946ccc1201a8d1caec8bf5/gui/src/app/state/settings/wallet.rs#L152) was not catch in `SettingState.update()`.

### DIFF
--- a/gui/src/app/state/settings/mod.rs
+++ b/gui/src/app/state/settings/mod.rs
@@ -84,6 +84,13 @@ impl State for SettingsState {
                     .map(|s| s.reload(daemon, wallet))
                     .unwrap_or_else(Command::none)
             }
+            Message::WalletUpdated(Ok(wallet)) => {
+                self.wallet = wallet.clone();
+                self.setting
+                    .as_mut()
+                    .map(|s| s.update(daemon, cache, message))
+                    .unwrap_or_else(Command::none)
+            }
             _ => self
                 .setting
                 .as_mut()


gui/src/app/state/settings/mod.rs
```diff
@@ -84,6 +84,10 @@ impl State for SettingsState {
                     .map(|s| s.reload(daemon, wallet))
                     .unwrap_or_else(Command::none)
             }
+            Message::WalletUpdated(Ok(wallet)) => {
```
edouardparis, line 87: The message must also be passed to the self.setting update, otherwise the "Updated" green text is not shown and the button keeps the "Processing" state.

pythcoiner: done

edouardparis (approved):
ACK e34062a417a3533a2c42f776a495c2810775983b

PR #1201, opened 2024-07-18
gui: TransactionPanel.update() -> update label if tx already exists

### LINKED ISSUE #1174: [GUI] Transaction labels don't refresh from PSBTs screen
After editing the label of a PSBT, the updated value doesn't appear on the Transactions page (either list view or after selecting the individual transaction), unless the Transactions page has not yet been loaded since starting the GUI.

The updated label appears on Transactions page when restarting the GUI.

The other way works: a label that is edited on the Transaction page appears on the PSBTs page.

image

image

pythcoiner:
close #1174 

labels were not updated if tx already exists

### DIFF
--- a/gui/src/app/state/transactions.rs
+++ b/gui/src/app/state/transactions.rs
@@ -99,7 +99,9 @@ impl State for TransactionsPanel {
                 Ok(txs) => {
                     self.warning = None;
                     for tx in txs {
-                        if !self.txs.iter().any(|other| other.tx == tx.tx) {
+                        if let Some(t) = self.txs.iter_mut().find(|other| other.tx == tx.tx) {
+                            t.labels = tx.labels;
+                        } else {
                             self.txs.push(tx);
                         }
                     }


pythcoiner:
@edouardparis wdyt if we change (maybe in a follow-up) [HistoryTransaction.txs](https://github.com/wizardsardine/liana/blob/48d34904b0f29035062260632676f78baf9f04f3/gui/src/app/state/transactions.rs#L39) to an HashMap? it should be better for perf w/ huge amount of txs?

edouardparis:
The list of transactions mus be sorted. If we are using a hashmap, we will have to sort the list each time the view is reloaded, I think it is better then to use a BTreeMap that keeps the items order.

pythcoiner:
> The list of transactions must be displayed sorted. If we are using a hashmap, we will have to sort the list each time the view is reloaded (in the view method), I think it is better then to use a BTreeMap that keeps the items order and allows efficiency of research.

right

edouardparis (approved):
ACK fdf01cc095787c5cf947841c1902a22d500b81d7

PR #1209, opened 2024-07-19
gui: disable replacement fee input if tx is already replaced

### LINKED ISSUE #1138: [GUI] RBF modal feerate is still editable once replacement PSBT generated
After generating a fee-bumped PSBT, it's still possible to edit the feerate, but this won't do anything.

Either the feerate should become read-only in this case or editing the feerate should give the option to generate a new replacement PSBT.

image

pythcoiner:
this PR disable input on feerate TextInput if the tx already have been replaced, it's fine to do so because we still can RBF the new replacing tx.

image

close #1138

### DIFF
--- a/liana-ui/src/component/form.rs
+++ b/liana-ui/src/component/form.rs
@@ -46,6 +46,19 @@ where
         }
     }
 
+    /// Creates a new [`Form`] that has a disabled input.
+    ///
+    /// It expects:
+    /// - a placeholder
+    /// - the current value
+    pub fn new_disabled(placeholder: &str, value: &Value<String>) -> Self {
+        Self {
+            input: text_input::TextInput::new(placeholder, &value.value),
+            warning: None,
+            valid: value.valid,
+        }
+    }
+
     /// Creates a new [`Form`] that trims input values before applying the `on_change` function.
     ///
     /// It expects:
--- a/liana-gui/src/app/view/transactions.rs
+++ b/liana-gui/src/app/view/transactions.rs
@@ -257,13 +257,17 @@ pub fn create_rbf_modal<'a>(
                         .push(Container::new(p1_bold("Feerate")).padding(10))
                         .spacing(10)
                         .push(
-                            form::Form::new_trimmed("", feerate, move |msg| {
-                                Message::CreateRbf(CreateRbfMessage::FeerateEdited(msg))
-                            })
-                            .warning(
-                                "Feerate must be greater than previous value and \
-                                less than or equal to 1000 sats/vbyte",
-                            )
+                            if replacement_txid.is_none() {
+                                form::Form::new_trimmed("", feerate, move |msg| {
+                                    Message::CreateRbf(CreateRbfMessage::FeerateEdited(msg))
+                                })
+                                .warning(
+                                    "Feerate must be greater than previous value and \
+                                    less than or equal to 1000 sats/vbyte",
+                                )
+                            } else {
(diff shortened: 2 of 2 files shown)


gui/ui/src/component/form.rs
```diff
@@ -45,6 +45,19 @@ where
         }
     }
 
+    /// Creates a new [`Form`] that have a disabled input.
+    ///
+    /// It expects:
+    /// - a placeholder
+    /// - the current value
+    pub fn new_disabled(placeholder: &str, value: &Value<String>) -> Self {
```
jp1ac4, line 53: Perhaps instead of adding this method, you could reuse the `new()` method by changing the `on_change` parameter to `Option`.

What do you think, @edouardparis?

gui/src/app/view/transactions.rs
```diff
@@ -244,13 +244,17 @@ pub fn create_rbf_modal<'a>(
...
-                            })
-                            .warning(
-                                "Feerate must be greater than previous value and \
+                            if replacement_txid.is_none() {
+                                form::Form::new_trimmed("", feerate, move |msg| {
+                                    Message::CreateRbf(CreateRbfMessage::FeerateEdited(msg))
+                                })
+                                .warning(
+                                    "Feerate must be greater than previous value and \
                                 less than or equal to 1000 sats/vbyte",
```
jp1ac4, line 253: nit: This line might look better with an extra tab to be aligned with the line above.

pythcoiner: done

gui/ui/src/component/form.rs
jp1ac4: I had a couple of further thoughts:
1) Unrelated to this PR (perhaps could be a separate commit), we could refactor the `new_trimmed()` and `new_amount_btc()` methods to call `Self::new()` with the modified `on_change` parameter.
2) If these other methods also use `Option`, then you can use the `new_trimmed()` whether or not the replacement txid exists, and the only thing that will change is whether you pass `None` or `Some` to the `on_change` parameter (rather than different methods being called in each case).

nondiremanuel:

pythcoiner:
ping @edouardparis

gui/ui/src/component/form.rs
edouardparis: We can have new_disabled for now. I prefer we keep the change small.

edouardparis:
> https://github.com/wizardsardine/liana/pull/1209#discussion_r1912912232

I just answered here, if it was the reason of your ping.

pythcoiner:
rebased on master, this is ready for review given previous [comment](https://github.com/wizardsardine/liana/pull/1209#discussion_r1912912232)

liana-ui/src/component/form.rs
```diff
@@ -46,6 +46,19 @@ where
         }
     }
 
+    /// Creates a new [`Form`] that have a disabled input.
```
jp1ac4, line 49: "have" -> "has"

jp1ac4 (commented):
Tested 0017c195212f924c1531d4154c4d28c904636f36 successfully.

There's one small fix to the docstring that I think is worth making.

liana-ui/src/component/form.rs
pythcoiner: done

jp1ac4 (approved):
tACK https://github.com/wizardsardine/liana/commit/ad2c096715da49223f7c2cf3389e575b7e62a064.

PR #1214, opened 2024-07-23
gui: add the version in the window title bar

### LINKED ISSUE #1181: [GUI] show version in the "installer"
We currently only show the version in the Settings, which is after the installer.
Could we display a version number at the bottom/top of the window at all time? Or even in the title of the window, as long as it's easy for support to explain where to find it.

pythcoiner:
This PR add the version (major+minor) into the window titlebar, for both installer & wallet.
close #1181 

image

image

### DIFF
--- a/gui/src/main.rs
+++ b/gui/src/main.rs
@@ -119,8 +119,8 @@ impl Application for GUI {
 
     fn title(&self) -> String {
         match self.state {
-            State::Installer(_) => String::from("Liana Installer"),
-            _ => String::from("Liana"),
+            State::Installer(_) => format!("Liana v{} Installer", VERSION),
+            _ => format!("Liana v{}", VERSION),
         }
     }
 


jp1ac4:
Just to check, do we prefer uppercase over lowercase for the "v" here?

nondiremanuel:
I would go with lower case, but that's not a strong opinion

darosior:
Agree with Manuel. I think it's customary to use a lowercase v.

gui/src/main.rs
```diff
@@ -119,8 +119,8 @@ impl Application for GUI {
 
     fn title(&self) -> String {
         match self.state {
-            State::Installer(_) => String::from("Liana Installer"),
-            _ => String::from("Liana"),
+            State::Installer(_) => format!("Liana v{}.{} Installer ", VERSION.major, VERSION.minor),
```
jp1ac4, line 122: Is the extra space at the end intentional?

pythcoiner: (for trigger CI)

jp1ac4: Ah OK, I guess better to remove it again though?

pythcoiner: done

jp1ac4 (approved):
Tested ACK 5c6bef667573386330ff3ad06ec7774901b3f638.

gui/src/main.rs
```diff
@@ -119,8 +119,8 @@ impl Application for GUI {
 
     fn title(&self) -> String {
         match self.state {
-            State::Installer(_) => String::from("Liana Installer"),
-            _ => String::from("Liana"),
+            State::Installer(_) => format!("Liana v{}.{} Installer", VERSION.major, VERSION.minor),
+            _ => format!("Liana v{}.{}", VERSION.major, VERSION.minor),
```
darosior, line 123: `Version` already implements `Display`. Let's use the existing implementation instead of reimplementing it:
```suggestion
            State::Installer(_) => format!("Liana v{} Installer", VERSION),
            _ => format!("Liana v{}", VERSION),
```

This also carries the update used for dev branches vs release branches.

pythcoiner: done

darosior (commented):
ACK 79f3f5968f5d97e2588bd9f3a08d9d5b8cbbaefc

PR #1215, opened 2024-07-24
Ledger upgrade feature

### LINKED ISSUE #1135: Upgrade Bitcoin app version if above minimum but below Taproot minimum
Enough people have been testing https://github.com/darosior/ledger_installer that i feel comfortable starting to integrate pieces into Liana. A good small first step would be to make it possible for Ledger users with a Bitcoin app whose version is below the version which introduced Taproot support to upgrade to the latest Bitcoin app. Note however those users must already have a Liana-compatible app (above our minimum supported version) otherwise the update of the app would also necessitates a firmware upgrade, which is not implemented at the moment (and not the best candidate for a first step).

pythcoiner:
This PR introduce a feature to upgrade the ledger app if an app already supporting miniscript but not taproot installed on device.
If upgrade possible this is displayed to user:

image

After clicking on `Upgrade` liana close the ledger bitcoin app and start upgrade:

image

After upgrading, the user should validate app opening on device.

Note: There is some ledger firmware that allows to install app that support miniscript but do not allow install app that support taproot (i hit this whith Nano S+ w/ firmware 1.1.0 the latest upgradable app is 2.1.3), to handle this case we perform a version check after upgrade, if the latest app does not support taproot (< 2.2.0) we ask user to upgrade his firmware with Ledger Live:

image

This feature is implemented in several locations:
- Installer: Descriptor creation step
- Installer: Descriptor registration step
- App: PSBT signing step
- App: Settings for descriptore registration step
- App: Receive for address verification step

Note: This PR rely on changes made in [`ledger_manager`](https://github.com/darosior/ledger_installer/pull/16).

closes #1135

### DIFF
--- a/gui/src/hw.rs
+++ b/gui/src/hw.rs
@@ -1,16 +1,24 @@
-use iced::Command;
+use iced::{Command, Subscription};
+use ledger_manager::utils::InstallStep;
 use std::{
     collections::HashMap,
     path::PathBuf,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
-use crate::app::{settings, wallet::Wallet};
+use crate::{
+    app::{settings, wallet::Wallet},
+    ledger_upgrade::{
+        ledger_upgrade_subscriptions, maybe_start_upgrade, update_upgrade_state, UpgradeMessage,
+    },
+};
 use async_hwi::{
     bitbox::{api::runtime, BitBox02, PairingBitbox02},
     coldcard,
     jade::{self, Jade},
-    ledger, specter, DeviceKind, Error as HWIError, Version, HWI,
+    ledger::{self, DeviceInfo, HidApi},
+    specter, DeviceKind, Error as HWIError, Version, HWI,
 };
 use liana::miniscript::bitcoin::{bip32::Fingerprint, hashes::hex::FromHex, Network};
 use serde::{Deserialize, Serialize};
@@ -24,6 +32,7 @@ pub enum UnsupportedReason {
     Method(&'static str),
     NotPartOfWallet(Fingerprint),
     WrongNetwork,
+    Taproot,
 }
 
 // Todo drop the Clone, to remove the Mutex on HardwareWallet::Locked
@@ -50,6 +59,17 @@ pub enum HardwareWallet {
         version: Option<Version>,
         registered: Option<bool>,
         alias: Option<String>,
+    },
+    NeedUpgrade {
+        id: String,
+        kind: DeviceKind,
+        fingerprint: Fingerprint,
+        version: Option<Version>,
+        upgrade_in_progress: bool,
+        upgrade_step: Option<InstallStep<Version>>,
+        upgrade_log: Vec<String>,
+        upgrade_testnet: bool,
+        upgraded_version: Option<Version>,
     },
 }
 
@@ -84,11 +104,12 @@ impl HardwareWallet {
         })
     }
 
-    fn id(&self) -> &String {
+    pub fn id(&self) -> &String {
         match self {
             Self::Locked { id, .. } => id,
             Self::Unsupported { id, .. } => id,
             Self::Supported { id, .. } => id,
+            Self::NeedUpgrade { id, .. } => id,
         }
     }
 
@@ -97,6 +118,7 @@ impl HardwareWallet {
             Self::Locked { kind, .. } => kind,
             Self::Unsupported { kind, .. } => kind,
             Self::Supported { kind, .. } => kind,
+            Self::NeedUpgrade { kind, .. } => kind,
         }
     }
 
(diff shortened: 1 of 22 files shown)


pythcoiner:
@darosior i actually display the upgrade progress logs in a 'scrollable' console, what do you think?

image

(we can add a progress bar after 'Upgrading:' also)

nondiremanuel:
I'm not sure the logs are adding much value to the user. I personally think it clutters the UI for little or no value to the user, which only cares about the progress and the success / failure statuses.

darosior:
Yeah. Do you think some are necessary to show the users? I can think of "please confirm on your device" but don't have others on the top of my mind. If it's only this one we could even just have this message: "Upgrading, please check your device".

pythcoiner:
Several are necessary,as it can takes some time to establish the connexion w/ the ledger HSM (w/ bad connection), and imho it can feel very confusing as sometimes nothing changing on device screen.

we should at least display:
- quit the bitcoin app in order to start upgrading
- accept app listing
- accept to upgrade the app

those are 3 actions the user had to perform.

pythcoiner:
note to myself: seems we can get the install progress [see](https://github.com/LedgerHQ/ledger-live/blob/5a0a1aa5dc183116839851b79bceb6704f1de4b9/libs/ledger-live-common/src/hw/installApp.ts#L31)

gui/src/hw.rs
```diff
@@ -341,60 +468,156 @@ impl HardwareWallets {
...
+    sleep: bool,
 }
 
 async fn refresh(mut state: State) -> (HardwareWalletMessage, State) {
-    let api = if let Some(api) = &mut state.api {
-        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
-        if let Err(e) = api.refresh_devices() {
-            return (HardwareWalletMessage::Error(e.to_string()), state);
-        };
+    // do not sleep on first call
```
edouardparis, line 519: We do not need the .sleep field, we can use `state.api.is_some()` to detect if it is the first call

gui/src/hw.rs
```diff
@@ -341,60 +468,156 @@ impl HardwareWallets {
...
     keys_aliases: HashMap<Fingerprint, String>,
     wallet: Option<Arc<Wallet>>,
     connected_supported_hws: Vec<String>,
     api: Option<ledger::HidApi>,
     datadir_path: PathBuf,
+    taproot: bool,
+    hws: Vec<HardwareWallet>,
+    still: Vec<String>,
+    hws_upgrade: Vec<HardwareWallet>,
+    still_upgrade: Vec<String>,
```
edouardparis, line 514: Previously we did not need the still field and now we have
still and still_upgrade. Can we just pass one unique `still: &mut Vec<String>` that we fill with ids from "ready" device and `upgrading` to the poll_ledger etc
and use it to pass device new connection has they are either ready or in upgrade state ?

I may miss a subtility

gui/src/installer/step/descriptor.rs
```diff
@@ -1301,7 +1325,14 @@ impl Step for RegisterDescriptor {
         true
     }
     fn subscription(&self, hws: &HardwareWallets) -> Subscription<Message> {
-        hws.refresh().map(Message::HardwareWallets)
+        let mut subs = ledger_upgrade_subscriptions(hws);
```
edouardparis, line 1328: I think it is better to encapsulate this to the hw module and have only
```
 fn subscription(&self, hws: &HardwareWallets) -> Subscription<Message> {
     hws.subscription().map(Message::HardwareWallets) // include the refresh and the upgrading subscription.
 }
 ```

Comment (edouardparis) [reply gui/src/installer/step/descriptor.rs:1328]: Ah I understand it is a problem of message.

File: gui/src/installer/step/descriptor.rs [hunk 4]
```diff
@@ -989,17 +992,29 @@ impl DescriptorEditModal for EditXpubModal {
                     }
                 }
             },
+            Message::UpgradeLedger(id, network) => {
```
edouardparis, line 995: message could be instead a Message::HardwareWallets(hws::HardwareWalletMessage::UpgradeLedger(id))
and then would be handle not here but in the upper level in the hws.update

gui/src/hw.rs
pythcoiner: `still_upgrade` & `hws_upgrade` need NOT to be reset at the end of `refresh()`: `hws` & `still` are need to be reset, they are not stricly needed in the state, but i add them for convenience & readability as we need to pass them to all `poll_xxx()` and `handle_xxx_device()`

gui/src/installer/step/descriptor.rs
pythcoiner: I handle the message here (and in other modals) as i need to access to `EditXpubModal.upgrading` to disable th `on_blur()` action to edit the modal.

Moving the handling to `hws.update()` will centralized the handling but could add some complexity no notify each modal of the `upgrading` state change(some modals in App and some in Installer).

nondiremanuel:

Issue #1218, opened 2024-07-30
No amount displayed for 0.0BTC


pythcoiner:
intruduces by #1198 i guess
![image](https://github.com/user-attachments/assets/344a7201-080c-4502-a88f-8edab58ce9d8)

pythcoiner:
and i'm not sure we previously had this space:
![image](https://github.com/user-attachments/assets/92dd5f64-66c8-48e7-8c9d-dd9befa5ebd1)

edouardparis:
Arf good catch

aaroncarlucci:
I've taken a look at this and submitted a potential solution in #1219. Removing the spacing has fixed the issue @pythcoiner points out in the second screenshot. I haven't reproduced the conditions in the first screenshot locally, but I assumed it has to do with defaulting blank strings? I've made the `None` case return "0.00000000".

I hope it works, just let me know if you see other issues.

nondiremanuel:
This issue should have been addressed by #1219. Thanks for that!

PR #1219, opened 2024-07-30
GUI: Fix amount rendering

aaroncarlucci:
This PR attempts to address #1218 by removing extra spacing for amount rendering and making the default for amounts which don' contain a non-zero character to "0.00000000".

### DIFF
--- a/gui/ui/src/component/amount.rs
+++ b/gui/ui/src/component/amount.rs
@@ -72,11 +72,10 @@ fn render_amount<'a, T: 'a>(amount: String, size: u16) -> Row<'a, T> {
 
     let (before, after) = match split_at_first_non_zero(amount) {
         Some((b, a)) => (b, a),
-        None => (String::from(""), String::from("")),
+        None => (String::from("0.00 000 000"), String::from("")),
     };
 
     let row = Row::new()
-        .spacing(spacing)
         .push(text(before).size(size).style(color::GREY_3))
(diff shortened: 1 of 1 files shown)


gui/ui/src/component/amount.rs
```diff
@@ -92,9 +91,7 @@ fn render_amount<'a, T: 'a>(amount: String, size: u16) -> Row<'a, T> {
 fn render_unconfirmed_amount<'a, T: 'a>(amount: String, size: u16) -> Row<'a, T> {
     let spacing = if size > P1_SIZE { 10 } else { 5 };
 
-    let row = Row::new()
-        .spacing(spacing)
-        .push(text(amount).size(size).style(color::GREY_3));
+    let row = Row::new().push(text(amount).size(size).style(color::GREY_3));
```
edouardparis, line 94: This Row::new seems to be useless, please move text(amount) to the next Row::with_children block

aaroncarlucci: Good catch, thanks. I just made this change.

aaroncarlucci:
I adjusted the default value  again. Tested on a new, empty wallet to reproduce the issue and it looks ok for me locally. Thanks for your patience with this -- let me know if you see anything else.

edouardparis (approved):
It is better
ACK e79674972d668d823d582cb01923af8598cbe674

PR #1228, opened 2024-08-14
WIP: Upgrade rust-miniscript / rust-bitcoin to latest version

### LINKED ISSUE #1224: Upgrade bitcoin dependency to v0.32
We currently use v0.31.

Using a later version will enable us to use more recent BDK releases than currently used in #1222, for example.

darosior:
I need to go over the changes again, and this introduces a massive performance hit. From quick profiling i think it's from upstream changes in the miniscript policy compiler. 

Running `cargo test descriptors` before:
```
real    0m5.166s
user    0m10.630s
sys     0m0.075s
```
After:
```
real    4m16.438s
user    7m14.968s
sys     0m0.099s
```

Fixes #1224.

### DIFF
--- a/src/descriptors/analysis.rs
+++ b/src/descriptors/analysis.rs
@@ -2,11 +2,11 @@ use miniscript::{
     bitcoin::{
         self, bip32,
         hashes::{sha256, Hash},
-        secp256k1, Sequence,
+        secp256k1,
     },
     descriptor,
     policy::{Concrete as ConcretePolicy, Liftable, Semantic as SemanticPolicy},
-    ScriptContext,
+    RelLockTime, ScriptContext, Threshold,
 };
 
 use std::{
@@ -14,6 +14,7 @@ use std::{
     convert::TryFrom,
     error, fmt,
     str::FromStr,
+    sync,
 };
 
 #[derive(Debug)]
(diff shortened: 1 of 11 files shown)


src/descriptors/analysis.rs
```diff
@@ -593,15 +602,23 @@ impl LianaPolicy {
         // primary path with at least one key, and at least one timelocked recovery path with at
         // least one key.
         let subs = match policy {
-            SemanticPolicy::Threshold(1, subs) => Some(subs),
-            _ => None,
-        }
-        .ok_or(LianaPolicyError::IncompatibleDesc)?;
+            SemanticPolicy::Thresh(thresh) if thresh.is_or() && thresh.n() > 1 => {
```
pythcoiner, line 605: shouldn't we check thresh.k() == 1?
```suggestion
            SemanticPolicy::Thresh(thresh) if thresh.is_or() && thresh.n() > 1 && threshk() == 1 => {
```

darosior:
Yes. In general i don't think the code here is correct. I need to go through the descriptor core logic one more time.

pythcoiner:
also not from these PR, but [here](https://github.com/wizardsardine/liana/blob/969046c25ba66d2c2125a3545dcb7ff90bf1eab8/src/descriptors/analysis.rs#L248) if assert fail we will panic shouldn'nt we error instead?

src/descriptors/analysis.rs
jp1ac4: I think this is already checked by `thresh.is_or()`.

jp1ac4: I think it makes sense to keep this `thresh.n() > 1` condition here, but perhaps this is not strictly necessary as the previous version didn't have  `subs.len() > 1` and we already ensure further below that the primary path and recovery paths are not empty.

jp1ac4:
Replaced by #1555.

PR #1232, opened 2024-08-18
Simplify UI of Welcome screen

ktecho:
In the welcome screen, Liana has this:

image

Each option has a text that implies an action, and then a button with the action text "Select". The UI can be changed and simplified so it only reflects the two actions that can be chosen. The end result would be this one:

image

And when putting the mouse over the button:

image

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -74,45 +74,43 @@ pub fn welcome<'a>() -> Element<'a, Message> {
                                 .spacing(20)
                                 .push(
                                     Container::new(
-                                        Column::new()
-                                            .spacing(20)
-                                            .align_items(Alignment::Center)
-                                            .push(
-                                                image::create_new_wallet_icon()
-                                                    .width(Length::Fixed(100.0)),
-                                            )
-                                            .push(
-                                                p1_regular("Create a new wallet")
-                                                    .style(color::GREY_3),
-                                            )
-                                            .push(
-                                                button::secondary(None, "Select")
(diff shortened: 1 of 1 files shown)


edouardparis:
Thank you for the PR. Sadly we cannot use it as it is because we are trying to develop a UX/UI development process where design choices are discussed first with our product manager, designers and helpful users.

ktecho:
> Thank you for the PR. Sadly we cannot use it as it is because we are trying to have a UX/UI development process where design choices are discussed first with our product manager, designers and helpful users.

I understand, no problem.

What should I have done? I'm in Liana's Discord. Should I have suggested it first there? Thanks!

nondiremanuel:
> > Thank you for the PR. Sadly we cannot use it as it is because we are trying to have a UX/UI development process where design choices are discussed first with our product manager, designers and helpful users.
> 
> I understand, no problem.
> 
> What should I have done? I'm in Liana's Discord. Should I have suggested it first there? Thanks!

Thanks for the contribution and the interest. Yes, discussing it via discord could be a good way. In this specific case, for example, we are already designing a new layout for that screen, which would make the work on this PR short-living. 

A good way could also be to take some issues from the "Good first issue" basket, especially the ones related to the GUI: https://github.com/wizardsardine/liana/issues?q=is%3Aopen+is%3Aissue+label%3A%22Good+first+issue%22
Those can also be discussed directly in the each specific github issue.

ktecho:
Perfect, thanks. You can close the PR.

PR #1234, opened 2024-08-23
fix wrong datadir in app cache using remote backend

edouardparis:
Detected after Manuel failed to have in cache the correct the refresh token. It was not correctly stored after 1h in the settings file because of the wrong datadir.


edouardparis (commented):
Self-ACK 8610afeea635c0ffe91db76fb10ed46e72478d82

PR #1236, opened 2024-08-26
Display details of wallet policy in installer

### LINKED ISSUE #620: Display the policy nicely (like in installer) in the settings
Such a great UI to create the wallet. 

Why not use the same interface to visualize what are the different recovery fallbacks are after creation?

edouardparis:
20240827_14h53m19s_grim
20240827_14h56m31s_grim

close #620

### DIFF
--- a/gui/src/app/state/settings/wallet.rs
+++ b/gui/src/app/state/settings/wallet.rs
@@ -5,7 +5,10 @@ use std::sync::Arc;
 
 use iced::{Command, Subscription};
 
-use liana::miniscript::bitcoin::{bip32::Fingerprint, Network};
+use liana::{
+    descriptors::LianaDescriptor,
+    miniscript::bitcoin::{bip32::Fingerprint, Network},
+};
 
 use liana_ui::{
     component::{form, modal},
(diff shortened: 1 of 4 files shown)


nondiremanuel:
Thanks for implementing it! A couple of quick comments looking at the screenshots:
- I think we miss `'s` in the primary path: "can always spend this wallet**'s** funds ..."
- Why the first recovery path is "Recovery path # 0?
- In case of single sig, I think the "out of 1" part would be better to be omitted. But it's a minor thing and if it complicates the logic we can keep it as it is

Regarding the wording for recovery paths which was being discussed in the original issue, I agree with @jp1ac4's idea of referring to coins, but I believe that the concept of "older" for it can be tricky to understand for the user. I also agree on the fact that we should use a sentence as short as possible. I would prefer a way to more explicitly say what we mean by older, like "can recover coins not spent/refreshed for more than X blocks ..."

edouardparis:
> can recover coins not spent/refreshed for more than X blocks ...

What do you think of "untouched" ?

jp1ac4:
I agree that "untouched" could be ambiguous. I think "inactive" or perhaps "dormant" would be clearer.

edouardparis:
> "inactive" or perhaps "dormant" would be clearer.

ACK for inactive

edouardparis (commented):
Self-ACK 59e7131aeef4478933c6923155ed59535e218c96

PR #1237, opened 2024-08-28
Change remote backend server url

edouardparis:
Remote backend has the auth service credentials.
We rely now uniquely on the following urls:
* https://api.signet.lianalite.com for signet
* https://api.lianalite.com for mainnet


edouardparis (commented):
Self-ACK eef55b41c97146476234bbd445e23d9f8f98cb20

Issue #1223, opened 2024-08-08
[GUI] Add Electrum backend option to Installer & update Settings


jp1ac4:
Following #1222, we will need to add an Electrum option in the Installer and update the Settings page accordingly.

jp1ac4:
For the installer, we currently have this page:

![image](https://github.com/user-attachments/assets/9ddce050-3eea-4812-a374-907f0abf2c7e)

We could either:
1. add the Electrum option to this same page (perhaps one row per option instead of columns), or
2. add a preceding page "Select Bitcoin backend type" with options "Bitcoin node" (which then goes to above page) & "Electrum" (a new page to enter Electrum details and check connection).

Perhaps we should opt for fewer pages/clicks, in which case (1) would be better.

edouardparis:
I like the idea of having the 3 options in columns, each in a card with a title that needs to be clicked on to collapse and show the text and the button next.

nondiremanuel:
Intuitively (but probably I'm missing something), I would expect to have a three step choice:

1. Do you want to use the remote backend (Liana Connect) or your own node? (this is the part of @edouardparis' PR)
2. If your node, do you already have one or do you want us to install a pruned one? (basically the same choice we have now changing the wording to mention that there is also Electrum in the former option)
3. If you have one, local bitcoind or electrum?

The choice branching would be something like:

- Liana Connect [Default]
- My own Node
   - Connect to an existing Node
      - Connect to my local Bitcoin Core
      - Connect to Electrum
   - Download a pruned liana-managed one

In this case, electrum would be an option in the following page in which you put your IP and so on:
![image](https://github.com/user-attachments/assets/1f2b9f50-3f87-447d-b271-8ed227ca7305)

The reason for that is that people could be confused by the fact of having two options to link to their own node at that stage.

What am I missing?

jp1ac4:
Thanks for the suggestions. It makes sense to consider how these options and the flow will link with the Liana Connect option.

>1. Do you want to use the remote backend (Liana Connect) or your own node? (this is the part of @edouardparis' PR)

I tend to think of "node" as Bitcoin Core, so I think we could change this to "or your own backend (Bitcoin node or Electrum server)".

>2. If your node, do you already have one or do you want us to install a pruned one? (basically the same choice we have now changing the wording to mention that there is also Electrum in the former option)

Changing the existing wording would be the simplest thing, but having the three cards would also be nice.

> In this case, electrum would be an option in the following page in which you put your IP and so on

[...]

PR #1222, opened 2024-08-08
Add Electrum backend

jp1ac4:
This is to add the Electrum backend as part of https://github.com/wizardsardine/liana/issues/56#issuecomment-2183063784.

This requires the database to be running version 5 following #1180. The migration from a previous DB version must be done using bitcoind. Thereafter, the daemon can be run using an Electrum backend by replacing the `[bitcoind_config]` section in the daemon.toml config file with:
```
[electrum_config]
addr = "127.0.0.1:50001" # adjust IP:port as required
```
Remaining tasks:

- [x] Include ancestors and descendants when getting a transaction's `MempoolEntry`
- [ ] Additional sanity checks, e.g. check Electrum version
- [x] Check if logic regarding ongoing rescan needs to be adjusted
- [x] Add Electrum backend to functional tests
- [x] Add Electrum backend to CI pipline functional tests
- [ ] Add unit tests
- [x] Update README & other docs

Issue #1082, opened 2024-04-20
Can't connect LIana to an Umbrel node


piratebiscuit:
Hey guys,

I'm trying to connect Liana to an Umbrel node I have lying around, I get an unexpected result...
Altough the "Check connection" succeeds, I still get an error saying to "Enter a correct address".

Did the check connection actually fail or did it succeed and Liana simply fails a check because of how the umbrel local address is written?

<img width="968" alt="Screenshot 2024-04-19 at 23 40 14" src="https://github.com/wizardsardine/liana/assets/154634278/547b9068-3c50-4b6f-86be-d53a860aded2">

Thanks

pythcoiner:
humm can you try w/ the direct IP address instead? I think it's oubout the text input validation in our side

pythcoiner:
I guess [here](https://github.com/wizardsardine/liana/blob/f9ea8ec0b6edc533159601d828ffd25134d324ae/gui/src/installer/step/bitcoind.rs#L447) `SocketAddr::from_str()` accept only ip addresses

piratebiscuit:
> Is the `next` button clickable?
> 
> if not can you try w/ the direct IP address instead?
> 
> I think it's oubout the text input validation in Liana side

The Next button is not clickable in this case.

jp1ac4:
> Did the check connection actually fail or did it succeed

I think the check succeeded. The connection check here is done using just the string without trying to convert it to an IP address.

It might be possible to use something like https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html to convert the hostname to an IP address, but the IP address would still be saved in the Liana config file and so you might still have trouble connecting the next time. Please let us know if you're able to connect using just the IP address and we can consider how to proceed.

pythcoiner:
> > Did the check connection actually fail or did it succeed
> 
> I think the check succeeded. The connection check here is done using just the string without trying to convert it to an IP address.
> 
> It might be possible to use something like https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html to convert the hostname to an IP address, but the IP address would still be saved in the Liana config file and so you might still have trouble connecting the next time. Please let us know if you're able to connect using just the IP address and we can consider how to proceed.

i'm wondering if umbrel is using local url because the node ip can change?

piratebiscuit:
For some reason, when I use the IP address of the umbrel node, it fails...

[...]

PR #1241, opened 2024-08-28
[GUI] Installer - Add Electrum node option

jp1ac4:
This is for #1223.

For now, it's possible to edit the node's settings but not to change node type.

Remaining tasks:

- [x] Revert Cargo.toml once #1222 is merged.
- [x] Update wording as per https://github.com/wizardsardine/liana/issues/1223#issuecomment-2286483134.


jp1ac4:
I made a further change to the behaviour while checking the connection. It's safer to prevent the user editing config values until the connection check completes to ensure the values don't change in the meanwhile, which could mean the connection status applies to the wrong values.

I think this was already an issue with the current installer but I guess connection checks are usually quicker with a local bitcoind and so it was less likely to cause problems.

jp1ac4:
I updated the liana dependency commit.

nondiremanuel:
For the first page, I was proposing to take the opportunity to make the text simpler and more understandable (at least in my view):

> - I already have a node
Select this option if you already have a bitcoin node running locally or remotely. Liana will connect to it.
> - I want Liana to automatically install a Bitcoin node on my device
Liana will install a pruned (i.e. "compressed") node on your computer. You won't need to do anything except having some disk space available (~25GB required) and waiting for the initial synchronization with the network (it can take some days according to your internet connection speed).
[the two parenthesis with the estimation space and time possibly should be shown only if the network chosen is mainnet]

If there is no consensus with this, we can also leave it as you did (i.e. adding the Electrum mentions) and discuss the general wording for another time.

The second page was exactly what I had in mind. Thanks!

jp1ac4:
I've updated the wording as suggested with some minor changes:
- I felt "truncated" might be more accurate than "compressed", but probably pruned already conveys this so didn't include that part.
- I added "on mainnet" after ~25 GB. We don't currently have the network selection available in the view. It should be simple to add as a follow-up commit if required and only display the additional info for mainnet.
- I used "depending on" instead of "according to" as I think it's more common this way but either should be fine.

image

I removed "full" from the title:

image

pythcoiner:
30 GB looks more correct (signet chain included but it's only 2.2GB):
```shell
~/.liana/bitcoind ❯ sudo du datadir -sh                                                                    
29G     datadir
```

pythcoiner:
image
can we add `127.0.0.1:50001` as placeholder here? 
even if it's not likely the electrum server is on localhost, it indicate the user it should specify the port, and indicate 50001 is the default port.

pythcoiner:
i'm wondering about one thing here, did the address still only accept ip addresses? 
i think the default config on umbrel is to give access via `umbrel.local` instead of an ip as (i guess) almost user do not configure a static ip for their node.

gui/src/installer/view.rs
```diff
@@ -1088,12 +1091,107 @@ pub fn help_backup<'a>() -> Element<'a, Message> {
     text(prompt::BACKUP_DESCRIPTOR_HELP).small().into()
 }
 
-pub fn define_bitcoin<'a>(
+pub fn define_bitcoin_node<'a>(
     progress: (usize, usize),
+    available_node_types: impl Iterator<Item = NodeType>,
+    selected_node_type: &NodeType,
```
pythcoiner, line 1097: `NodeType` implement `Copy`
```suggestion
    selected_node_type: NodeType,
```

gui/src/installer/step/node/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+use liana_ui::widget::Element;
+
+#[derive(Clone)]
+pub enum NodeDefinition {
+    Bitcoind(DefineBitcoind),
+    Electrum(DefineElectrum),
+}
+
+impl NodeDefinition {
+    fn new(node_type: &NodeType) -> Self {
```
pythcoiner, line 28: `NodeType` implement `Copy`
```suggestion
    fn new(node_type: NodeType) -> Self {
```

gui/src/installer/step/node/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+pub struct Node {
+    definition: NodeDefinition,
+    is_running: Option<Result<(), Error>>,
+    waiting_for_ping_result: bool,
+}
+
+impl Node {
+    fn new(node_type: &NodeType) -> Self {
```
pythcoiner, line 92: `NodeType` implement `Copy`
```suggestion
    fn new(node_type: NodeType) -> Self {
```

gui/src/installer/view.rs
```diff
@@ -1088,12 +1091,107 @@ pub fn help_backup<'a>() -> Element<'a, Message> {
...
+                        Row::new()
+                            .spacing(10)
+                            .align_items(Alignment::Center)
+                            .push(icon::circle_cross_icon().style(color::RED))
+                            .push(text("Connection failed").style(color::RED)),
+                    )
+                }
+            })
+        } else {
+            Some(Container::new(Space::with_height(Length::Fixed(25.0))))
```
pythcoiner, line 1157: nit (maybe depending on renderer/platform)
```suggestion
            Some(Container::new(Space::with_height(Length::Fixed(21.0))))
```

pythcoiner:
> i'm wondering about one thing here, did the address still only accept ip addresses? i think the default config on umbrel is to give access via `umbrel.local` instead of an ip as (i guess) almost user do not configure a static ip for their node. cf [this comment](https://github.com/wizardsardine/liana/issues/1082#issuecomment-2071211975)

tested on linux adding `192.168.1.21 electrum.local` to /etc/hosts:
image

gui/src/installer/step/node/electrum.rs
```diff
@@ -0,0 +1,73 @@
...
+        Self {
+            address: form::Value::default(),
+        }
+    }
+
+    pub fn can_try_ping(&self) -> bool {
+        !self.address.value.is_empty() && self.address.valid
+    }
+
+    pub fn load_context(&mut self, _ctx: &Context) {}
```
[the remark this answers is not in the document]
jp1ac4: I think it can be removed. It's a leftover from an earlier version where I defined a trait that each node type had to implement, but now it's not required.

gui/src/installer/step/node/electrum.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    pub fn update(&mut self, message: message::DefineNode) -> Command<Message> {
+        if let message::DefineNode::DefineElectrum(msg) = message {
+            match msg {
+                message::DefineElectrum::ConfigFieldEdited(field, value) => match field {
+                    ConfigField::Address => {
+                        self.address.value.clone_from(&value);
+                        self.address.valid = !self.address.value.is_empty();
```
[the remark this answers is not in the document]
jp1ac4: Yes I think that would make sense, will add it.

jp1ac4:
I've made the suggested changes.

I now need to rebase on master as there's a conflict.

gui/src/installer/step/node/electrum.rs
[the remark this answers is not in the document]
jp1ac4: (for future reference: I removed the trait as I was having trouble including the `ping()` method in an object-safe way. I only wanted the ping method to return the `Result` and not the `Command` to handle the `PingResult`)

jp1ac4:
Rebased on master.

jp1ac4:
Sorry, I made a further change to the Electrum address validation.

pythcoiner:
testing w/ several remote servers:
image
image
image

seems ok w/ both ssl/tcp

pythcoiner:
Here we should rename "Bitcoin Core" into "Electrum server":
image

pythcoiner:
Here we should also remove the rescan feature if electrum, but we should keep the block height, i often use it for know if liana is sync w/ the backend:

image

pythcoiner:
I'm actually trying electrum (on an other machine on my local network) on regtest w/ 550 coins:
image

for every sync (so every 30 seconds) electrs needs around 10 seconds to answer the gui request.

if i click on `Receive` menu during this 10 seconds i got not address util all the requests done:

image

i wonder about 2 things:
- do we make some call to te electrum server for generate address? or it's the iced "event loop" that became overloaded?

- why we still make so much requests to electrum server after we are sync?

pythcoiner:
trying to connect a mainnet wallet on regtest electrum fail as expected:
image
maybe we can try to detect network by genesis hash and display a more explicit message to the user?
(i don't think this should be tackle in this PR but rather in a follow up)

jp1ac4:
Updated liana dependency commit hash.

jp1ac4:
Added a couple of commits to update the Settings and allow to view & edit Electrum settings.

It's NOT possible to switch from one node type to another. This will require more changes and we'll probably need to leave that for a follow-up.

I've renamed the Settings page to "Node":

image

We can view the Electrum settings (and some chain info):

image

We can edit the address:

image

jp1ac4:
TODO: Update comment in installer about rescan after importing wallet.

jp1ac4:
> TODO: Update comment in installer about rescan after importing wallet.

Updated the wording.

jp1ac4:
Updated liana dependency commit hash.

pythcoiner:
image
we do not ping the electrum server before let user save settings, so if the address is wrong `ElectrumSettings.processing` is still true and we cannot edit the address anymore

pythcoiner:
image
if we conect to a wrong network electrum server we should have a clearer message

jp1ac4:
Connecting to the node only after saving the changes is also done for bitcoind. Ideally, we would ping the server and check the network before saving, which could avoid that error about hash mismatch. Or perhaps I can add something to display that error differently in the GUI. Probably in a follow-up.

For Electrum, it's using a timeout of 180 seconds but I checked that it eventually gives a daemon error. I'll see if I can use a shorter timeout in #1222.

jp1ac4:
> For Electrum, it's using a timeout of 180 seconds but I checked that it eventually gives a daemon error. I'll see if I can use a shorter timeout in #1222.

Hmm don't think I can change it just for the initial connection. I think better to add the pre-save ping on the GUI settings page for both bitcoind and Electrum, reusing what's already done in the installer.

jp1ac4:
> The loader still refers to Bitcoin Core syncing. We can change that depending on node type in a follow-up.

Created https://github.com/wizardsardine/liana/issues/1256.

pythcoiner:
> > For Electrum, it's using a timeout of 180 seconds but I checked that it eventually gives a daemon error. I'll see if I can use a shorter timeout in #1222.
> 
> Hmm don't think I can change it just for the initial connection. I think better to add the pre-save ping on the GUI settings page for both bitcoind and Electrum, reusing what's already done in the installer.

maybe we can just disable edit for now (or keep the bug) and doing it in a follow up

jp1ac4:
> maybe we can just disable edit for now (or keep the bug) and doing it in a follow up

I'll create a follow-up issue. I think this is the same as current behaviour for bitcoind, just that Electrum takes longer to fail.

jp1ac4:
Rebased on master and updated liana dependency commit hash.

TODO: incorporate daemon changes related to rescan logic in #1222.

pythcoiner:
Maybe we can add (in a follow up) the name/type of the electrum server (by sending a `server.banner` request

jp1ac4:
Updated liana dependency commit hash.
Changed `DaemonControl` to mutable following changes to `start_rescan` in #1222.
Allow to start rescan also for Electrum.

jp1ac4:
Added a commit to enable changing from one node type to another in the settings.

Note that in its current form, it'll be possible for someone who's using managed bitcoind to switch to Electrum in the settings. In this case, the GUI will not try to start the managed bitcoind the next time they launch the GUI. I could also prevent such a user from switching node type (indeed, they cannot edit the bitcoind settings currently).

I also made a fix to a previous commit so that we take into account whether a rescan is in progress in order to determine whether the Electrum settings can be edited.

pythcoiner:
ACK 819eb920c0210b9d95689f40cfbeba65c03280af

PR #1245, opened 2024-08-30
Merge launcher and first installer step

edouardparis:
When the gui launches, it checks if there is
a wallet in the network datadirs with bitcoin mainnet in priority, then other networks as fallbacks. If it is the case, it tries to display information about the wallet by reading the settings.json file.
If no wallet was found, then the gui displays the choice to either create a wallet or add it.

20240830_10h55m36s_grim
20240830_10h55m53s_grim


jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/pull/1245/commits/c99df6ccfa88a75ab31ae1edb2b17ef771f50b42.

Tested with both local and remote backends.

PR #1246, opened 2024-08-30
fix hard coded network signet for remote backend

edouardparis:
remote backend supports now also mainnet, we have to load downstream the chosen network.


edouardparis (commented):
Self-ACK a331f9fd706c34ba8ab639ca1c7b16a2a80a4fc3

PR #1257, opened 2024-09-04
qa: pipe stderr to stdout, not PIPE

darosior:
It turns out that setting both to PIPE would result in stderr not being read for some reason. So follow
https://docs.python.org/3/library/subprocess.html and set stdout to PIPE and stderr to STDOUT.

Tested with a process (electrs) which logs on stderr.


darosior:
ACK 26ade7ebdd3b35667e07ba34986c7e25dd857f89 -- tested by running the functional tests in parallel for #1222 and could correctly get the logs to debug the issue.

PR #1263, opened 2024-09-04
Use different msg for reloading or extending history tx list

### LINKED ISSUE #1260: After reorg `Last payments` are not removed
image

but they disapear on after Liana restart

edouardparis:
After reorg last payments were not removed as the reloading message was handle as an extension of the list.

close #1260


pythcoiner:
ACK tested w/ bitcoind backend

Issue #1276, opened 2024-09-06
electrum: Check connection before creating client


jp1ac4:
The timeout is currently set at 180 seconds, but we should first check connectivity in the client constructor.

This is the same as done for bitcoind.

h/t @darosior

PR #1277, opened 2024-09-06
electrum: check connectivity before creating client

jp1ac4:
This is to resolve https://github.com/wizardsardine/liana/issues/1276.

Tested that this change stops GUI freezing in https://github.com/wizardsardine/liana/pull/1241 when using invalid address in Electrum settings.


darosior (commented):
utACK 0f48db7a88b1dcaee68190a6a6ee5e73b5630f01

Issue #1240, opened 2024-08-28
QA - Change requests (wording, UX) on the Liana Connect flow


nondiremanuel:
As agreed, I report here some of the change requests to wording and similar related to the liana backend integration:

- [x] 1. The text in the choice of backend step should be adjusted. Now it looks like this:
![image](https://github.com/user-attachments/assets/174bf86d-280a-4b0b-8adf-6a2b6dc21393)
We should insert something like:
- **Your own node**
"Use your already existing Bitcoin node or automatically install one. The Liana wallet will not connect to any external server."
Info/warning: This is the most private option, but the data is locally stored on this computer, only. You must perform your own backups, and share the descriptor with other people you want to be able to access the wallet"
[Select]
- **Liana Connect**
"Use a hosted service to talk to the Bitcoin network and store data. Wizardsardine runs the infrastructure, keeps a backup of your wallet descriptor, and allow for synchronization between multiple computers and participants."
Info/warning: This is a safer option for users who want Wizardsardine to keep a backup of your descriptor. Wizardsardine will be able to see the information of your wallet, associated to an email address. Privacy focused users should run their own infrastructure instead
[Select]
- [x] If feasible, the email and token insertion should be in a following step or in a dedicated popup if the "Liana Connect" option is selected. The text should be something like: "Enter the email address you want to associate to this wallet:"

- [x] 2. The token email title is not correct:
![image](https://github.com/user-attachments/assets/fc108571-b3c7-49b0-ae4b-1958698b99d0)
It should be "Login to Liana Connect"
- [ ] 3. The token email body should be edited to: 
"Connect to the Wizardsardine infrastructure.
This is your one time authentication token: XXXXXX"
The size of the authentication token should be increased.
- [x] 4. When the user enters the token, we should go to the next step without needing the user to click on "Continue"
- [x] 5. In the "Add an existing wallet", right now we have the following options:
![image](https://github.com/user-attachments/assets/82151c03-8766-4a2b-a7df-8e69574ef7d3)
Wording should be reviewed to something like:
- **Load a previously used wallet**
(list the wallets) -> is it possible to show this option only if at least one wallet is present?
- **Load a shared wallet**
Type the invitation token you received by email
Paste token:
- **"Load from a descriptor"**
Use only if the previous don't apply
- [x] 6. The invitation email needs more detail on how to participate in a wallet. Right now it looks like this:
[...]

PR #1278, opened 2024-09-06
installer: Go directly to next step once login

edouardparis:
Once the user is connected we redirect him to
the next step in the install process, but
we keep the Connected state in case the user
want to edit email after clicking on Previous

As discussed in #1240


edouardparis (commented):
Self-ACK cc02ac043b647b22b0533dec3f99cbcf62e7ee8f

PR #1281, opened 2024-09-06
Order history txs in home state

edouardparis:


edouardparis (commented):
this is to fix  a regression introduced by #1263
Self-ACK 887b71367bebe277740443ff4e2d0f1470d2c0ab

PR #1282, opened 2024-09-06
electrum: ping server before returning client

jp1ac4:
Ping server as an extra connectivity check.

### DIFF
--- a/src/bitcoin/electrum/client.rs
+++ b/src/bitcoin/electrum/client.rs
@@ -57,7 +57,9 @@ impl Client {
     pub fn new(electrum_config: &config::ElectrumConfig) -> Result<Self, Error> {
         // First use a dummy config to check connectivity (no retries, short timeout).
         let dummy_config = Config::builder().retry(0).timeout(Some(3)).build();
+        // Try to ping the server.
         bdk_electrum::electrum_client::Client::from_config(&electrum_config.addr, dummy_config)
+            .and_then(|dummy_client| dummy_client.ping())
             .map_err(Error::Server)?;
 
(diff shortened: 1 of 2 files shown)


jp1ac4:
I haven't been able to reproduce the test failures on my local machine. Still investigating.

src/bitcoin/electrum/client.rs
```diff
@@ -57,7 +57,9 @@ impl Client {
     pub fn new(electrum_config: &config::ElectrumConfig) -> Result<Self, Error> {
         // First use a dummy config to check connectivity (no retries, short timeout).
         let dummy_config = Config::builder().retry(0).timeout(Some(3)).build();
+        // Try to ping the server.
         bdk_electrum::electrum_client::Client::from_config(&electrum_config.addr, dummy_config)
+            .and_then(|dummy_client| dummy_client.raw_call("server.ping", []))
```
darosior, line 62: What if you just used the `ping` method?
```suggestion
            .and_then(|dummy_client| dummy_client.ping())
```

jp1ac4:
> `lianad` logs for CI failure:

Thanks :pray: 

> Please don't re-push just now, i'm trying to debug on Cirrus. Also, we should address the root cause of the issue, not just throw more retries until it passes CI.

I agree. We probably need to look for something else in the logs when starting electrs to consider it ready to receive the ping request.

darosior:
We could let the more robust solutions as a followup.

darosior:
utACK 537ebb2fb3301e82a297a5c8c4c5043d8d0938e4

Issue #1280, opened 2024-09-06
Electrum sync/rescan crash


darosior:
```
/rust/liana/gui @819eb920 *37 !1 ?2 ❯ cargo run --release                                         11m 46s 12:16:10
    Finished `release` profile [optimized] target(s) in 0.24s
     Running `target/release/liana-gui`
  2024-09-06T10:35:00.542782Z  INFO electrum_client::raw_client:205: Trying to connect to 192.168.1.21:50003 (attempt 1/1) with timeout 3s

  2024-09-06T10:35:01.088328Z  INFO liana:452: Created a new data directory at '/home/pyth/.liana/regtest'

  2024-09-06T10:35:01.256325Z  INFO liana::database::sqlite:156: Created a fresh database at /home/pyth/.liana/regtest/lianad.sqlite3.

  2024-09-06T10:35:01.256346Z  INFO liana::database::sqlite:162: Checking if the database needs upgrading.

  2024-09-06T10:35:01.260472Z  INFO liana:252: Database initialized and checked.

  2024-09-06T10:35:01.260488Z  INFO electrum_client::raw_client:205: Trying to connect to 192.168.1.21:50003 (attempt 1/1) with timeout 180s

  2024-09-06T10:35:01.490636Z  INFO liana:516: Bitcoin poller started.

  2024-09-06T10:35:01.490684Z  INFO liana::bitcoin::poller:87: Bitcoin poller was told to shut down.

  2024-09-06T10:35:01.490697Z  INFO liana:518: Bitcoin poller stopped.

  2024-09-06T10:35:01.490975Z  INFO liana_gui::installer:357: daemon checked

  2024-09-06T10:35:01.491216Z  INFO liana_gui::installer:375: Daemon configuration file created

  2024-09-06T10:35:01.491373Z  INFO liana_gui::installer:420: Gui configuration file created

  2024-09-06T10:35:01.491650Z  INFO liana_gui::installer:432: Settings file created

  2024-09-06T10:35:01.533628Z  INFO liana_gui::daemon::client:49: getinfo

  2024-09-06T10:35:01.533765Z ERROR liana_gui::daemon::client:51: method getinfo failed: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })

  2024-09-06T10:35:01.554174Z  INFO liana::database::sqlite:162: Checking if the database needs upgrading.

  2024-09-06T10:35:01.554794Z  INFO liana::database::sqlite::utils:313: Database is up to date.

[...]

Issue #1270, opened 2024-09-05
electrum: fix flaky reorg tests


jp1ac4:
See https://github.com/wizardsardine/liana/pull/1222#issuecomment-2331947433.

PR #1285, opened 2024-09-06
electrum: changes to common ancestor logic for reorgs

jp1ac4:
There might be an issue if a reorg happens during a rescan.

This is an idea of a change we could make.


jp1ac4:
For now, just treat any full scan as a reorg with common ancestor genesis block if the height was previously positive.

jp1ac4:
This is possibly related to https://github.com/wizardsardine/liana/issues/1280.

jp1ac4:
The second commit is now part of #1299 instead.

jp1ac4:
> The second commit is now part of #1299 instead.

I've added a third commit here. It seems the tests fail if any one of these commits is missing, so I'll keep them all as part of this PR.

src/bitcoin/electrum/mod.rs
```diff
@@ -204,24 +204,22 @@ impl Electrum {
...
-                            None
-                        } else {
-                            log::info!("Block chain reorganization detected.");
-                            // We can assume height is positive as genesis block will not have changed.
-                            Some(self.bdk_wallet.find_block_before_height(height))
-                        };
-                    }
-                    Some((_, None)) => continue,
-                    None => break None,
+            if let Some((height, _)) = changes_iter.next() {
```
darosior, line 207: Makes sense. Anything inserted or removed at or before our current height should make us reorg, no need for this loop.

nondiremanuel:
As per the meeting, it could potentially fix both #1270 and #1280. If not we need to understand what's the origin of those issues.

src/bitcoin/electrum/mod.rs
```diff
@@ -198,25 +200,42 @@ impl Electrum {
...
+                    }
+                    Some((_, None)) => continue,
+                    None => break None,
                 }
-                Some((_, None)) => continue,
-                None => break None,
+            }
+        } else {
+            // In case of a full scan, just replace our local chain with the new one.
+            self.bdk_wallet.replace_chain_from_tip(chain_update);
```
darosior, line 226: Why is this necessary? `apply_connected_chain_update` should have exactly the same effect if the only block in common is the genesis.

src/bitcoin/electrum/mod.rs
```diff
@@ -198,25 +200,42 @@ impl Electrum {
...
+            // In case our chain tip was previously positive, treat this full scan
+            // as a reorg with common ancestor being the genesis block.
+            if local_chain_tip.height() > 0 {
+                Some(tip_from_block_id(
+                    self.local_chain()
+                        .get(0)
+                        .expect("contains genesis")
+                        .block_id(),
+                ))
```
darosior, line 235: I don't understand the rationale for this. What should we unconditionally treat rescans as reorgs?

darosior (commented):
The last two commits look good to me. I have trouble understanding why the first one should be necessary.

src/bitcoin/electrum/wallet.rs
```diff
@@ -295,10 +295,10 @@ impl BdkWallet {
         })
     }
 
-    /// Find the first block in the local chain whose height is less than or equal to this.
-    pub fn find_block_at_or_before_height(&self, height: u32) -> BlockChainTip {
+    /// Find the highest block in the local chain whose height is below this.
+    pub fn find_block_before_height(&self, height: u32) -> BlockChainTip {
         for cp in self.local_chain.iter_checkpoints() {
-            if cp.height() <= height {
+            if cp.height() < height {
```
darosior, line 301: As such this may trigger the `unreachable!()` below in case the passed `height` is `0`. We need to special case the genesis block or return `None` (and `unwrap_or(genesis_block)` in the caller). I think the latter is cleaner.

src/bitcoin/electrum/mod.rs
jp1ac4: That may be true, but my thinking is as follows: When we do the full scan, it's from the genesis block, and so the chain update we receive is with respect to the genesis block only and has no knowledge of our local chain tip. If there's a reorg, I'm worried that the chain update with respect to the genesis block may not correctly update our local chain.

src/bitcoin/electrum/mod.rs
jp1ac4: This was mainly to keep it simple and avoid adding logic to check for a reorg and get common ancestor (assuming we can't rely on the chain update for that). Given that we'll roll back the DB to genesis anyway as part of rescan logic, I felt the safest thing was just to do that also here.

src/bitcoin/electrum/mod.rs
darosior: You mean if there was a reorg, then we rescan, and then arrive to this line, right? So a situation like:
- Our local chain is: `G` <- `A` <- `B` <- `C`
- The update is: `G` <- `D` <- `E` <- `F`

Then i see no reason why `apply_connected_chain_update` would not correctly update our local chain? It's exactly what it's designed to do.

darosior:
Could you PR the last two commits separately while we discuss the first commit? So we can already get those in.

jp1ac4:
> Could you PR the last two commits separately while we discuss the first commit? So we can already get those in.

Those two commits are in https://github.com/wizardsardine/liana/pull/1299.

jp1ac4:
Closing this as existing chain update logic for a full scan should work also in case of a reorg.

jp1ac4:
Re-opening as it seems to prevent #1280.

jp1ac4:
Rebased on master.

src/bitcoin/poller/looper.rs
```diff
@@ -275,8 +275,16 @@ fn updates(
         Ok(Some(reorg_common_ancestor)) => {
             // The block chain was reorganized. Rollback our state down to the common ancestor
             // between our former chain and the new one, then restart fresh.
-            db_conn.rollback_tip(&reorg_common_ancestor);
-            log::info!("Tip was rolled back to '{}'.", &reorg_common_ancestor);
+            if reorg_common_ancestor.height < current_tip.height {
```
jp1ac4, line 278: We should also roll back DB if height is same but hash has changed.

jp1ac4:
Closing again in favour of #1315 and #1316.

PR #1296, opened 2024-09-08
gui(intsaller): ping electrum backend both with tcp & ssl

pythcoiner:
This PR add a double ping (both tcp & ssl) when defining an electrum server address.
After ping if the prefix is not ssl we display a warning to the user:
image

### DIFF
--- a/contrib/loop_test.sh
+++ b/contrib/loop_test.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+iteration=0
+
+while true; do
+    iteration=$((iteration+1))
+
+    pytest -vvvv --log-cli-level=DEBUG $1
+
+    if [ $? -ne 0 ]; then
+        echo "Test failed on iteration $iteration!"
+        break
+    fi
+done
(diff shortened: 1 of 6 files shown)


nondiremanuel:
I won't comment on the technical solution, but on the UI side I would use yellow/orange (the usual warning color) for warnings. I would possibly make it a little bigger

pythcoiner:
- the size is the usual size we use for warning/error on not valid field, so i think we should keep this one
- will change the color

jp1ac4:
Do we need to warn about SSL if people are using on their local network?

jp1ac4:
Wouldn't it be simpler to have a "Use SSL" checkbox?

pythcoiner:
the advantage of ping both is i guess almost the time our user will not know what is ssl, so w/ this solution they don't have to care about.

pythcoiner:
> Do we need to warn about SSL if people are using on their local network?

i think it's not easy to be sure wether we are a local network or not

pythcoiner:
i close this for now, as we are not sure it's the way we want go

PR #1299, opened 2024-09-08
electrum: common ancestor is before first changed block

jp1ac4:
The common ancestor is the highest block that is strictly lower than the first change.

When finding the point of change, we can take the first height with any change (including an invalidated block) rather than the first height with a new block, which avoids a loop. The height found in both cases (loop vs no loop) should be almost always the same as we apply the update to our local chain before finding the common ancestor, but perhaps in a test if there's a reorg down to 0 and the chain update only contains invalidated blocks and no new blocks, then checking for the first invalidated block would be better as there would be no new blocks to find and the reorg would be undetected.

### DIFF
--- a/src/bitcoin/electrum/mod.rs
+++ b/src/bitcoin/electrum/mod.rs
@@ -201,28 +201,26 @@ impl Electrum {
         let changeset = self.bdk_wallet.apply_connected_chain_update(chain_update);
 
         let mut changes_iter = changeset.into_iter();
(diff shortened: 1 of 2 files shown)


jp1ac4:
Tests fail here. I think we need https://github.com/wizardsardine/liana/pull/1285 instead.

darosior (commented):
ACK d551aeeeacf5356a0ba25f4929230dd099b01c32

PR #1303, opened 2024-09-09
backend settings

### LINKED ISSUE #1305: Issues with switching backend in settings
Recently (https://github.com/wizardsardine/liana/pull/1241#issuecomment-2330778018) we introduced the possibility to switch from one node type to the other in the settings.
This is creating some issues, such the one signaled by @kloaec:

[Screencast from 2024-09-06 12-50-56.webm](https://github.com/user-attachments/assets/9e258494-e1dc-4748-88ef-09dad70243bf)

The idea would be to not give this possibility in v7 but addressing such issues in a followup

### LINKED ISSUE #1306: Issues with inserting a wrong IP address in Electrum settings
Reported by @kloaec:

[Screencast from 2024-09-06 12-50-56.webm](https://github.com/user-attachments/assets/9e258494-e1dc-4748-88ef-09dad70243bf)

pythcoiner:
22ed096 fixes #1306 
2bd22b4 fixes #1305 
fd50913 in order to make clippy happy

### DIFF
--- a/gui/src/installer/step/node/electrum.rs
+++ b/gui/src/installer/step/node/electrum.rs
@@ -14,16 +14,14 @@ use crate::{
     node::electrum::ConfigField,
 };
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct DefineElectrum {
     address: form::Value<String>,
 }
 
 impl DefineElectrum {
     pub fn new() -> Self {
-        Self {
-            address: form::Value::default(),
-        }
+        Self::default()
     }
 
     pub fn can_try_ping(&self) -> bool {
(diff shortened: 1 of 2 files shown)


gui/src/installer/step/node/electrum.rs
```diff
@@ -84,3 +69,22 @@ impl DefineElectrum {
         Ok(())
     }
 }
+
+pub fn is_electrum_address_valid(value: &str) -> bool {
```
jp1ac4, line 71: I think it would be better to put this function in src/node/electrum.rs if it will be shared by both the app and the installer.

pythcoiner:
comment adressed

jp1ac4:
Are we sure we want to disable switching node type? Was the issue reported specifically due to switching or could it happen in general when editing the Electrum address?

darosior:
It can happen in general. The proper patch would be to handle the error and keep going with whatever is the current config if the modification fails, but it was suggested we could just disable the switch as a short term workaround for v7. If this isn't a workaround at all then maybe we should just correctly handle the error.

jp1ac4 (approved):
ACK 6349df95d3a690ac35be6dc18e5c6e0843d6c940.

Tested for both bitcoind & electrum. Could edit settings for same node type only.

Issue #1292, opened 2024-09-07
Don't overload the Electrum server


darosior:
If we could reduce the lookahead at least (and possibly the gap limit?) it would be nice.

pythcoiner:
i think it's the first step, but finally we should do like sparrow: subscriptions instead polling

darosior:
I think we can just lower the lookahead to something like 20. So that's 30 + 20 = 50, times 2 for change. So a hundred addresses queried per poll if i'm not mistaken? Right now it's 230 times 2.

nondiremanuel:
It was discussed that for v7 we need to understand what happens if a gap limit is hit and what could be the workaround.

jp1ac4:
We'll try 200 gap limit and 30 for lookahead. I'll create a PR for this.

darosior:
Addressed by #1311. Let's see how it goes now.

PR #1311, opened 2024-09-10
electrum: reduce load on server

jp1ac4:
Changing values as discussed in #1292.


darosior:
For what it's worth there is one nit i think i should clarify. In Liana we have both a gap limit (implicitly in bitcoind) and a look-ahead for the address->derivation index mapping. For the Electrum work you've introduced both a gap limit and look-ahead value there but i think it should be a single one, the Electrum-module-look-ahead is effectively the bitcoind-gap-limit. And it's called look-ahead in BDK iirc.

No need to change it now but i think clearer names for those constants would be:
- `GAP_LIMIT`
- `FULL_SCAN_GAP_LIMIT`

darosior:
utACK 4567e3280c2a3238249ae7f7d2ce862aa8493a98

PR #1312, opened 2024-09-10
installer: change liana connect description

### LINKED ISSUE #1308: Wording issue in Liana Connect choice
We just realized that in the Liana Connect choice the current wording could lead to misinterpretation in "legal" and practical terms:

image

In particular:
- "hosted" could give a wrong idea to the user, since we don't want anyone to think that it is a hosted wallet
- we should be clearer about the fact that we don't control the wallet and that only the user does

The proposed wording for the Liana Connect choice would be:

```
Use our service to instantly be ready to transact. Wizardsardine runs the infrastructure, allowing multiple computers or participants to connect and synchronize.

This is a simpler and safer option for people who want Wizardsardine to keep a backup of their descriptor. 
You are still in control of your keys, and Wizardsardine does not have any control over your funds, but it will be able to see the your wallet's information, associated to an email address.
Privacy focused users should run their own infrastructure instead.
```

Sorry @edouardparis for the 2nd rework on the wording, we didn't realize this issue with the old one.

edouardparis:
close #1308

### DIFF
--- a/gui/src/installer/view.rs
+++ b/gui/src/installer/view.rs
@@ -2508,7 +2508,7 @@ pub fn connection_step_connected<'a>(
         .into()
 }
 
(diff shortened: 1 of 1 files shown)


edouardparis (commented):
Self-ACK ead2047e804e284bec0a623f834de1c09cfc839c

PR #1316, opened 2024-09-11
electrum: don't roll the tip forward upon reorg

jp1ac4:
For Electrum, we detect reorgs while syncing based on our local chain.

When a user requests a rescan, we roll the DB back to genesis block, but leave the local chain as it is. If there's then a reorg, we would find the common ancestor with respect to our local chain, which would be later than the DB's tip. In order not to miss any updates, we shouldn't roll the DB tip forward in that case.

### DIFF
--- a/src/bitcoin/poller/looper.rs
+++ b/src/bitcoin/poller/looper.rs
@@ -275,8 +275,20 @@ fn updates(
         Ok(Some(reorg_common_ancestor)) => {
             // The block chain was reorganized. Rollback our state down to the common ancestor
             // between our former chain and the new one, then restart fresh.
-            db_conn.rollback_tip(&reorg_common_ancestor);
-            log::info!("Tip was rolled back to '{}'.", &reorg_common_ancestor);
+            // Make sure the common ancestor is not higher than the current DB tip, which could
+            // happen if a rescan has been detected and the DB tip rolled back accordingly.
+            if reorg_common_ancestor.height <= current_tip.height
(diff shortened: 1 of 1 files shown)


darosior:
Just re-kicked all functional in CI to see if this really fixed it.

darosior:
Concept ACK

jp1ac4:
Note some failures for both bitcoind & electrs in PR rebased on current master:
https://github.com/wizardsardine/liana/pull/1303/checks?check_run_id=29991124649
https://github.com/wizardsardine/liana/pull/1303/checks?check_run_id=29991124486

darosior:
Are we sure those are due to a roll-forward and would be fixed by this patch?

jp1ac4:
> Are we sure those are due to a roll-forward and would be fixed by this patch?

Not sure at this stage. It may be something else due to bitcoind also failing.

darosior:
Well i'm going to... Re-run with Terminal access!

darosior (commented):
utACK 87593befd4a85accbab7e7417b322ea74393b69d.

I'd very much like to see a regression test for this in a followup.

jp1ac4:
> I'd very much like to see a regression test for this in a followup.

Note to anyone implementing this: we need the poller to detect a rescan, roll the DB back to genesis block, and then for a reorg to occur before the next poller iteration starts so that a common ancestor with height > 0 is returned by `sync_wallet()` while the DB tip is still 0.

PR #1318, opened 2024-09-11
build: add Makefile for Rust noobs

knocte:
With this minimal wrappers, one could just download&install Rust and after that just type 'make', no need to be knowledgeable about how to build things in Rust ecosystem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,2 @@
+all:
+	cargo build
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,5 @@
-all:
+configure:
+	@which cargo > /dev/null || (echo "Please install Rust 1.70 (or newer) toolchain; e.g. with 'apt install cargo', or using rustup" && exit 1)
+
+all: configure
 	cargo build
--- a/make.bat
+++ b/make.bat
@@ -1,0 +1,1 @@
+cargo build


darosior:
So that's effectively simply an alias for `cargo build`? I don't think that's worth it. The annoying part is installing the Rust toolchain, once that's done typing `cargo build` in place of `make` makes no difference.

knocte:
> once that's done typing cargo build in place of make makes no difference.

For developing, it's much faster to type it. I could also add a Makefile target for doing a release build via `make rel`?

knocte:
> For developing, it's much faster to type it. I could also add a Makefile target for doing a release build via make rel?

Also, it is a convention to run the tests with `make check`, and from the CI file I can gather that the best flags for running them are kinda long to type.

pythcoiner:
> So that's effectively simply an alias for `cargo build`? I don't think that's worth it. The annoying part is installing the Rust toolchain, once that's done typing `cargo build` in place of `make` makes no difference.

agree, what is annoying for a noob is toolchain + linux dependencies

knocte:
> agree, what is annoying for a noob is toolchain + linux dependencies

To that end, I could contribute a configure.sh script that checks if all necessary Rust tools (and proper minimal versions) are installed, and if they are not, it would advise the user how to install them.

knocte:
> To that end, I could contribute a configure.sh script that checks if all necessary Rust tools (and proper minimal versions) are installed, and if they are not, it would advise the user how to install them.

I added a 2nd commit that adds this check inside the Makefile actually (and gives the user a hint of 2 examples of how to install Rust). I could improve this later so that it checks the specific version of Rust that is needed.

RandyMcMillan:
I will submit ‘make help’ configuration PR

RandyMcMillan:
https://github.com/wizardsardine/liana/pull/1352

knocte:
> I will submit ‘make help’ configuration PR

IMHO that PR doesn't follow the K.I.S.S. principle 😅

pythcoiner:
I'm not sure this will be merged anytime soon, closing for now.

PR #1320, opened 2024-09-11
gui(settings): fix daemon reload handling for electrum

jp1ac4:
In case of error reloading daemon for Electrum, set edited to false (same handling as for bitcoind).


darosior:
tested ACK ff3d497fe12a70c606e9953cfb29fa4af0520859. Not bulletproof as i need to get back to the settings and re-click save with the previous settings for the daemon to be restarted, but still helps a lot to have the previous-working-config ready to be saved.

PR #1321, opened 2024-09-11
Deserialize error from create_spend http response

edouardparis:
Instead of having an unclear and ugly message like:
data did not match any variant of untagged enum DraftPsbtResult


edouardparis (commented):
Self-ACK b73193ae77dccde519d43e5042324742dbee5523

Issue #1322, opened 2024-09-11
Fail to broadcast a transaction: minimum relay fee not met


nondiremanuel:
During the last phase of tests, there was an occurrence of not being able to relay a transaction because the minimum fee was not met. 

![image](https://github.com/user-attachments/assets/3642538d-332e-416a-9115-60fafcd8c6f5)

Since it is not clear where the rounding/estimation issue comes from (not clear if #1132 or else), we agreed that if after a quick further investigation the origin is still unknown, we will go with a temporary patch for v7, i.e. either using 1.1 sats/vb as a multiplier when 1 sats/vb is selected or adding 10 sats to the total fee amount if 1 sats/vb was selected. This should give some margin (at a small cost) to avoid that the user is not able to relay the transaction. In such case, a more detailed fix would then be released in a followup.

darosior:
Marking as done for v7 since a temporary hack has been applied for this release in #1323. But leaving open as we should pin down the root cause of this issue.

nondiremanuel:
Thanks. I moved it to v8 so that we don't loose track of it and it still counts as "done" for v7.

jp1ac4:
I think this particular issue was fixed by #1451, which addressed a situation in which we could end up with a feerate below 1 sat/vb. That PR also reverted the temporary change from #1323 mentioned above.

There's another issue relating to feerates in #1132, but I don't think that could have caused this issue (as explained in https://github.com/wizardsardine/liana/issues/1132#issuecomment-2340396972). I'll therefore close this issue

jp1ac4:
I've just encountered this error with "min relay fee not met, 111 < 112" for a Taproot descriptor with a single key in the primary path.

jp1ac4:
Seems to be related to Specter DIY.

For two similar descriptors, a self-send signed with Specter DIY has an extra weight unit:
https://mempool.space/signet/tx/a221a6a833d0621e5fc44840f7eb28674765f61dfc5c72b1b30ddcf7506c7498

compared to otherwise:
https://mempool.space/signet/tx/d3b7ae90ab94704c391e25794382f0f5486f85db3deee900d404a59dde8e57c4

jp1ac4:
The issue seems to be that we need to add an extra 1 to allow for a sighash flag when using a Schnorr signature. The Specter DIY is adding a sighash flag.

The rust miniscript library's plan module is returning 64 for the Schnorr signature size and 1 for the push, but needs another 1 for the sighash flag.
[...]

PR #1323, opened 2024-09-11
spend: ensure fee above min relay value

jp1ac4:
This is a quick change to address #1322.

It adds 10 sats to the fee for a non-replacement transaction targeted at 1 sat/vb.


darosior:
utACK 74a53baa39c409d8a39a27affc5ebac34128cb24. Looking forward to a real fix.

PR #1324, opened 2024-09-11
electrum: fix update of spend_txid for spending coins

jp1ac4:
If the spend txid of a coin changes, we should update with the new txid.

This would affect a spending coin for which the spending transaction changes and the new spend transaction is confirmed in a single poll.

### DIFF
--- a/src/bitcoin/mod.rs
+++ b/src/bitcoin/mod.rs
@@ -502,7 +502,18 @@ impl BitcoinInterface for electrum::Electrum {
                     expired_spending.push(*op);
                 }
                 if let Some(block) = w_c.spend_block {
-                    spent.push((*op, *spend_txid, block.height, block.time));
+                    if let Some(w_spend_txid) = w_c.spend_txid {
+                        spent.push((*op, w_spend_txid, block.height, block.time));
+                    } else {
+                        log::error!(
+                            "Coin at {} has spend_block {:?} \
+                            without any spend_txid. It previously \
+                            had spend_txid {}",
+                            op,
+                            block,
+                            spend_txid,
+                        );
+                    }
                 }
             }
         }
--- a/tests/test_rpc.py
+++ b/tests/test_rpc.py
@@ -1226,6 +1226,62 @@ def test_rbfpsbt_bump_fee(lianad, bitcoind):
         )
     )
 
+    # Now we test that if a tx is replaced externally to Liana and
+    # immediately confirmed, the spend_txid of the now-spent coin
+    # is set to the replacement txid.
+    # The replacement has to be done externally so that Liana doesn't
+    # see it until it is confirmed.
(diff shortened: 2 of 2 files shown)


darosior:
A regression test would demonstrate whether the fix is necessary. :shrug:

src/bitcoin/mod.rs
```diff
@@ -502,7 +502,13 @@ impl BitcoinInterface for electrum::Electrum {
                     expired_spending.push(*op);
                 }
                 if let Some(block) = w_c.spend_block {
-                    spent.push((*op, *spend_txid, block.height, block.time));
+                    spent.push((
+                        *op,
+                        w_c.spend_txid
+                            .expect("must have spend_txid if spend_block present"),
```
darosior, line 508: Can we wrap it in an `if let Some(coin_spend_txid) = ` and `log::error` in the `else` branch instead, just in case?

jp1ac4: Sure.

jp1ac4:
> A regression test would demonstrate whether the fix is necessary. 🤷

Given that we iterate over the spending coins from the DB first and then the updated spending info, I think we will always use the correct value without this change as we'll first insert the wrong value and then the right value. I wasn't able to write a functional test to trigger this error.

jp1ac4:
I think we should still include this change in the future so that we don't rely on the order of iteration.

src/bitcoin/mod.rs
jp1ac4: Done.

jp1ac4:
I've pushed the functional test that would fail without this PR if the order of iteration of spending coins were reversed.

tests/test_rpc.py
```diff
@@ -1226,6 +1226,62 @@ def test_rbfpsbt_bump_fee(lianad, bitcoind):
         )
     )
 
+    # Now we test that if a tx is replaced externally to Liana and
+    # immediately confirmed, the spend_txid of the now-spent coin
+    # is set to the replacement txid.
+    # The replacement has to be done externally so that Liana doesn't
+    # see it until it is confirmed.
```
darosior, line 1233: It would still see it when it's included in the mempool.

jp1ac4: I should add to this comment that I stop and start lianad while the replacement is broadcast and confirmed to ensure that it isn't seen by Liana in the mempool.

darosior:
> EDIT: Related to this, we should probably modify how the two iterators are chained together so that we only include spending coins from the DB if they're not part of the other, up-to-date iterator. We shouldn't be passing the same outpoint more than once to spent_coins().

I agree, looks like this code is only accidentally correct.

pythcoiner:
This PR is stale for 1+ years, I'm closing, feel free to re-open if you re-gain some interest in it.

PR #1346, opened 2024-09-17
Check coldcard version and request user to upgrade

### LINKED ISSUE #1343: Edit Coldcard descriptor registration error message
In the descriptor registration on the signing devices step, we currently don't check the firmware version of a Coldcard to tell the user that miniscript is supported only on the Edge one.

The temporary quick fix would be to add for all registration failures happening on Coldcard, independent from the firmware version, an informative note. The error message if the device is Coldcard would then become:

"Failed to register descriptor. Note: the Edge firmware is currently required for Miniscript support"

Is it possible to insert it in the v7 release? If not, I will modify this issue to properly check the firmware version in the following release (which is already WIP if I understood correctly).

edouardparis:
Close #1343

### DIFF
--- a/gui/ui/src/component/hw.rs
+++ b/gui/ui/src/component/hw.rs
@@ -348,6 +348,37 @@ pub fn unsupported_hardware_wallet<'a, T: 'a, K: Display, V: Display>(
     .padding(10)
 }
 
+pub fn unsupported_version_hardware_wallet<'a, T: 'a, K: Display, V: Display, S: Display>(
+    kind: K,
+    version: Option<V>,
+    requested_version: S,
+) -> Container<'a, T> {
+    container(
+        row(vec![
+            column(vec![
+                text::p1_bold("Unsupported firmware version").into(),
+                text::p1_regular(format!("Install version {} or later", requested_version)).into(),
+                Row::new()
+                    .spacing(5)
+                    .push(text::caption(kind.to_string()))
+                    .push_maybe(version.map(|v| text::caption(v.to_string())))
+                    .into(),
+            ])
+            .width(Length::Fill)
+            .into(),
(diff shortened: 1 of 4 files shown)


pythcoiner:
image
the message is the same wether we choose PSWSH or taproo, maybe if taproot is selected we should ask for version 6.3.3?

jp1ac4 (approved):
utACK 7371c9df40e4201876c38f4dad11fa92ec1396ce.

gui/ui/src/component/hw.rs
```diff
@@ -348,6 +348,41 @@ pub fn unsupported_hardware_wallet<'a, T: 'a, K: Display, V: Display>(
...
+    kind: K,
+    version: Option<V>,
+    requested_version: S,
+) -> Container<'a, T> {
+    container(
+        row(vec![
+            column(vec![
+                text::p1_bold("Unsupported firmware version").into(),
+                text::p1_regular(format!(
+                    "Install version superior or equal to {}",
```
jp1ac4, line 361: nit: I think it would sound more natural to say "Install version {} or later".

edouardparis:
> the message is the same wether we choose PSWSH or taproo, maybe if taproot is selected we should ask for version 6.3.3?

We avoid to add more logic for now, it is past rc and goes straight into release.

pythcoiner:
tACK 7371c9df

jp1ac4 (approved):
utACK https://github.com/wizardsardine/liana/commit/7124e09b7555f46a9c16e265eed4bf247ea8326c.

edouardparis (commented):
Self-ACK 7124e09b7555f46a9c16e265eed4bf247ea8326c

PR #1350, opened 2024-09-18
Add compression algo flag to dpkg command for arch build

edouardparis:
Arch package failed to be extracted during pacman update because of the default zstd algo


edouardparis (commented):
Self-ACK 953a85b94321093d29926283df951c3f4f00d2aa

PR #1358, opened 2024-09-20
hw: check if app is open before sending old version error message

edouardparis:
We check if we are able to fetch master fingerprint to see if the bitcoin app is open. The error is then displayed with the generic connection error message.


pythcoiner:
FYI https://github.com/darosior/ledger_installer/issues/14#issue-2429603453

edouardparis (commented):
Self-ACK 37cea480b83651171739edc453a41c3dde1c14ba
I have taken into account Pythcoiner comment

PR #1359, opened 2024-09-20
fix history events and txs pagination

edouardparis:
The problem:
We display the button if the size of the list is a multiple of the page size. Otherwise, we assume that the user has reached a partial page, indicating they've reached the bottom. However, since we chose to include all transactions with the same block time on a single page (because it's not possible to order them by block index in the database, as we don't store the block index), this can cause the page size to exceed the set limit. As a result, the list size may not be a multiple of the page size.

The solution:
keep in the state if the last fetched page has a length inferior of the expected page size and do not display the button 'See more' then as we reached the bottom.

### DIFF
--- a/gui/src/app/state/mod.rs
+++ b/gui/src/app/state/mod.rs
@@ -21,6 +21,8 @@ use liana_ui::widget::*;
 
 use super::{cache::Cache, error::Error, menu::Menu, message::Message, view, wallet::Wallet};
 
+pub const HISTORY_EVENT_PAGE_SIZE: u64 = 20;
+
 use crate::daemon::{
     model::{remaining_sequence, Coin, HistoryTransaction, Labelled},
     Daemon,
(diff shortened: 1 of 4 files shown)


jp1ac4:
Perhaps it's not directly related to this PR, but when testing on my home page, I initially see one payment shown, and then  I have to click "See more" four times before the remaining two payments appear and the button disappears. Would it be possible to filter the txs each time so that as soon as I click once on "See more", the remaining payments appear?

edouardparis:
https://github.com/wizardsardine/liana/pull/1359#issuecomment-2363746516
Why do you mean unrelated to this PR ? Do you have the same behavior on master ?

jp1ac4:
Sorry, I meant that it might be an existing issue that should be addressed in a different PR.

The behaviour is different on master: on the home page, when I click "See more" once, the button disappears and no new rows appear, whereas in this PR, I can click on "See more" more than once, and after the fourth click, two more payments appear.

edouardparis:
I added a processing state, that makes the button not clickable the time the new page is loaded. Does it make the behavior more understandable ?

jp1ac4:
> I added a processing state, that makes the button not clickable the time the new page is loaded. Does it make the behavior more understandable ?

It's the same behaviour as before, except now each time I click the button, the "Fetching..." message appears.

It seems that the issue is that each time I click "See more", it's fetching new transactions, but then there's an additional filter on the GUI that means none of those newly fetched items are shown.

Issue #1363, opened 2024-09-26
[GUI][Connect] Refresh cache more often while wallet height is 0


jp1ac4:
The cache is refreshed every 120 seconds when using Liana Connect as a backend compared to 10 seconds otherwise.

When first creating the wallet, we will indicate to the user in #1361 that the wallet is still syncing. For Liana Connect, the wallet may appear to be syncing even once this has completed due to the longer cache refresh period.

For Liana Connect backends, we could refresh the cache more often while the wallet is still performing its initial sync (i.e. its height is still 0) and then revert to the current frequency.

edouardparis:
I agree this would be better UX :+1:

pythcoiner:
#1421 have been merged, can we close this issue?

nondiremanuel:
Sure, I forgot to close it.

PR #1370, opened 2024-10-08
[GUI] Indicate wallet is syncing after completing installer

### LINKED ISSUE #1361: [GUI] Indicate wallet is syncing after creation
When the GUI first opens after creating a wallet, the wallet is not yet synced against the blockchain (its height is returned as 0 by `getinfo`).

We should indicate that the wallet is still syncing, which is especially important for users who've imported a wallet and expect to see transactions.

It may look something like this:
image

With the currently available information, this would be shown while the wallet height is 0 (this might mean on regtest a wallet is shown as syncing even if there is only the genesis block).

A follow-up would be to include the blockchain height in `getinfo` so that we can compare the wallet height with the blockchain height (which could be useful to detect that a wallet has not been opened for a while and needs to catch up).

Note that for users with a local bitcoind, a warning may also be shown regarding the need for a rescan (#804).

jp1ac4:
This is to resolve #1361.

The home page considers the wallet to be syncing if its height is 0. In this case, the balance will slowly blink and a "Syncing..." text will appear just below.

The home page will check the wallet's height upon each cache refresh, and will reload the home page once the syncing has completed so that the updated balance is displayed without the user needing to do anything.

Both the blinking balance and "Syncing..." text use a new `Carousel` widget that cycles through different child widgets at a specified rate.

EDIT: I've added an extra commit to address #1363 as that is also related to the wallet height and cache refresh.

### DIFF
--- a/gui/src/app/cache.rs
+++ b/gui/src/app/cache.rs
@@ -2,7 +2,7 @@ use crate::daemon::model::Coin;
 use liana::miniscript::bitcoin::Network;
 use std::path::PathBuf;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Cache {
     pub datadir_path: PathBuf,
     pub network: Network,
--- a/gui/src/app/message.rs
+++ b/gui/src/app/message.rs
@@ -20,6 +20,7 @@ use crate::{
 pub enum Message {
     Tick,
     UpdateCache(Result<Cache, Error>),
+    UpdatePanelCache(/* is current panel */ bool, Result<Cache, Error>),
     View(view::Message),
     LoadDaemonConfig(Box<DaemonConfig>),
     DaemonConfigLoaded(Result<(), Error>),
--- a/gui/src/app/mod.rs
+++ b/gui/src/app/mod.rs
@@ -221,12 +221,16 @@ impl App {
         Subscription::batch(vec![
             time::every(Duration::from_secs(
                 // LianaLite has no rescan feature, the cache refresh loop is only
-                // to fetch the new block height tip which is only used to warn user
-                // about recovery availability.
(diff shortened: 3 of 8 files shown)


jp1ac4:
Here's how it looks when importing a descriptor:
[Screencast from Import Descriptor](https://github.com/user-attachments/assets/c90229cd-2337-499a-b0dc-861c2fd91575)

jp1ac4:
I've made the following changes:
1. Update the home page from the cache even when the home page is not currently selected. This way, the user won't need to wait on the home page for the syncing status to disappear and can browse other pages in the meanwhile.
2. Given (1), the home page will only reload automatically once syncing is completed if the home page is currently selected. Otherwise, it will reload once the user returns to the home page.
3. The home page's `reload` method will only reload data if the wallet is not currently syncing.

edouardparis:
Note: while importing descriptor with a bitcoin core, the synced state is still empty which is normal because user has to trigger a rescan but it may be confusing as the "syncing" implies that the wallet is updating to a correct state.

edouardparis (approved):
Self-ACK 6f4eb79a5a45751a1c919565ea9c88f59fbda0ab

kloaec:
That's a good point, and bad UX from our import descriptor flow.

I suppose the user should ask for the rescan at import time, before going to the wallet.
We could also, instead or at the same time, put an explanation banner on every empty wallet (no tx), telling them to check the Support page for rescan? (to be written)

nondiremanuel:
As per what was discussed offline:

> We could also not show the "Syncing..." message for a local bitcoind wallet with height 0 to avoid misleading the user. In that case, the UX will be the same as currently

This should be a valid option since in the cases in which wallet height is 0 we don't expect any update from the first poll.

Issue #1373, opened 2024-10-09
When opening an existing Liana wallet, advice users that wallet is syncing


nondiremanuel:
This is part of the "Wallet syncing / scanning state at wallet opening" activity included in v8.

The requirement would be to have the same syncing interaction that was developed for a new wallet ([#1370](https://github.com/wizardsardine/liana/pull/1370)) to when a user opens a wallet already present in his Liana until the first poll runs. This would mean for the user to have a syncing state for a maximum of `30 sec (poller frequency) + poll execution time` every time he opens the wallet. In that time, the user would see the last known amount blinking and a text indicating that the syncing has not yet been completed.

If we are worried that this could be detrimental for users that open the wallet frequently, we could potentially evaluate to have this behavior only if the wallet has not been open for a while (> `arbitrary threshold`). I personally think it's more consistent to keep it for every opening.

One way that was thought to achieve this requirement was to get the blockchain height and compare it to the wallet height, but since the mismatch between this two values occurs also in different moments than the wallet opening, this would mean repeating the syncing behavior every time we detect the block height to raise until the next poll, which is not ideal imho (see: https://github.com/wizardsardine/liana/issues/1362#issuecomment-2402359386). We decided to explore different ways of approaching it and see what could be best.

jp1ac4:
One approach could be for the `getinfo` command, which the GUI calls every 10 seconds for a local backend, to return the timestamp of the last poll, if any (note that we don't currently store this value). Then the GUI can check this value when it opens and consider the wallet to be syncing until the first poll, i.e. the first change of this timestamp.

For Liana Connect, we could leave this timestamp empty for now as a wallet using Liana Connect as a backend will remain synced even when the GUI is closed. As such, we would need to make sure the check of this timestamp is only done for a local backend.

For a local backend, I think this check on first poll completing could be done instead of the check on wallet height becoming positive in #1370 in order to cover opening both new and existing wallets, but for a Liana Connect backend, we would still need to check the height.

nondiremanuel:
> One approach could be for the `getinfo` command, which the GUI calls every 10 seconds for a local backend, to return the timestamp of the last poll, if any (note that we don't currently store this value). Then the GUI can check this value when it opens and consider the wallet to be syncing until the first poll, i.e. the first change of this timestamp.
> 
[...]

PR #1376, opened 2024-10-11
Get timestamp of last completed poll of the blockchain

jp1ac4:
This is a first step towards #1373.

The timestamp of the last completed poll of the blockchain will be stored in the database, and this value will be made available via the `getinfo` command.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -309,10 +309,10 @@ impl DaemonControl {
     /// Get information about the current state of the daemon
     pub fn get_info(&self) -> GetInfoResult {
         let mut db_conn = self.db.connection();
-
         let block_height = db_conn.chain_tip().map(|tip| tip.height).unwrap_or(0);
-        let rescan_progress = db_conn
-            .rescan_timestamp()
+        let wallet = db_conn.wallet();
+        let rescan_progress = wallet
+            .rescan_timestamp
             .map(|_| self.bitcoin.rescan_progress().unwrap_or(1.0));
         GetInfoResult {
             version: VERSION.to_string(),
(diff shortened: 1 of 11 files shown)


jp1ac4:
I added a commit to ensure the poller doesn't run while the blockchain is syncing. This was already the case for our regular polling, but if a user were to broadcast a spend while syncing is in progress, which could be done only via liana-cli currently, the poller would indeed run.

Running the poller during syncing could take a while to finish as the tip height would be more likely to change before completion, causing the poller to restart each time. We may want to allow this in future (e.g. to sync the wallet during IBD?) but for now I think the poller behaviour should be consistent between regular polling and that triggered by a spend broadcast.

This change also ensures that we only store the last poll timestamp for a poll against a synced blockchain, which is the implicit assumption for downstream logic in the GUI, i.e. when the first poll completes, the wallet is synced. Although we don't want downstream logic to determine the upstream implementation, I don't think storing the timestamp of a poll against a non-synced blockchain would be very useful on its own.

jp1ac4:
I made a couple of changes to the sqlite method `set_wallet_last_poll_timestamp()`:
- update a single row in wallets table where `id = WALLET_ID`
- return `Result<(), SqliteDbError>`

src/commands/mod.rs
```diff
@@ -324,6 +324,7 @@ impl DaemonControl {
             },
             rescan_progress,
             timestamp: db_conn.timestamp(),
+            last_poll_timestamp: db_conn.last_poll_timestamp(),
```
edouardparis, line 327: seems to me that we do two database call to get db_wallet for timestamp() and last_poll_timestamp(), can we have instead wallet = db_conn.wallet() and then wallet.timestamp and wallet.last_poll_timestamp

jp1ac4: Yes I think that would be better. I'll add a new `wallet()` method to the DB interface.

jp1ac4:
I've added a `Wallet` struct and corresponding `wallet()` method to the database interface.

Perhaps I should use `u32` for `last_poll_timestamp` for consistency with the other timestamps. I did consider it previously, but now we have the `Wallet` struct the inconsistency is more apparent.

edouardparis:
yes, I agree for consistency u32 would be better

jp1ac4:
I changed the `last_poll_timestamp` to `u32`.

edouardparis (approved):
ACK dee9554c518793b57b71965d835905300ef86fbc

PR #1377, opened 2024-10-11
[GUI] Consider wallet to be syncing until first poll completes after opening

jp1ac4:
This PR uses the changes from #1376 to complete #1373.

It builds on the changes from #1370 to consider an existing wallet (that uses a local backend) to be syncing until the first poll completes after opening the GUI, where "existing wallet" means one that has positive height (since a newly created wallet has height 0).

Note that for an external Liana daemon,  this logic is only applied in case the last poll timestamp has already been set when starting the GUI. Otherwise, we can't be sure if this external daemon will ever set this value.

### DIFF
--- a/gui/Cargo.lock
+++ b/gui/Cargo.lock
@@ -2624,7 +2624,7 @@ dependencies = [
 [[package]]
 name = "liana"
 version = "7.0.0"
-source = "git+https://github.com/wizardsardine/liana?branch=master#18950c219a9233e584fabb1f5b168e17fa450902"
+source = "git+https://github.com/wizardsardine/liana?branch=master#c2ce025a78891307e1f0d83c52cb983eb098d41d"
 dependencies = [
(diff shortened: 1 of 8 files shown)


jp1ac4:
I moved some commits to #1370 as they also applied there (see https://github.com/wizardsardine/liana/pull/1370#issuecomment-2407650398).

jp1ac4:
I added the `last_poll_timestamp` to the cache when loading the application.

jp1ac4:
Updated liana dependency commit hash.

jp1ac4:
Rebased on master.

jp1ac4:
Rebased on #1387.

jp1ac4:
Rebased on master.

edouardparis (approved):
ACK ec7556ee12f13c87229974769667bb892695889c

PR #1381, opened 2024-10-14
Change primary and secondary button for consistency

### LINKED ISSUE #1380: Distinguish between primary and secondary buttons
In the context of the installer UI/UX revamping, it was recently discussed the need to have a distinction between primary buttons (i.e. the ones we suggest the user to click in order to follow a specific "default" flow) and the secondary ones (i.e. other buttons which can be useful but that are not usually part of the "default" flow). The former should be more visible, most likely already filled with green. The latter can probably stay as they are, but on hover, instead of being filled in green, they should behave as it happen with fields, with the border becoming green and the filling remaining the same.

One other remark regarding the primary buttons usage is that we should have maximum one primary button for each screen. We could also have situation in which there is no primary button, especially in situations where the hierarchy of the buttons is not defined (e.g. choosing among different templates).

What to treat as primary button in each screen should be evaluated properly and it can be source of discussion in this issue.

edouardparis:
close #1380

We removed all the primary buttons, they will be introduced back in the GUI incrementally to follow a one primary button per page policy

The secondary button is border grey then border and text green on hover, whereas primary button will have its background green to be a call to action to the user.
The background of the secondary button is changed for a light grey in order to be consistent with items list selection.

20241014_12h47m01s_grim
20241014_12h47m24s_grim
20241014_12h47m42s_grim

### DIFF
--- a/gui/src/app/view/coins.rs
+++ b/gui/src/app/view/coins.rs
@@ -279,7 +279,7 @@ fn coin_list_view<'a>(
                         } else {
                             Column::new().push(
                                 Row::new().push(Space::with_width(Length::Fill)).push(
-                                    button::primary(Some(icon::arrow_repeat()), "Refresh coin")
+                                    button::secondary(Some(icon::arrow_repeat()), "Refresh coin")
                                         .on_press(Message::Menu(Menu::RefreshCoins(vec![
                                             coin.outpoint,
                                         ]))),
--- a/gui/src/app/view/home.rs
+++ b/gui/src/app/view/home.rs
@@ -87,9 +87,10 @@ pub fn home_view<'a>(
                             .width(Length::Fill),
                         )
                         .push(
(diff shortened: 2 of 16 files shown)


jp1ac4:
Perhaps the "Ignore" button here should not be filled in order to be the same as "Save" (except for the colour):

image
Hover "Ignore":
image
Hover "Save":
image

nondiremanuel:
> Perhaps the "Ignore" button here should not be filled in order to be the same as "Save" (except for the colour):
> 
> image Hover "Ignore": image Hover "Save": image

I personally think that in such case "Save" should be primary (green) and "Ignore" secondary (without any red). But maybe this is part of the activity of reviewing all the buttons hierarchy.

edouardparis (commented):
Self-ACK 3e33353745212e6d344c2b172941c8eae1bdcb53

PR #1386, opened 2024-10-17
[GUI] Skip loading screen for existing wallets

### LINKED ISSUE #1384: Don't block the opening of an existing wallet that uses a local backend even when blockchain syncing ongoing
An activity related to #1377 would be to allow a user that has a Liana wallet already installed with local backend to not need to wait the blockchain syncing to open it. This is particularly relevant for the case of Liana managed bitcoind.
A practical usecase would be one that we encountered recently: if a user needs to urgently receive funds and he didn't open the wallet for a while, he should not have to wait for the blockchain synchronization to be able to generate an address. He should be able to open the wallet while being aware of the fact that the wallet is still syncing and that the information could be not up-to-date.

Since the solution for #1377 was to show the "syncing" state until the first poll runs and since the first poll typically runs only after the blockchain synchronization*, as a first step it should be enough to just eliminate the syncing step in the GUI and directly open the wallet while still performing the sync in background. The fact that the wallet is syncing is already communicated to the user by the syncing state implemented in the above PR. 

A second step of this work (which is not part of this issue), could be to improve the information communicated to the user by showing the progress on the blockchain syncing if this is happening, while keeping it generic for the cases in which it is synced and we are waiting for the wallet to catch up (i.e. the poller has not run yet).

For the moment, I would not apply this change to the IBD situation, i.e. if the user never opened the wallet before (checked via the wallet height). This is because the progress bar step after the backend choice seems to be an important moment for the user to realize the time and space implications of using a managed bitcoind and could affect its choice of using other more immediate solutions such as Liana Connect. This is valid also considering that we don't currently give the possibility of switching to a different backend after opening the wallet (other than recreating it). This point is open to discussion and could be also targeted later, when we will have a proper communication of the progress of the syncing also in the wallet home.

*The only corner case in which this is not the case seems to be when the user broadcasts a transaction while the blockchain is still syncing. This automatically triggers a poll and would make the syncing state disappear. As per the discussion we had with @jp1ac4, we could manage such case by not automatically performing the poll after broadcasting a transaction if the blockchain sync is still ongoing, which seems a consistent behavior also in general.

jp1ac4:
This is to resolve #1384 by skipping the loading screen for an existing wallet even if the blockchain is still syncing.

This change will only make a difference for local backends given that the blockchain for a Liana Connect wallet is always considered to be synced and so the loading screen was already loading the application directly.

It builds on #1377, using the logic added there to indicate on the home page that syncing is in progress if the poller has not yet run. Polling does not start until the blockchain has fully synced and so the same logic can be applied here.


jp1ac4:
Rebased on #1377.

jp1ac4:
Rebased on #1377.

jp1ac4:
Rebased on https://github.com/wizardsardine/liana/pull/1377.

edouardparis:
Works like a charm. I think have better UX/UI to imagine, but I am happy with the result.
One think I did not check, what if a wallet was synced (blockheight > 0), then user deleted his whole bitcoind datadir and restarted it ? 
does it skip the launcher and user has a home with 2% syncing blockchain instead of a usual 99.8% synced ?

jp1ac4:
> does it skip the launcher and user has a home with 2% syncing blockchain instead of a usual 99.8% synced ?

That's right, it would load the home page directly and show the sync progress as returned by `getinfo`, so starting from 0%.

edouardparis:
Ok thank you for the answer, it is not really inconvenient.

edouardparis:
ACK 0a674d591affd6190d5b4979718fb09d3d8f79f3

PR #1387, opened 2024-10-18
[GUI] Don't treat wallets using bitcoind as syncing

jp1ac4:
This is a follow-up to #1370.

If the user has imported a descriptor and is using bitcoind as a local node, then they will need to perform a rescan in order 
to see past transactions. Treating the wallet as syncing in this case could mislead the user that a rescan is being performed. Therefore, it's better to keep the past behaviour here to avoid further confusion.

### DIFF
--- a/gui/src/app/mod.rs
+++ b/gui/src/app/mod.rs
@@ -63,7 +63,12 @@ impl Panels {
     ) -> Panels {
         Self {
             current: Menu::Home,
-            home: Home::new(wallet.clone(), &cache.coins, cache.blockheight),
+            home: Home::new(
+                wallet.clone(),
+                &cache.coins,
+                cache.blockheight,
+                daemon_backend.clone(),
+            ),
             coins: CoinsPanel::new(&cache.coins, wallet.main_descriptor.first_timelock_value()),
             transactions: TransactionsPanel::new(wallet.clone()),
             psbts: PsbtsPanel::new(wallet.clone()),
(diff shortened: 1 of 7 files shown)


edouardparis:
I think you can modify DaemonBackend for:

```rust
use crate::node::NodeType;

#[derive(Debug, Clone, PartialEq, Eq)]
pub enum DaemonBackend {
    EmbeddedLianad(NodeType),
    ExternalLianad,
    RemoteBackend,
}
```
It may be more practical for the gui.

jp1ac4:
> It may be more practical for the gui, wdyt ?

That looks good, I'll change it to that.

Do you think it would make sense at some point to add "node_type" to `getinfo` ? In that case, we would also know for an external daemon, but it might not be worth making such a change just for this use case of opening a wallet for the first time.

edouardparis:
> Do you think it would make sense at some point to add "node_type" to getinfo ? In that case, we would also know for an external daemon, but it might not be worth making such a change just for this use case of opening a wallet for the first time.

Yes, I believe it is not worth it for now.

jp1ac4:
I've made the suggested change except I decided to use `Option<NodeType>` just to be on the safe side as the config has `Option<BitcoinBackend>`. We could consider changing that in future, but for this current change I didn't want to risk introducing a panic.

edouardparis (approved):
ACK 9208cd11b23a9d234cba008fff3a820bc9d3a702

PR #1389, opened 2024-10-21
gui: display unconfirmed changes as confirmed

### LINKED ISSUE #1375: Change should not be shown as unconfirmed
During the testing phase of v7, we realized that when sending a tx that doesn't spend the whole coin, the change is shown in the home screen as "Unconfirmed" until the tx is mined. This is not the expected behavior for change, since it should be considered already confirmed.

A first analysis was already performed. Quoting from the discussion we had: 

> GUI does not use the specific endpoint /refresh of the backend, because it uses the common interface with the daemon which is the generate_spend. 
So the change address is just added to the recipients. 
> The fix:
change the payload to pstbs/generate to include a field change_address, that takes the derivation index and the address generated by the gui.

pythcoiner:
This PR fixes #1375.
It displays all changes as confirmed even if the coin is not in a block.
Note: we assert `is_change` means we control the coin but it's not always true: it just means the coin belongs to the change descriptor, not that the coins is controlled by this wallet. To solve this we need to add as `from_self` field in `Coins` that identify if all input of the funding tx are controlled by this wallet, this will be done in a follow-up.

### DIFF
--- a/gui/src/app/state/mod.rs
+++ b/gui/src/app/state/mod.rs
@@ -180,6 +180,8 @@ impl State for Home {
                                 } else {
                                     self.remaining_sequence = Some(seq);
                                 }
+                            } else if coin.is_change {
+                                self.balance += coin.amount;
                             } else {
                                 self.unconfirmed_balance += coin.amount;
                             }


jp1ac4:
Would it be possible to look up the coin's transaction in `self.pending_events` and check if it has `TransactionKind::SendToSelf`? I suppose on the first page reload after the coin is created, whether this lookup finds anything would depend on whether the `coins` or `pending_events` are populated first, but after that, I guess it should be quite reliable.

pythcoiner:
the follow up PR is implemented already, i'll open it tmr and we can check wich way is better

PR #1391, opened 2024-10-22
lianad: Add `from_self` to `ListCoinEntry`

pythcoiner:
This PR is a follow-up to #1389:
 - add `from_self` field to `ListCoinEntry` in order to keep track if all input of a tx producing the coin are owned by the wallet
 - add `from_self` field to `coins` table in db
 - add a `get_transaction(txid)` method to `DatabaseConnection` trait

`from_self` value is `true` if all input funding the transaction that produce the coin are owned by this wallet


pythcoiner:
Note: db migration is keeped for a follow-up as it seems other PR will do some changes on the db during this release cycle, i guess a global migration should be implemented before v8 release.

PR #1394, opened 2024-10-23
electrum: use debug log level for sync messages

jp1ac4:
These sync messages appear typically every 30 seconds and are only useful for debugging purposes.

The full scan messages can remain at info level for now as they are only expected to occur once for a new wallet and then only in case a rescan is performed.


edouardparis (approved):
utACK f9b7a25b40f7599c71e1e5be9bc03d9a25de7647

PR #1403, opened 2024-10-30
Disable edit of defined key in multisig template recovery path

edouardparis:
The primary keys shouldn't be editable from the recovery path in the Multisig Savings template but only from the primary path. It could bring to weird behaviors and corner cases.

### DIFF
--- a/gui/src/installer/view/editor/mod.rs
+++ b/gui/src/installer/view/editor/mod.rs
@@ -128,6 +128,39 @@ pub fn path(
     .into()
 }
 
+pub fn uneditable_defined_key<'a>(
+    alias: &'a str,
+    color: iced::Color,
+    title: impl Into<Cow<'a, str>>,
+    warning: Option<&'static str>,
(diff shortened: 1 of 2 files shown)


gui/src/installer/view/editor/mod.rs
```diff
@@ -128,6 +128,39 @@ pub fn path(
     .into()
 }
 
+pub fn uneditable_defined_key<'a>(
```
jp1ac4, line 131: This and the `defined_key` function seem quite similar so would it be worth creating a common function with an `editable` parameter, say?

edouardparis: I was thinking about it, but it felt not good to have multiple bool args (with the `fixed` arg) that concern multiple templates for the single safe multisig case.

jp1ac4 (approved):
Tested ACK 59c73001a4c9eaa835f2a9345d995f263653ab48.

PR #1417, opened 2024-10-31
gui(installer): re-add next button when starting managed bitcoind

jp1ac4:
This Next button was accidentally removed in commit bd03cc9cfffb7c5c45dd0dd60d54db807d4e02a2.


edouardparis:
Good, can you refac this to remove the unecessary Row and use press_maybe ?

PR #1419, opened 2024-11-01
gui(edit key modal): increase size of the closing icon

### LINKED ISSUE #1409: Increase the size of closing the popup icon ("X")
The closing icon of the key setup popup seems a bit small. We should possibly increase its size:

image

pythcoiner:
Fixes #1409 
image

### DIFF
--- a/gui/src/installer/view/editor/mod.rs
+++ b/gui/src/installer/view/editor/mod.rs
@@ -271,11 +271,13 @@ pub fn edit_key_modal<'a>(
             Column::new()
                 .spacing(25)
                 .push(Row::new()
+                    .push(h3(title))
                     .push(Space::with_width(Length::Fill))
-                    .push(button::transparent(Some(icon::cross_icon()), "").on_press(Message::Close)))
+                    .push(button::transparent(Some(icon::cross_icon().size(40)), "").on_press(Message::Close))
+                    .align_items(Alignment::Center)
+                )
                 .push(
                     Column::new()
-                        .push(h3(title))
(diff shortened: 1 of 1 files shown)


jp1ac4:
This new size looks better. The only thing is I think the empty space above "Set your key", which was already quite a lot, has increased. Perhaps the X should be in the same row as the title, or otherwise maybe there's a way to reduce some of the empty space.

pythcoiner:
agree, i've tryed to limit the height of the cross button, but it makes the icon disapear

pythcoiner:
i've moved the title in the same column:
image

pythcoiner:
and centered widgets of the first row:
image

gui/src/installer/view/editor/mod.rs
```diff
@@ -271,11 +271,14 @@ pub fn edit_key_modal<'a>(
             Column::new()
                 .spacing(25)
                 .push(Row::new()
+                    .push(Space::with_width(10))
```
jp1ac4, line 274: The title and the "Select the signing device..." wording below previously had the same left-alignment. Would it be possible to remove this space in order to keep them aligned?

pythcoiner:
i first add this space (before centering the title) as spacing from top & left were looking different, but now we are not so close from the top border, you're right it's looks better:
image

jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/commit/f4596ab4b0974d0b3986c0431337b8a304333278.

Looks great :heart_eyes:

PR #1420, opened 2024-11-01
gui(installer): change collapse icon from left arrow to righ arrow

### LINKED ISSUE #1416: Collapsed "Advanced Settings" indicator should point to the right (>) instead of left (<)
As @indiarodeia was pointing out, it is common practice to have an arrow pointing to the right (>) when a section is collapsed, down when it's expanded. At the moment we have one pointing to the left (<) when collapsed, while correctly down when expanded. We should align this behavior with the most used standards.

image

pythcoiner:
Fixes #1416 

image

### DIFF
--- a/liana-ui/src/icon.rs
+++ b/liana-ui/src/icon.rs
@@ -92,7 +92,7 @@ pub fn pencil_icon() -> Text<'static> {
 }
 
 pub fn collapse_icon() -> Text<'static> {
-    bootstrap_icon('\u{F284}')
+    bootstrap_icon('\u{F285}')
 }
 
 pub fn collapsed_icon() -> Text<'static> {


jp1ac4:
I wonder if the arrow should be on the left-hand side, i.e. ` > Advanced settings` ?

nondiremanuel:
I've seen it used in both ways so I think it's up to what we prefer. 
Since we will need to edit the "Advanced Settings" again in the future as part of the general UX/UI improvement (it is not clear rn and it only contains the segwit vs taproot choice), I would do what's quickest / what you like most and not take too much time to discuss it.

nondiremanuel:
I would keep the arrow on the right of the Menu name (i.e. "Advanced Settings >"), right-pointing when the menu is collapsed and down-pointing when it's expanded.
It should be aligned in the other parts of the software in which such element is present (ideally it should be at design system level), but let me know if it's better to manage it separately and we can create another issue for that.

edouardparis (approved):
utACK 53e773a0ff28d0428b32fd57b13df1fec93f851e

PR #1425, opened 2024-11-04
gui(send): display max calculated amount w/ 8 decimals

### LINKED ISSUE #1424: Don't trim trailing zeros on the send screen
The send screen on v7.0.0 trims trailing zeros of the amounts calculated automatically by checking the MAX box. Displaying all 8 decimal digits makes it easier to check the amounts.

pythcoiner:
fixes #1424 
image

### DIFF
--- a/liana-gui/src/app/view/spend/mod.rs
+++ b/liana-gui/src/app/view/spend/mod.rs
@@ -8,7 +8,7 @@ use iced::{
 
 use liana::{
     descriptors::LianaPolicy,
-    miniscript::bitcoin::{bip32::Fingerprint, Amount, Network},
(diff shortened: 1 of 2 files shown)


jp1ac4:
It might look even nicer if you could use the `amount_as_string` function:
https://github.com/wizardsardine/liana/blob/ff968ea51fc67274d5a71bf8aa5ba1989f8244ea/gui/ui/src/component/amount.rs#L42

There would be some inconsistency with non-MAX recipients, but it might still look fine as those other amounts are editable. Perhaps later we could use this same formatting for all recipients somehow.

pythcoiner:
done, ideed it's better.
image

edouardparis (approved):
utACK 7ab6d05f0bc0c00cfe9a160e8320f4c7de1ae212

PR #1412, opened 2024-10-31
gui(installer): make edit key modal scrollable

jp1ac4:
This is to fix #1402 following the suggestion in https://github.com/wizardsardine/liana/issues/1402#issuecomment-2448730840.

PR #1426, opened 2024-11-04
gui(key modal): make the list of keys/signers scrollable

### LINKED ISSUE #1402: Installer - "Apply" button sometimes not filled when setting up a key
When trying to setup the keys in the different templates, sometimes the "Apply" button is not correctly formatted, missing the content and text:

image

image

pythcoiner:
this is a variant of #1412, it make only signer/key list scrollable instead the whole modal
fixes #1402

### DIFF
--- a/gui/src/installer/view/editor/mod.rs
+++ b/gui/src/installer/view/editor/mod.rs
@@ -265,6 +265,12 @@ pub fn edit_key_modal<'a>(
     duplicate_master_fg: bool,
 ) -> Element<'a, Message> {
     let key_len = hws.len() + keys.len();
+    let scrollable_enable = match (chosen_signer.is_some(), manually_imported_xpub, key_len) {
+        (true, _, _) => true,
+        (_, true, l) => l > 0,
+        (false, _, 0) => false,
+        _ => false,
+    };
     let key_column = Column::new()
         .spacing(10)
         .push(Column::with_children(hws).spacing(10))
@@ -328,7 +334,7 @@ pub fn edit_key_modal<'a>(
         )
             .padding(15)
     ;
-    let scroll = scrollable(key_column).height(if key_len > 1 {
+    let scroll = scrollable(key_column).height(if scrollable_enable {
         Length::Fill
     } else {
         Length::Shrink
--- a/gui/src/installer/view/editor/mod.rs
+++ b/gui/src/installer/view/editor/mod.rs
@@ -1,6 +1,6 @@
 pub mod template;
 
-use iced::widget::{container, pick_list, slider, Button, Space};
+use iced::widget::{container, pick_list, scrollable, slider, Button, Space};
 use iced::{Alignment, Length};
 
 use liana::miniscript::bitcoin::Network;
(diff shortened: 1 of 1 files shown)


jp1ac4:
When first opening the modal, there's a lot empty space below the keys. Is it possible to remove that?

image

pythcoiner:
hum, the issue is if we define a scrollable w/o height or with Length::Shrink, it will not be scrollable.

One solution is to put a fixed size for the scrollable:
```diff
diff --git a/gui/src/installer/view/editor/mod.rs b/gui/src/installer/view/editor/mod.rs
index 3b7bdf9..b3fe63a 100644
--- a/gui/src/installer/view/editor/mod.rs
+++ b/gui/src/installer/view/editor/mod.rs
@@ -327,7 +327,7 @@ pub fn edit_key_modal<'a>(
         )
             .padding(15)
     ;
-    let scroll = scrollable(key_column).height(Length::Fill);
+    let scroll = scrollable(key_column).height(300);
     Column::new()
         .padding(25)
         .push_maybe(error.map(|e| card::error("Failed to import xpub", e.to_string())))

```

(we can modify this fixed height dinamically, but we need to get the window height, and iirc, should add a subscription to catch the signal...)

jp1ac4:
Perhaps you could you set the height to `Length::Fill` only if there's at least one non-hot key?

pythcoiner:
> Perhaps you could you set the height to `Length::Fill` only if there's at least one non-hot key?

done!

jp1ac4:
Unfortunately there's still a problem when an xpub has been entered manually:

image

Maybe this can also be taken into account in the keys length. Or perhaps we can do something simpler for now, such as the fixed height approach you suggested above.

pythcoiner:
or i can put the `Key name` card in the scrollable and trigger the scroll if more than 3 items into, but i think w/ many keys, user will not see `Key name` card appear, as it will be in the hidden part of the scrollable....

jp1ac4:
> i reduced the number of key to > 1 for trigger the scrollable

The problem still occurs with this condition. It doesn't seem to occur with  `> 0`, but then this gives a lot of empty space when the window is large. Perhaps the fixed height approach would be better? Or otherwise #1412 for now.

> or i can put the `Key name` card in the scrollable and trigger the scroll if more than 3 items into, but i think w/ many keys, user will not see `Key name` card appear, as it will be in the hidden part of the scrollable....

I agree it's nicer to have the scrollable part only including the keys.

pythcoiner:
@jp1ac4 wdyt about the last change (i think it can be better if we auto scroll down when manual inserting an xpub & window minimised - a lot of conditions :smile:  - , but it's maybe a bit tricky)

jp1ac4:
Hmm, there's still this problem with having a lot of empty space when the window is large. Even though having only the keys scrollable is preferable, I would go with #1412 for now until we find a better way to manage the height.

nondiremanuel:
I agree with going with the simpler solution for now. I will move it to the following milestone in case we want to investigate more on this.

PR #1437, opened 2024-11-05
gui: reload home if wallet is not syncing

jp1ac4:
Following #1386, this PR will show past payments on the home page while the blockchain is still syncing. Currently, these are not loaded until the wallet has finished syncing.

This change won't apply if the wallet itself is still syncing. The reason is that this is not expected to take so long and so it's better to wait for the updated data. Furthermore, the DB may be locked if the poller is running, in which case we wouldn't be able to load past transactions until the wallet has synced anyway. The home page reloads automatically as soon as the wallet finishes syncing.


jp1ac4:
Rebased on master.

edouardparis (approved):
utACK 71ad55e06bfc804754983debd7502db02b6bc524

PR #1441, opened 2024-11-06
gui(installer): fix word plurality when adding keys

### LINKED ISSUE #1440: Nit: if I switch from 2of2 to 1of2, "keys" becomes "key" in the threshold definition, but it shouldn't
Key/keys should follow the singular/plural case of the total keys and not of the threshold ones (1 out of 2 keyS)

Below a couple of screenshots:
- With 2-of-2:
Screenshot from 2024-11-05 15-54-36

- With 1-of-2:
Screenshot from 2024-11-05 15-55-51

jp1ac4:
This is to fix #1440.

### DIFF
--- a/gui/src/installer/view/mod.rs
+++ b/gui/src/installer/view/mod.rs
@@ -1438,7 +1438,7 @@ pub fn defined_threshold<'a>(
                     "{} out of {} key{}",
                     threshold.0,
                     threshold.1,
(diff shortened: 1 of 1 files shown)


edouardparis (commented):
utACK d1bcb71ea4d4bdafe4b81fcfbfe78558b7fd8366

PR #1442, opened 2024-11-06
gui(installer): use fixed keys ordering

### LINKED ISSUE #1439: Aliases order in the wallet policy descriptor changes according to check on descriptor registration
See the following screencast:

https://github.com/user-attachments/assets/2e45a74e-bfa1-45eb-a4cc-76aa15080b30

This could be confusing for the user

jp1ac4:
This is to fix #1439. It follows the same approach as used in the settings to ensure consistent ordering when iterating over keys.


edouardparis:
I think it would be cleaner to change the keys alias map for a BTreeMap for which iterators always respect an order.

edouardparis (approved):
utACK 86cebc2ddf70b1cfe2d717eb379999ee68699a5e

PR #1446, opened 2024-11-07
gui(installer): list wallets using secondary buttons

### LINKED ISSUE #1445: When adding an existing wallet with Liana connect, all options are highlighted as primary
image

They shouldn't be highlighted but "secondary".

jp1ac4:
This is to fix #1445:

image


edouardparis (approved):
ACK ef55e8f01e65f67faff5bd2f07c2c1302d562dd9

Issue #1371, opened 2024-10-08
spend: feerate can be lower than target value


jp1ac4:
The fee of a Taproot (only affects Taproot?) transaction can sometimes be 1 sat lower than required to achieve the target feerate as defined by the coin selection algorithm, i.e. feerate = fee / (weight/4).

Note that #1132 is similar but relates to whether or not the vbytes are rounded up when calculating the feerate.

If we fix this, then we should revert the change from https://github.com/wizardsardine/liana/pull/1323.

jp1ac4:
This is possibly related to #1322.

jp1ac4:
I was able to reproduce this on Segwit with txid `a63c4a69be71fcba0e16f742a2697401fbc47ad7dff10a790b8f961004aa0ab4`.

It's a spend from the primary path of a Taproot descriptor, where the primary path is a 2-of-2 multisig, with target feerate 2 sat/vb.

The tx weight is 646 and fee 322. Even if calculating the feerate without rounding up the vbytes, this gives 322 / (646/4) = 1.99.

jp1ac4:
I think we're currently missing the varint length for `TapControlBlock` and `TapScript`, which is why this issue doesn't affect a key spend:
https://docs.rs/miniscript/11.0.0/src/miniscript/util.rs.html#19-39

This is also mentioned here:
https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1028718255

jp1ac4:
Perhaps something like this could work:

```diff
diff --git a/src/descriptors/mod.rs b/src/descriptors/mod.rs
index 818ecb19..6ee1034a 100644
--- a/src/descriptors/mod.rs
+++ b/src/descriptors/mod.rs
@@ -6,6 +6,7 @@ use miniscript::{
         secp256k1,
     },
     descriptor,
+    miniscript::satisfy::Placeholder,
     plan::{Assets, CanSign},
     psbt::{PsbtInputExt, PsbtOutputExt},
     translate_hash_clone, ForEachKey, TranslatePk, Translator,
@@ -270,15 +271,21 @@ impl LianaDescriptor {
                 .expect("unhardened index");
             let witscript_size = der_desc
                 .explicit_script()
-                .map(|s| varint_len(s.len()) + s.len())
-                .unwrap_or(0);
+                .map(|s| varint_len(s.len()) + s.len());
 
             // Finally, compute the satisfaction template for the primary path and get its size.
-            der_desc
[...]

PR #1451, opened 2024-11-08
descriptors: fix max satisfaction weight for Taproot multisig primary path spends

jp1ac4:
This is a fix to #1371 and the related #1322.

As per https://github.com/wizardsardine/liana/issues/1371#issuecomment-2446921093, this adds the varint length for the script and control block elements of a Taproot primary path spend.

I think this is what caused #1322 and so I have reverted the change from #1323.

### DIFF
--- a/src/descriptors/mod.rs
+++ b/src/descriptors/mod.rs
@@ -6,6 +6,7 @@ use miniscript::{
         secp256k1,
     },
     descriptor,
+    miniscript::satisfy::Placeholder,
     plan::{Assets, CanSign},
     psbt::{PsbtInputExt, PsbtOutputExt},
     translate_hash_clone, ForEachKey, TranslatePk, Translator,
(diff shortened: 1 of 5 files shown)


src/descriptors/mod.rs
```diff
@@ -259,26 +260,35 @@ impl LianaDescriptor {
...
+            plan.witness_size()
+                + witscript_size.unwrap_or_else(|_| {
+                    plan.witness_template()
+                        .iter()
+                        .map(|elem| match elem {
+                            // We need to calculate the size manually before calculating the varint length.
+                            // See https://docs.rs/miniscript/11.0.0/src/miniscript/util.rs.html#35-36.
+                            Placeholder::TapScript(s) => varint_len(s.len()),
+                            Placeholder::TapControlBlock(cb) => varint_len(cb.serialize().len()),
+                            _ => 0,
```
darosior, line 288: So any stack item but the leaf script and control block account for the varint length? This is incredibly confusing, it'd be worth opening an issue there. (And if they fix it it would be nice to also fix https://github.com/rust-bitcoin/rust-miniscript/issues/701.)

darosior:
light-ACK. The code looks correct to me but i haven't reviewed the test nor reviewed in depth but from a quick skim of their codebase it seems correct.

src/descriptors/mod.rs
jp1ac4: It appears so. I did some debugging to inspect all the elements and their sizes against an actual transaction and the sizes of these two placeholder elements didn't include the varint lengths.

I've created https://github.com/rust-bitcoin/rust-miniscript/issues/771 to track this.

edouardparis:
From Andrew Poelstra comment: https://github.com/rust-bitcoin/rust-miniscript/issues/771, the issue is acknowledged and it seems you are on the good track. I am okay to proceed to merge these changes.

edouardparis (approved):
utACK 6e74a96e12feb00fcf0f6974f56002c0356e0960

PR #1452, opened 2024-11-08
descriptors: fix unsigned tx max satisfaction weight

### LINKED ISSUE #1406: feerate of unsigned transaction can be overestimated
This is taken from https://github.com/wizardsardine/liana/issues/1371#issuecomment-2448131283.

This issue is about how we estimate the feerate of an unsigned transaction after creation. It's different from #1132 and #1371, which are about how we set the fee when creating a transaction.

The issue concerns the following function:
https://github.com/wizardsardine/liana/blob/cbc46e9e26e5978f194751baf142a52ad0c25a1b/src/descriptors/mod.rs#L541-L557

It seems to estimate the weight to be 2 less than the true weight of the signed transaction. I think this is because we call `tx.weight()` where all inputs of `tx` have empty witnesses, and so the transaction is serialized as non-Segwit:
https://docs.rs/bitcoin/0.31.0/src/bitcoin/blockdata/transaction.rs.html#752-753
https://docs.rs/bitcoin/0.31.0/src/bitcoin/blockdata/transaction.rs.html#968-979

This function is used by the GUI to display the estimated feerate of an unsigned PSBT as well as by the daemon to perform sanity checks on a PSBT after creation. By underestimating the weight, we may overestimate the feerate.

The suggested fix is to use `tx.weight() + 2` if `tx` has >= 1 input and all these inputs have empty witnesses.

jp1ac4:
This fixes #1406, building on the fix from #1451 for estimating the weight of each input.

An extra 2 is added to the unsigned transaction's weight to account for the Segwit marker and flag.

### DIFF
--- a/liana/src/descriptors/mod.rs
+++ b/liana/src/descriptors/mod.rs
@@ -555,6 +555,18 @@ impl LianaDescriptor {
         tx.weight()
             .to_wu()
             .checked_add(max_sat_weight.checked_mul(num_inputs).unwrap())
+            .and_then(|weight| {
+                weight.checked_add(
+                    // Make sure the Segwit marker and flag are included:
+                    // https://docs.rs/bitcoin/0.31.0/src/bitcoin/blockdata/transaction.rs.html#752-753
+                    // https://docs.rs/bitcoin/0.31.0/src/bitcoin/blockdata/transaction.rs.html#968-979
+                    if num_inputs > 0 && tx.input.iter().all(|txin| txin.witness.is_empty()) {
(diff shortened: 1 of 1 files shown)


darosior:
Haven't reviewed the test nor the context but the code looks good to me, for what it's worth.

jp1ac4:
Rebased on master.

edouardparis (approved):
utACK e4ae6dc368d6600fbbdd5adaea1d52feb0f4b2a9

Issue #907, opened 2024-01-04
GUI: Show blockchain sync progress as percentage to 2 decimal places


jp1ac4:
Currently, the percentage is shown to 2 decimal places, but the second decimal is always 0:

![image](https://github.com/wizardsardine/liana/assets/121959000/45822a5a-03e3-45e4-8b31-e0fa6bf60204)

I think this can be resolved by increasing the number of decimal places of the `"sync"` entry in Liana's `getinfo` command to 4.

kloaec:
@darosior can you tag this one "good first issue"? It should be a super easy one :)

AnmolSirola:
@jp1ac4 Can I take this one ?

jp1ac4:
Sure, that would be great thanks. Please note that it might not be possible to update the GUI's `liana` dependency yet as its rust-bitcoin dependency needs to be updated to v0.31 still, but it should be enough for this issue to update the "sync" entry as mentioned above and then the GUI's `liana` dependency will be updated elsewhere.

edouardparis:
We need to update liana-gui with liana master then

pythcoiner:
> We need to update liana-gui with liana master then

@thomask7b can you do this?

nondiremanuel:
> @nondiremanuel In #1386, the sync progress is shown on the home screen to 1 decimal place. Following #1457, we can now show it to 2 decimal places. Shall we use 2 decimal places on both the loading screen (for IBD) and home page (for existing wallets)?

If it's not too complicated, I agree to align the behavior. Not a major priority though.

PR #1457, opened 2024-11-12
Fix precision, merge roundup methods, add tests

thomask7b:
I've merged roundup_progress in rounded_up_progress since was in utils and was no longer used by any other method.
I also added some tests to verify the precision's update but also for specific cases that we manage. By the way tests has been moved in the related file.

Solves #907


jp1ac4:
Thanks very much for this PR. I tested the change is working as expected.

I think merging the two functions makes sense in some ways, but it could still be useful to keep them separate for testing the rounding logic by itself so I would keep them separate for now.

thomask7b:
I can adapt if you want but I'm not sure to understand what you mean, because the rounding logic is still tested with even more coverage.
Also think it adds complexity to have the function in another file when it can be in SyncProgress.
So we could move both metohds in SyncProgress. And then why not merge them and test the final method.
I don't understand what you dislike.
Let me know what you think and I'll adapt.

jp1ac4:
I meant to test the rounding logic just in terms of the floats without any regard to the number of blocks or headers. But I think your changes are good and either approach would be fine.

Given that there's already a separate utils function and in order to keep the changes small, I would keep this separate function as I don't think there's a clear advantage not to.

thomask7b:
I took in account your remarks.
I kept specific tests for `SyncProgress::rounded_up_progress` to verify that we "don't return a 100% progress until we are actually done syncing"

src/bitcoin/d/mod.rs
```diff
@@ -1262,7 +1262,7 @@ impl SyncProgress {
         let progress = roundup_progress(self.percentage);
         if progress == 1.0 && self.blocks != self.headers {
```
pythcoiner, line 1263: not related to this PR:
```suggestion
        if progress == 1.0 && self.blocks < self.headers {
```
@jp1ac4 i don't think this is likely to happen as iiuc as of now it's only used w/ a bitcoind backend, but wdyt we should do if `self.blocks` > `self.headers`? we actually return 0.999 but maybe 1.0 should be more relevant?  (i'm not sure from a security POV, if returning 0.999 i guess we will stuck in the syncing process while if returning 1.0 we will assumed we are synced)

pythcoiner:
lgtm:
image

src/bitcoin/d/utils.rs
```diff
@@ -371,12 +371,13 @@ mod tests {
...
-        assert_eq!(roundup_progress(0.67891), 0.678);
+        assert_eq!(roundup_progress(0.67891), 0.6789);
         assert_eq!(roundup_progress(0.98), 0.98);
         assert_eq!(roundup_progress(0.998), 0.998);
-        assert_eq!(roundup_progress(0.9476), 0.947);
-        assert_eq!(roundup_progress(0.998), 0.998);
-        assert_eq!(roundup_progress(0.9998), 1.0);
-        assert_eq!(roundup_progress(0.9991), 1.0);
+        assert_eq!(roundup_progress(0.9476), 0.9476);
+        assert_eq!(roundup_progress(0.94761), 0.9476);
```
jp1ac4, line 378: It would be good to include an additional test to show that numbers below 0.9999 are truncated rather than rounded up, e.g.
```diff
diff --git a/src/bitcoin/d/utils.rs b/src/bitcoin/d/utils.rs
index c317172b..ddd6d8af 100644
--- a/src/bitcoin/d/utils.rs
+++ b/src/bitcoin/d/utils.rs
@@ -376,6 +376,7 @@ mod tests {
         assert_eq!(roundup_progress(0.998), 0.998);
         assert_eq!(roundup_progress(0.9476), 0.9476);
         assert_eq!(roundup_progress(0.94761), 0.9476);
+        assert_eq!(roundup_progress(0.94769), 0.9476);
         assert_eq!(roundup_progress(0.9998), 0.9998);
         assert_eq!(roundup_progress(0.99998), 1.0);
         assert_eq!(roundup_progress(0.99991), 1.0);

```

pythcoiner:
tACK dc4feacdf30051479ad4eb324ee00256b7ef9361

jp1ac4:
You can squash your commits again before pushing. Thanks.

src/bitcoin/d/mod.rs
jp1ac4: Hmm I'm not sure really but I guess if they're not equal in either direction then something still needs to happen in terms of syncing.

pythcoiner:
ACK 439af75

jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/commit/439af752a57410f4667664a2d5de93ddb9fa705e.

Thanks very much.

Issue #255, opened 2022-12-18
Release: static binaries for Linux against musl


darosior:
I found a way to build against a downloaded stdlib in Docker for the MacOS and Windows reproducible builds.

The Guix reproducible build of Linux binaries unfortunately links against a somewhat recent `glibc`, `2.33`: https://github.com/revault/liana/pull/140#issuecomment-1337399445.

I think we should be able to build against musl within the Guix container, and have a `x86_64-linux-musl` artifact in the release in case users want to run on somewhat old systems (for instance, Debian stable...). 

Let's see if i can reasonably achieve that before the release.

pythcoiner:
not sure if this can be usefull there but i post it in case:

![image](https://github.com/wizardsardine/liana/assets/124568858/83a0766e-bc03-411b-a346-73f16a0fb43b)

find this [here](https://doc.rust-lang.org/rustc/codegen-options/index.html#link-self-contained)

darosior:
The latest about this is probably https://github.com/wizardsardine/liana/pull/1080#issue-2250989624.

Issue #414, opened 2023-04-06
Try to link against an older glibc for releases


darosior:
The release build for Linux is a reproducible and bootstrappable one using Guix (see https://github.com/wizardsardine/liana/pull/140 and the documentation at [`contrib/reproducible/`](https://github.com/wizardsardine/liana/tree/master/contrib/reproducible)).

The glibc this build links against is [Guix's `gcc-toolchain`](https://github.com/wizardsardine/liana/blob/73a90b412dac1d17d95fa4196771ab92af9149ad/contrib/reproducible/guix/manifest.scm#L10) which ships Glibc 2.33. Unfortunately, it's quite recent (February 2021) and some operating systems still only provide older glibc. See https://sourceware.org/glibc/wiki/Release for a table of glibc version per distribution version. I'd like to at least get Ubuntu Focal (Glibc 2.31) and Debian 10 (Glibc 2.28).

Unfortunately building against an older Glibc within the Guix container is non-trivial (to me). Another approach could be to link against musl (see https://github.com/wizardsardine/liana/issues/255).

The quickest solution for anyone running into an issue of the type:
```
./liana-gui: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./liana-gui)
```
Is to compile the project from source. There are instruction at [`doc/BUILD.md`](https://github.com/wizardsardine/liana/blob/master/doc/BUILD.md) on how to do this reasonably. But you can also just use the simple (not really decent) way of [installing the Rust toolchain from their website](https://www.rust-lang.org/learn/get-started) and build Liana like so (example for v1):
```
git clone https://github.com/wizardsardine/liana
cd liana/
git checkout v1.0
cd gui/
cargo run
```

darosior:
Removing from the 1.0 project. I've had another look at it and it is really non-trivial (to me) to build against an older glibc within the Guix container.

DesobedienteTecnologico:
As commented in this Merged PR, I will try to get hands on it by next week. It doesn't see that trivial to fix.
I you fix it I will try to keep track on this PR.

Thanks!

darosior:
I've added details about this issue to the OP. Since it's been hit (as expected..) by a couple users already i've added troubleshooting instructions (basically you have to build from source).

> I will try to get hands on it by next week

Good to hear. A PR for this would be very welcome. But note i do not want to compromise on reproducible and bootstrappable builds, so it has to happen within the Guix toolchain.

[...]

PR #1469, opened 2024-11-18
liana: remove daemon feature

edouardparis:
We expect user to use systemd service to manage a daemon or
to handle their process by themself.

The json-rpc server is exposed only for unix systems through
unix sockets, it is moved to its own module while the implementation
for windows is marked as TODO.


pythcoiner:
she should check if #255 and #414 can be closed after merging this (but i don't believe libc was used only for forking the process...)

edouardparis:
You are right, libc is required by rust anyway to call OS-level functions and system calls directly. I do not think the current change resolves #255 and #414

liana/src/config.rs
```diff
@@ -322,7 +312,6 @@ mod tests {
         {
             let toml_str = r#"
             data_dir = '/home/wizardsardine/custom/folder/'
-            daemon = false
```
jp1ac4, line 325: I think you could now remove the `#[cfg(unix)]` attribute above and elsewhere in this test.

jp1ac4 (commented):
I was able to run lianad, executing CLI commands and running the GUI.

As well as my comment in the code, I think you can also remove the `StartupError::Daemonization` variant now.

jp1ac4 (approved):
ACK https://github.com/wizardsardine/liana/commit/c14c55beeea56655096eb873ba0bce4d37ccd6bb.

pythcoiner:
tACK c14c55b run GUI and lianad

PR #1473, opened 2024-11-18
Update liana gui to import liana directly

edouardparis:
We import directly liana with the directory path instead of the git url.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2822,33 +2822,11 @@ dependencies = [
 ]
 
 [[package]]
-name = "liana"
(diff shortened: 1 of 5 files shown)


jp1ac4 (approved):
ACK b0fbe0839d15914cf1d56d35184bdc49683adfd3.

Tested running GUI with both external and embedded lianad.

edouardparis (commented):
Self-ACK 6cefd2464a55fbfaea5c4e9a3a370fb034c5462e
tested the reproducible build with this change

PR #1475, opened 2024-11-19
remove commands import in spend module

edouardparis:
This is preparatory work to create a specific liana lib.
We keep modules separated.


edouardparis (commented):
Self-ACK 4668980ff12c1026bf3a1b3b520d1d38db8939cb
Very small change

PR #1483, opened 2024-11-22
commands: indicate in `listcoins` response whether coin is from self

jp1ac4:
:warning: This PR upgrades and migrates DB version so a wallet opened against this PR will no longer work on Liana v8.

This is a first step towards #1375.

Following the approach from #1391, this PR adds a new `is_from_self` field to the response of the `listcoins` command.

The underlying information is stored in a new `is_from_self` column in the coins database table. This column could instead have been added to the transactions table, but for consistency with other transaction-related columns, I added it to the coins table. It's also important to note that being from self is wallet-dependent, so adding it to the transactions table would not work if multiple wallets were supported in the same DB (h/t edouardparis).

A subsequent PR will then use this field in the GUI to determine which unconfirmed coins, if any, can be included in the confirmed balance. It also needs to be added to the corresponding Liana Connect API call. Another use of this field will be to determine which unconfirmed coins to include in coin selection (https://github.com/wizardsardine/liana/issues/1484).

The first commit in this PR refactors some existing DB migration tests so that they will not be affected by future changes to the coins table schema.

### DIFF
--- a/lianad/src/database/sqlite/utils.rs
+++ b/lianad/src/database/sqlite/utils.rs
@@ -340,6 +340,148 @@ fn migrate_v6_to_v7(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError
     Ok(())
 }
 
+fn migrate_v7_to_v8(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError> {
+    // This migration is done as several database transactions in order not to
+    // have a very large database transaction containing all rows from the
+    // transactions table.
+    const TXIDS_BATCH_SIZE: u32 = 100;
+    loop {
+        let txids = db_query(
+            conn,
+            "SELECT txid FROM transactions WHERE num_inputs IS NULL LIMIT ?1",
+            rusqlite::params![TXIDS_BATCH_SIZE],
+            |row| {
+                let txid: Vec<u8> = row.get(0)?;
+                let txid: bitcoin::Txid = bitcoin::consensus::encode::deserialize(&txid)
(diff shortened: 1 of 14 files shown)


jp1ac4:
Set PR to draft as I might include changes to the `listcoins` command here.

jp1ac4:
I added commits to include the `is_from_self` field in the `listcoins` response.

I also added a check constraint on the coins table so that a coin cannot be both immature and from self. This would also apply to coinbase deposits in general, but we don't currently flag those.

jp1ac4:
I've now made the changes as outlined above.

For extra safety, I added a max number of iterations based on the number of unconfirmed coins when updating `is_from_self`. There should never be more than this number of iterations.

lianad/src/database/sqlite/utils.rs
```diff
@@ -307,6 +322,167 @@ fn migrate_v5_to_v6(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError
...
+        };
+        let mut i = 0;
+        loop {
+            i += 1;
+            if i > max_iterations {
+                panic!(
+                    "More than {} iterations while updating is_from_self.",
+                    max_iterations
+                );
+            }
```
edouardparis, line 436: I think you can directly do `for i in 0..max_iterations` with the updated == 0 as a break if it is finished early.

jp1ac4: I've changed it to a `for` loop and modified the logging slightly.

pythcoiner:
utACK 1c033861

PR #1488, opened 2024-11-28
Use cargo-zigbuild for reproducible build with glibc 2.31

### LINKED ISSUE #414: Try to link against an older glibc for releases

edouardparis:
Zig language has solved many issues regarding compilation linker.
It allows us to compile to a smaller version of glibc while using latest guix commit.
No need then to bootstrap multiple version of rust, the current supported version is 1.71.1

close #414

### DIFF
--- a/contrib/release/release.sh
+++ b/contrib/release/release.sh
@@ -38,13 +38,13 @@ TARGET_DIR="$BUILD_DIR" ./contrib/reproducible/docker/docker-build.sh
 (
     cd "$BUILD_DIR"
     create_dir "$LINUX_DIR_NAME"
-    cp "$BUILD_DIR/release/lianad" "$BUILD_DIR/release/liana-cli" "$BUILD_DIR/release/liana-gui" ../README.md "$LINUX_DIR_NAME"
+    cp "$BUILD_DIR/x86_64-unknown-linux-gnu/release/lianad" "$BUILD_DIR/x86_64-unknown-linux-gnu/release/liana-cli" "$BUILD_DIR/x86_64-unknown-linux-gnu/release/liana-gui" ../README.md "$LINUX_DIR_NAME"
     tar -czf "$LINUX_ARCHIVE" "$LINUX_DIR_NAME"
     cp "$LINUX_ARCHIVE" "$RELEASE_DIR"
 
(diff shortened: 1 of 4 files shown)


edouardparis:
Checked on #d2379bd with @jp1ac4 that liana-gui checksum is same for both of us.
```
f9612bde713fbe5e3e3c965d5d81631ee98e9763375df2d0c9da9ea1260203cf  release_build/x86_64-unknown-linux-gnu/release/liana-gui
```

edouardparis (commented):
Self-ACK e3acfa7f806a0694d29adb9369984c25760840e3

PR #1491, opened 2024-11-29
Enable cross compilation to windows and macos with nix

edouardparis:
The nix build replace the docker build. Sadly zip is not fully deterministic so release_assets may not have always the same hash, but users can still audit the binaries reproducible build.


edouardparis (commented):
Self-ACK 6b4e7b0521c7328647738b46afd67386273c403e

PR #1499, opened 2024-12-16
commands: use unconfirmed coins from self for coin selection

### LINKED ISSUE #1484: spend: include only those unconfirmed coins from self for coin selection
In https://github.com/wizardsardine/liana/pull/873, we included unconfirmed change for coin selection when creating a new spend. The intention of only considering change was to include only those unconfirmed coins that were not dependent on external transactions, but a coin being at a change address does not guarantee this.

Following #1483, we can make this condition exact by only including those unconfirmed coins that are from self.

jp1ac4:
This is to resolve #1484.

We previously used unconfirmed change for coin selection as a best-guess way of identifying those unconfirmed coins from self, but now we have the `is_from_self` field we can use it directly and be sure we only select coins that are fully under our control.

### DIFF
--- a/lianad/src/commands/mod.rs
+++ b/lianad/src/commands/mod.rs
@@ -495,7 +495,7 @@ impl DaemonControl {
         // the spend from a set of optional candidates.
         // Otherwise, only the specified coins will be used, all as mandatory candidates.
         let candidate_coins: Vec<CandidateCoin> = if coins_outpoints.is_empty() {
-            // From our unconfirmed coins, we only include those that are change outputs
+            // From our unconfirmed coins, we only include those that are from self
             // since unconfirmed external deposits are more at risk of being dropped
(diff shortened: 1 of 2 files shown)


edouardparis:
Can you change value here:
https://github.com/wizardsardine/liana/blob/master/liana-gui/src/lianalite/client/backend/mod.rs#L652

jp1ac4:
Ah, that's been changed as part of #1498. But I don't think that change is required in this PR as the logic for specifying candidates for selection is in the remote backend API handler so will need to be changed there separately.

edouardparis:
Ah right thanks, I merged it, can you rebase ?

jp1ac4:
Rebased on master.

edouardparis (approved):
Self-ACK 091d7728ee33d231fbafb839ac497f86122d0851

Issue #1291, opened 2024-09-07
Having to specify `ssl://` prefix for SSL connections is confusing


darosior:
The connection to the Electrum server might fail just because you forgot to manually specify to Liana to not talk in cleartext to the server thanks to a manually-input prefix to the URL. In terms of UX we can do better.

We could:
- Have a small text hint (possibly inside the text input itself) to not forget `ssl://`
- Have a dropdown between SSL and cleartext right before the text input with default to SSL
- Try to be smart and test both with and without SSL when not specified in the URL (could lead to silently downgrading from SSL to cleartext, so meh)
- Something else?

pythcoiner:
note we already have a hint but only if the address is invalid:
![image](https://github.com/user-attachments/assets/f6729606-2a47-4a9b-8e83-e2c9694c8e2c)

pythcoiner:
> i'd go for:
> 
> * in every case we don't care the prefix, ping both at install step, if one ping works we update input field
> * if input field is not ssl, we display a warning

i've implemented this approach in #1296

darosior:
At this point, this doesn't need to be in v7.

PR #1500, opened 2024-12-16
gui: allow to connect to an electrum server w/ a self signed certificate

### LINKED ISSUE #1300: electrum: fail to connect to a server w/ a self-signed certificate
we actually fails to connect (in ssl) to an electrum server that have a self-signed certificate, in order to allow it we should have a look to few points:

- add an option to installer/settings to allow user to disable the certificate check, we should keep the default behaviour to check the certificate.
- certificate check should be disabled if we set electrum_client::Config `validate_domain` to `false` but not works as intended if electrum_client is used w/ default `use-rusttls` feauture. It works fine w/ `use-openssl` feature btw.
( i've opened an [issue](https://github.com/bitcoindevkit/bdk/issues/1598) about this disrepancy in bdk repo)
- `bdk_electrum` crate does not offer `use-openssl` feature:
```
    Updating crates.io index
      Adding bdk_electrum v0.17.0 to dependencies
             Features:
             + use-rustls
             - use-rustls-ring
```
bdk_electrum `Cargo.toml`:
```
[dependencies]
bdk_core = { path = "../core", version = "0.1" }
electrum-client = { version = "0.21", features = [ "proxy" ], default-features = false }

[dev-dependencies]
bdk_testenv = { path = "../testenv", default-features = false }
bdk_chain = { path = "../chain", version = "0.18.0" }

[features]
default = ["use-rustls"]
use-rustls = ["electrum-client/use-rustls"]
use-rustls-ring = ["electrum-client/use-rustls-ring"]
```

pythcoiner:
closes #1300 
The issue about connecting to an electrum certificate using `rustls` have been fixed [upstream](https://github.com/bitcoindevkit/bdk/issues/1598) but in order to beneficiate from it we have to update `bdk_electrum` and `rust-bitvoin` dependencies.
Meanwhile, this PR introduce a workaround: the initial issue is related to `electrum-client` `use-rustls` feature and `use-openssl` feature is not reexported by `bdk_electrum` but we can use `electrum-client` crate directly and use `use-openssl` feature by this way:
 - [x] use `electrum-client` directly w/ `use-openssl`
 - [x] add and option to opt-out of ssl domain validation
 - [x] let user change the `validate_domain` values in `Settings` menu.
 
image

 
 Note: ssl://testnet.aranguren.org:51002 electrum server can be used to test this PR

### DIFF
--- a/contrib/lianad_config_example.toml
+++ b/contrib/lianad_config_example.toml
@@ -50,8 +50,13 @@ poll_interval_secs = 30
 # In order to connect, it needs the address as a string, which can be
 # optionally prefixed with "ssl://" or "tcp://". If omitted, "tcp://"
 # will be assumed.
+# `validate_domain` field is optional: used in case of SSL connection,
+# if set to `false`, internal electrum client will not try to validate
+# the domain associated to the certificate: it's useful in case of 
+# self-signed certificate. Its default value is `true`.
 # [electrum_config]
 # addr = "127.0.0.1:50001"
+# validate_domain = false 
 #
 #
 [bitcoind_config]
--- a/liana-gui/src/app/view/message.rs
+++ b/liana-gui/src/app/view/message.rs
@@ -88,6 +88,7 @@ pub enum RemoteBackendSettingsMessage {
 pub enum SettingsEditMessage {
     Select,
     FieldEdited(&'static str, String),
+    ValidateDomainEdited(bool),
     BitcoindRpcAuthTypeSelected(RpcAuthType),
     Cancel,
     Confirm,
--- a/liana-gui/src/app/view/settings.rs
+++ b/liana-gui/src/app/view/settings.rs
@@ -32,7 +32,7 @@ use crate::{
     hw::HardwareWallet,
     node::{
         bitcoind::{RpcAuthType, RpcAuthValues},
-        electrum,
+        electrum::{self, validate_domain_checkbox},
     },
 };
 
@@ -563,6 +563,7 @@ pub fn electrum_edit<'a>(
     blockheight: i32,
     addr: &form::Value<String>,
     processing: bool,
+    validate_domain: bool,
 ) -> Element<'a, SettingsEditMessage> {
     let mut col = Column::new().spacing(20);
     if is_configured_node_type && blockheight != 0 {
@@ -595,6 +596,9 @@ pub fn electrum_edit<'a>(
             .push(separation().width(Length::Fill));
     }
 
+    let checkbox = validate_domain_checkbox(addr, validate_domain, |b| {
+        SettingsEditMessage::ValidateDomainEdited(b)
+    });
     col = col.push(
         Column::new()
             .push(text("Address:").bold().small())
@@ -606,6 +610,7 @@ pub fn electrum_edit<'a>(
                 .size(P1_SIZE)
                 .padding(5),
             )
+            .push_maybe(checkbox)
             .push(text(electrum::ADDRESS_NOTES).size(P2_SIZE))
             .spacing(5),
     );
--- a/liana-gui/src/installer/message.rs
+++ b/liana-gui/src/installer/message.rs
@@ -81,6 +81,7 @@ pub enum DefineBitcoind {
 #[derive(Debug, Clone)]
 pub enum DefineElectrum {
     ConfigFieldEdited(electrum::ConfigField, String),
+    ValidDomainChanged(bool),
 }
 
 #[derive(Debug, Clone)]
--- a/liana-gui/src/installer/view/mod.rs
+++ b/liana-gui/src/installer/view/mod.rs
@@ -29,6 +29,7 @@ use liana_ui::{
     widget::*,
 };
 
+use crate::node::electrum::validate_domain_checkbox;
 use crate::{
     hw::{is_compatible_with_tapminiscript, HardwareWallet, UnsupportedReason},
     installer::{
@@ -1134,7 +1135,15 @@ pub fn define_bitcoind<'a>(
         .into()
 }
 
-pub fn define_electrum<'a>(address: &form::Value<String>) -> Element<'a, Message> {
+pub fn define_electrum<'a>(
+    address: &form::Value<String>,
+    validate_domain: bool,
+) -> Element<'a, Message> {
+    let checkbox = validate_domain_checkbox(address, validate_domain, |b| {
+        Message::DefineNode(DefineNode::DefineElectrum(
+            message::DefineElectrum::ValidDomainChanged(b),
+        ))
+    });
     let col_address = Column::new()
         .push(text("Address:").bold())
         .push(
@@ -1150,7 +1159,8 @@ pub fn define_electrum<'a>(address: &form::Value<String>) -> Element<'a, Message
             .size(text::P1_SIZE)
             .padding(10),
         )
-        .push(text(electrum::ADDRESS_NOTES).size(text::P2_SIZE))
+        .push_maybe(checkbox)
+        .push(text(electrum::ADDRESS_NOTES))
         .spacing(10);
 
     Column::new().push(col_address).spacing(50).into()
--- a/lianad/Cargo.toml
+++ b/lianad/Cargo.toml
@@ -28,6 +28,7 @@ miniscript = { version = "11.0", features = ["serde", "compiler", "base64"] }
 # For Electrum backend. This is the latest version with the same bitcoin version as
 # the miniscript dependency.
 bdk_electrum = { version = "0.14" }
+electrum-client = { version = "0.19", features = ["use-openssl"] }
 
 # Don't reinvent the wheel
 dirs = "5.0"
--- a/lianad/src/config.rs
+++ b/lianad/src/config.rs
@@ -128,6 +128,19 @@ pub struct ElectrumConfig {
     /// Include "ssl://" for SSL. otherwise TCP will be assumed.
     /// Can optionally prefix with "tcp://".
     pub addr: String,
+    /// If validate_domain == false, domain of ssl certificate will not be validated
+    /// (useful to allow usage of self signed certificates on local network)
+    #[serde(default)]
+    pub validate_domain: bool,
+}
+
+impl Default for ElectrumConfig {
+    fn default() -> Self {
+        Self {
+            addr: Default::default(),
+            validate_domain: true,
+        }
+    }
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
--- a/lianad/src/config.rs
+++ b/lianad/src/config.rs
@@ -122,7 +122,7 @@ pub struct BitcoindConfig {
 }
 
 /// Everything we need to know for talking to Electrum serenely.
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct ElectrumConfig {
     /// The URL the Electrum's RPC is listening on.
     /// Include "ssl://" for SSL. otherwise TCP will be assumed.
@@ -130,17 +130,12 @@ pub struct ElectrumConfig {
     pub addr: String,
     /// If validate_domain == false, domain of ssl certificate will not be validated
     /// (useful to allow usage of self signed certificates on local network)
-    #[serde(default)]
+    #[serde(default = "default_validate_domain")]
     pub validate_domain: bool,
 }
 
-impl Default for ElectrumConfig {
-    fn default() -> Self {
-        Self {
-            addr: Default::default(),
-            validate_domain: true,
-        }
-    }
+fn default_validate_domain() -> bool {
+    true
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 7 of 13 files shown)


jp1ac4:
In the GUI, could we add a "Use SSL" checkbox (related #1291), which, if selected, would then show the "validate domain" checkbox? If SSL is not required, then I feel it would be better not to show the validate option if it's not applicable.

Related to this, I wonder if we should use an enum in `ElectrumConfig` to include `validate_domain` only for the SSL case, e.g. something like `Tcp(url)` or `Ssl(url, validate_doman)`. I'm not sure `validate_domain` should be a manadatory field in the config file. Even in the SSL case, it could default to `false` if not present.

nondiremanuel:
As we were discussing, let's have for now the validation checkbox (which I would phrase as "Do not validate SSL Domain (check this only if you want to use a self-signed certificate)") only if the user uses ssl in the address.

pythcoiner:
the checkbox is now only displayed if the address is valid and contains "ssl://"

pythcoiner:
> Related to this, I wonder if we should use an enum in `ElectrumConfig` to include `validate_domain` only for the SSL case, e.g. something like `Tcp(url)` or `Ssl(url, validate_domain)`. I'm not sure `validate_domain` should be a mandatory field in the config file. Even in the SSL case, it could default to `false` if not present.

i think this can be done in a follow-up if we decide to add a checkbox for enable/disable ssl instead prepending it to the address.

jp1ac4:
> i think this can be done in a follow-up if we decide to add a checkbox for enable/disable ssl instead prepending it to the address.

OK yep, let's save that for a follow-up. We could probably keep the value saved in the config file as "ssl://..." and just change how it's parsed into `ElectrumConfig`.

For this present PR, I think you'll need to make `validate_domain` an `Option` so that it works with existing config files. In the case of SSL, it can then default to false if not present.

pythcoiner:
> > i think this can be done in a follow-up if we decide to add a checkbox for enable/disable ssl instead prepending it to the address.
> 
> OK yep, let's save that for a follow-up. We could probably keep the value saved in the config file as "ssl://..." and just change how it's parsed into `ElectrumConfig`.
> 
> For this present PR, I think you'll need to make `validate_domain` an `Option` so that it works with existing config files. In the case of SSL, it can then default to false if not present.

i think having a default as true can avoid footgun and make sense, as it's the default value of the bdk electrum client, see [here](https://github.com/bitcoindevkit/rust-electrum-client/blob/15f753f05f588b2e871e1dc4e3ca7201b09b27d7/src/config.rs#L146)

jp1ac4:
> i think having a default as true can avoid footgun and make sense, as it's the default value of the bdk electrum client, see [here](https://github.com/bitcoindevkit/rust-electrum-client/blob/15f753f05f588b2e871e1dc4e3ca7201b09b27d7/src/config.rs#L146)

Ah I see, and so this is also the value we've implicitly been using for SSL connections so far. I agree we can default to true then (also for the GUI checkbox).

pythcoiner:
> > i think having a default as true can avoid footgun and make sense, as it's the default value of the bdk electrum client, see [here](https://github.com/bitcoindevkit/rust-electrum-client/blob/15f753f05f588b2e871e1dc4e3ca7201b09b27d7/src/config.rs#L146)
> 
> Ah I see, and so this is also the value we've implicitly been using for SSL connections so far. I agree we can default to true then (also for the GUI checkbox).

the checkbox follow the default, but checked mean "do NOT validate", so it's unchecked as default

jp1ac4:
> the checkbox follow the default, but checked mean "do NOT validate", so it's unchecked as default

That sounds good then. I think it just remains to make `validate_domain` an `Option` in the config or otherwise to change how it's deserialized so that a missing value is treated as true.

pythcoiner:
> That sounds good then. I think it just remains to make `validate_domain` an `Option` in the config or otherwise to change how it's deserialized so that a missing value is treated as true.

i've implemented default deserialization

liana-gui/src/installer/message.rs
```diff
@@ -81,6 +81,7 @@ pub enum DefineBitcoind {
 #[derive(Debug, Clone)]
 pub enum DefineElectrum {
     ConfigFieldEdited(electrum::ConfigField, String),
+    ValidDomainChanged(bool),
```
jp1ac4, line 84: As `validate_domain` is part of the config, I feel that this should somehow also be part of `ConfigFieldEdited`. Perhaps you can use something like `ConfigFieldEdited(electrum::ConfigFieldValue)` where `ConfigFieldValue` is:
```
pub enum ConfigFieldValue {
    Address(String),
    ValidateDomain(bool),
}
```

liana-gui/src/app/view/settings.rs
```diff
@@ -583,6 +584,16 @@ pub fn electrum_edit<'a>(
             .push(separation().width(Length::Fill));
     }
 
+    let checkbox = checkbox(
+        "Do not validate SSL Domain(check this only if you want to use a self-signed certificate)",
+        !validate_domain,
+    )
+    .on_toggle(|b| SettingsEditMessage::ValidateDomainEdited(!b));
+    let checkbox = if addr.valid && addr.value.contains("ssl://") {
```
jp1ac4, line 592: nit: The function `is_electrum_address_valid(value: &str)` uses `value.starts_with("ssl://")` so you could use the same condition here (or even add it to a standalone `is_ssl()` function and then use it within `is_electrum_address_valid`).

Also, to avoid repeating this checkbox logic in the installer, you could add it to src/node/electrum.rs.

liana-gui/src/app/view/message.rs
```diff
@@ -88,6 +88,7 @@ pub enum RemoteBackendSettingsMessage {
 pub enum SettingsEditMessage {
     Select,
     FieldEdited(&'static str, String),
+    ValidateDomainEdited(bool),
```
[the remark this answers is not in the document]
pythcoiner: i think we should not do this until we need it? looks like adding uneeded complexity imho

lianad/src/config.rs
```diff
@@ -128,6 +128,19 @@ pub struct ElectrumConfig {
     /// Include "ssl://" for SSL. otherwise TCP will be assumed.
     /// Can optionally prefix with "tcp://".
     pub addr: String,
+    /// If validate_domain == false, domain of ssl certificate will not be validated
+    /// (useful to allow usage of self signed certificates on local network)
+    #[serde(default)]
```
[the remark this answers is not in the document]
pythcoiner: any advantage doing this? i feel it's more natural/standard to use the standard `Default` mechanism instead a custom one? having default values for the whole struct should not be a problem imho?

liana-gui/src/installer/message.rs
pythcoiner: `ValidDomainChanged(true)` is shorter than
`ConfigFieldEdited(electrum::ConfigFieldValue(electrum::ConfigFieldValue::ValidateDomain(bool)))`
the event system of iced is already overwhelming imho, we should avoid "overwrapping" it only make the codebase less readable & more error prone

jp1ac4: Wouldn't it be just `ConfigFieldEdited(electrum::ConfigFieldValue::ValidateDomain(bool))`?

And by removing `electrum::` it would be just `ConfigFieldEdited(ConfigFieldValue::ValidateDomain(bool))`.

It doesn't seem too bad, but we can leave it for now. If we have a separate message per field, then we should at some point change `ConfigFieldEdited` to `AddressEdited`.

liana-gui/src/app/view/message.rs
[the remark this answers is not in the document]
jp1ac4: I suppose so, we may do further refactoring at some point when we allow switching between bitcoind and Electrum nodes in the settings.

lianad/src/config.rs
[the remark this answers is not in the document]
jp1ac4: It wasn't clear to me that the default address value is useful or makes sense here, and I felt we should only add the necessary default values. But it's probably fine either way.

jp1ac4:
It would be good to add some comments to https://github.com/wizardsardine/liana/blob/master/contrib/lianad_config_example.toml about the new field.

liana-gui/src/installer/message.rs
pythcoiner: at this point of the release cycle i'd like to bring this PR quickly to a mergeable state or put it on hold: to avoid everyone testing under the christmas tree :smile:

pythcoiner:
> It would be good to add some comments to https://github.com/wizardsardine/liana/blob/master/contrib/lianad_config_example.toml about the new field.

adressed, feel free if you have better wording

liana-gui/src/installer/message.rs
jp1ac4: That's OK, I meant it as something to consider in the future.

contrib/lianad_config_example.toml
```diff
@@ -50,8 +50,13 @@ poll_interval_secs = 30
 # In order to connect, it needs the address as a string, which can be
 # optionally prefixed with "ssl://" or "tcp://". If omitted, "tcp://"
 # will be assumed.
+# `validate_domain` field is optionnal: used in case of SSL connection,
```
jp1ac4, line 53: typo: "optional".

contrib/lianad_config_example.toml
```diff
@@ -50,8 +50,13 @@ poll_interval_secs = 30
...
+# if set to `false`, internal electrum client will not try to validate
+# the domain associated to the certificate: it's usefull in case of
```
jp1ac4, line 55: typo: "useful"

contrib/lianad_config_example.toml
jp1ac4, line 53: You could add that it defaults to true.

jp1ac4 (commented):
Needs a rebase on master.

pythcoiner:
> Needs a rebase on master.

rebsed & addressed comments

jp1ac4:
In the case that SSL is being used, should we show in the read-only settings view a line to indicate whether `validate_domain` is true or false?

image

pythcoiner:
> In the case that SSL is being used, should we show in the read-only settings view a line to indicate whether `validate_domain` is true or false?
> 
> image

no strong opinion, ping @nondiremanuel

pythcoiner:
looks like CI failing on windows because missing openssl dependency...

lianad/src/config.rs
jp1ac4, line 133: This default doesn't seem to be working for me. If my config doesn't contain validate_domain, the value is set to false.

jp1ac4: It's probably worth adding some unit tests for the `ElectrumConfig` parsing to complement the other config parsing tests.

pythcoiner: fixed & added tests, the issue was serde use default of field type (false) and not default of the struct, my bad

pythcoiner:
[5e5e65e](https://github.com/wizardsardine/liana/pull/1500/commits/5e5e65e849e1a5d2c6bd03ac002ba2bfaf6e667c) seems to fix windows CI

nondiremanuel:
> > In the case that SSL is being used, should we show in the read-only settings view a line to indicate whether `validate_domain` is true or false?
> > image
> 
> no strong opinion, ping @nondiremanuel

It's not required at this stage imho.

contrib/lianad_config_example.toml
```diff
@@ -50,8 +50,13 @@ poll_interval_secs = 30
 # In order to connect, it needs the address as a string, which can be
 # optionally prefixed with "ssl://" or "tcp://". If omitted, "tcp://"
 # will be assumed.
+# `validate_domain` field is optional: used in case of SSL connection,
+# if set to `false`, internal electrum client will not try to validate
+# the domain associated to the certificate: it's useful in case of 
+# self-signed certificate. It's default value is `true`.
```
jp1ac4, line 56: nit: "It's" -> "Its"

liana-gui/src/installer/view/mod.rs
```diff
@@ -1150,7 +1159,8 @@ pub fn define_electrum<'a>(address: &form::Value<String>) -> Element<'a, Message
             .size(text::P1_SIZE)
             .padding(10),
         )
-        .push(text(electrum::ADDRESS_NOTES).size(text::P2_SIZE))
```
jp1ac4, line 1153: nit: Did you intend to remove the `.size(text::P2_SIZE)`?

lianad/Cargo.toml
```diff
@@ -28,6 +28,7 @@ miniscript = { version = "11.0", features = ["serde", "compiler", "base64"] }
 # For Electrum backend. This is the latest version with the same bitcoin version as
 # the miniscript dependency.
 bdk_electrum = { version = "0.14" }
+electrum-client = {version = "0.19", features =["use-openssl"]}
```
jp1ac4, line 31: nit(formatting): `features = ["use-openssl"] }`

jp1ac4 (approved):
Tested ACK https://github.com/wizardsardine/liana/commit/5e5e65e849e1a5d2c6bd03ac002ba2bfaf6e667c.

A couple of nits and a question, but we can come back to those if need be.

liana-gui/src/installer/view/mod.rs
pythcoiner: > yes I think I removed intentionally the size to stay consistant in the context, IIRC i notice the size was different from some others warnings

pythcoiner:
comments addressed

PR #1501, opened 2024-12-17
bump msrv 1.80

edouardparis:
The update of the reproducible system allow us to bump the minimal rust supported version. Although bumping a MSRV should be conservative, it is a requirement for us if we want to be able to support last https://github.com/iced-rs/iced version and improve the experience of our users.


edouardparis (commented):
Self-ACK c07c81659809ad23369fdcb56f3e3615816367e1

PR #1503, opened 2024-12-17
Introduce lighter struct Payment in home panel

edouardparis:
Some users may have txs with a lot of inputs/outputs. We wrongly iter over them in the view which is the origin of performance issues.


jp1ac4 (approved):
Tested ACK 79057a0cd9ea8010e91f9d7ea035a85c1b06bdb1.

PR #1506, opened 2024-12-30
lianad: fix check that rescan has started

jp1ac4:
The `startrescan` command uses the response from bitcoind's `listdescriptors` RPC method to confirm that the rescan has started.

This check includes a string comparison that can fail in some cases (Taproot descriptors with a spending path having only a single key) as the descriptor string returned by bitcoind may contain both `h` and `'` , whereas Liana's descriptor string only uses `'` (as per the `Descriptor::to_string()` method).

This PR changes the check to first parse each descriptor string as a `Descriptor` object to avoid any `h` / `'` mismatch issues.

### DIFF
--- a/lianad/src/bitcoin/d/mod.rs
+++ b/lianad/src/bitcoin/d/mod.rs
@@ -1060,11 +1060,7 @@ impl BitcoinD {
         let current_descs = self.list_descriptors();
 
         for desc in descs {
-            let present = current_descs
-                .iter()
-                .find(|entry| &entry.desc == desc)
-                .map(|entry| entry.timestamp == timestamp)
-                .unwrap_or(false);
+            let present = current_descs_contain_desc_timestamp(&current_descs, desc, timestamp);
             if !present {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


pythcoiner:
tACK w/ this (signet) wallet: 
```
tr([124b2e76/48'/1'/0'/2']tpubDFiSVLQkSqEZCopN3gby4hNDE7PG15mUrA5rjJWcMEzHFRzLY7bNCgeoVewjerasVejUizH7AWiZAZMKc7GWM7e4cj3TpUhsqLpgY3Rikca/<0;1>/*,and_v(v:pk([124b2e76/48'/1'/0'/2']tpubDFiSVLQkSqEZCopN3gby4hNDE7PG15mUrA5rjJWcMEzHFRzLY7bNCgeoVewjerasVejUizH7AWiZAZMKc7GWM7e4cj3TpUhsqLpgY3Rikca/<2;3>/*),older(65535)))#wvchffyz
```
rescan fail on master & success on [6b914fd](https://github.com/wizardsardine/liana/pull/1506/commits/6b914fd2491841bb3daa41f59641677c10cc880b)

edouardparis:
utACK 6b914fd2491841bb3daa41f59641677c10cc880b

PR #1507, opened 2024-12-31
fix: treat change outputs as send-to-self payments

jp1ac4:
This is a follow-up to #1503 to fix an issue with change outputs appearing as outgoing payments on the home page.

It expands the definition of `PaymentKind::SendToSelf` to include change outputs since these are also payments to self.


edouardparis (approved):
utACK 55b92b3da707310900fca1570269c5e13b709d64

PR #1508, opened 2024-12-31
[GUI] Improve coin selection logic when creating spend

### LINKED ISSUE #1326: Coin selection: no coins selected if the amount to send over balance amount
image
but `BTC left to select` amount looks correct

jp1ac4:
This is to resolve #1326. In case of insufficient funds when creating a spend, the selected coins will be updated to correspond to the missing amount. Previously, no changes were made to selected coins.

Furthermore, if a user selects the max option for a recipient, only owned coins (confirmed or unconfirmed from self) will be considered rather than all coins, since unconfirmed coins that are not from self should only be used if the user has selected them manually.

### DIFF
--- a/liana-gui/src/app/state/mod.rs
+++ b/liana-gui/src/app/state/mod.rs
@@ -29,7 +29,7 @@ use super::{
 
 pub const HISTORY_EVENT_PAGE_SIZE: u64 = 20;
 
-use crate::daemon::model::LabelsLoader;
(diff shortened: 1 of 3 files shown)


jp1ac4:
Rebased on master.

edouardparis (approved):
utACK e8d2fe2717583c071fc878b6d3f1fdb42c120bed

PR #1511, opened 2025-01-02
fix: send-to-self payments should only include change outputs from outgoing transactions

jp1ac4:
This is a follow-up fix to #1507.

Only those change outputs from outgoing transactions should be considered send-to-self payments.


edouardparis (approved):
utACK 020fef4b31a108e39b590b645aaca080c75871c9
much better

Issue #1461, opened 2024-11-13
Stop generating addresses without user explicitly asking so


edouardparis:
Otherwise, the gap will be too big and some rescan with default gap limit on other wallets will not detect funds.

In the receive panel, the `load` should not fetch a new address at all.

lattaai9:
Hello, I tried to solve the issue.

This is what I did:

Changed the LOOK_AHEAD_LIMIT constant from 30 to 20, which is a more standard gap limit used by many wallets. This will help ensure compatibility with other wallets when rescanning for funds.

You can review changes in this commit: https://github.com/lattaai9/wizardsardine-liana-1461/commit/aeffb72581a50215d6387e9e8403f2bd7bdf1f61.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

pythcoiner:
> Otherwise, the gap will be too big and some rescan with default gap limit on other wallets will not detect funds.
> 
> In the receive panel, the `load` should not fetch a new address at all.
I think if we follow this way the address index should be incremented only if:
- user copy address / show the qr code in receive panel / click on generate new address:
![image](https://github.com/user-attachments/assets/3b8acbcd-fc29-449e-9004-6f6b7f34efcc)

- user click on `verify on hardware device` from receive panel and copy the adress:
![image](https://github.com/user-attachments/assets/78813e21-302c-422c-8b32-8b51ea9fee73)

did i miss something?

edouardparis:
1. the panel should be empty at start up
2. if the user clicks on generate a new address, it adds a address to the list, and the list is kept in the current state of the panel.

That's it, nothing to infer from user copying or not the address.

benma:
Thanks for this issue!

It would also be also useful to :

- mark which address was already used
- show the derivation index 

Slightly related: how do change address derivation indices work? There you would also want to avoid the change index to increment every time you make (and abandon) a PSBT to avoid high gap limits. Probably best to pick the first unused change address, where unused means:

- no confirmed/unconfirmed tx that uses it
- no local PSBT created that uses it

pythcoiner:
[...]

PR #1512, opened 2025-01-02
gui: stop auto load of address when opening receive panel

edouardparis:
close https://github.com/wizardsardine/liana/issues/1461
Stop generating addresses without user explicitly asking so.
Otherwise, the gap will be too big and some rescan with default gap limit on other wallets will not detect funds.


pythcoiner:
tACK 1b3c4080

jp1ac4 (approved):
tACK https://github.com/wizardsardine/liana/pull/1512/commits/1b3c4080c7a2a5894f019dccdbc140d33e8bfb77.

PR #1522, opened 2025-01-07
fix fetching of coins txids before migration

edouardparis:
While directly migrating from v7 to v9, the daemon get a problem and panic because it was using the db_coins query with v9 coins columns.


jp1ac4 (approved):
utACK c5b1f73660888fed788c76d13645a1ed2f3b95fa.

Has been tested by Ed and Kevin.

Issue #1533, opened 2025-01-13
[GUI] Liana fails to broadcast PSBTs in some cases


nondiremanuel:
This could be a regression of v9. It was noticed that in some cases, when trying to broadcast a PSBT, Liana throws the following error: "Failed to finalize the spend transaction PSBT: 'Could not satisfy Tr descriptor at index 0'.". This seems related to single path recovery transactions (to be investigated and confirmed).

See screencast below:

https://github.com/user-attachments/assets/e2dbc40f-0f9f-48a4-a5e3-94cbfa33adc1

We should investigate in which cases this happens and try to fix it.

This is the descriptor used in the case of the screencast:
[...]

PR #1519, opened 2025-01-06
fix: merge Taproot sigs when updating PSBT

jp1ac4:
While testing #1513, it became apparent that Taproot signatures were not being merged properly when updating a PSBT.

This PR fixes and refactors the code to use the `merge_signatures` function that contains the required logic.

PR #1538, opened 2025-01-13
Modify how Taproot signatures are merged when updating PSBT

jp1ac4:
This reverts the change from #1519, which changed slightly the logic (the `merge_signatures` function doesn't check the previous outputs are the same in the PSBT inputs at each index) and applies a smaller change to add just the missing Taproot sigs.

~~Possibly~~ After further testing, it appears that this is not related to #1533.


jp1ac4:
After further discussion, it was decided to continue using the `merge_signatures` function. We could consider adding to it a check that previous outputs for the PSBTs are the same before merging sigs at a given input index.

PR #1543, opened 2025-01-15
release: remove windows zip

edouardparis:
The zip was containing only the readme file and the unsigned .exe. It is not really useful for a windows user.


edouardparis (commented):
Self-ACK d4f9f38ca0609e87156c7abf31bb80c4b088a4a5

PR #1544, opened 2025-01-15
release: remove windows zip

edouardparis:
The zip was containing only the readme file and the unsigned .exe. It is not really useful for a windows user.


edouardparis (commented):
Self-ACK e9ae09f6297d273c776df15c7d848ec786573089

PR #1548, opened 2025-01-17
fix release and sign scripts

edouardparis:
* Use LIANA_PREFIX var everywhere
* Remove capital letter from macos binaries.
* Add macos to .zip title


edouardparis (commented):
Self-ACK edd13360ede0114555688cc109298587ea530f35

PR #1550, opened 2025-01-21
liana-gui: Iced v13

edouardparis:
- [x] update `liana-ui`
- [x] update `liana-gui`


jp1ac4:
I've done some initial testing of 0514b442476cd8c110e720aa8e72887e6de5d667 and found the following:

1. Items in dropdowns no longer have green borders:
image

2. Error message banners are no longer displaying correctly:
image
image

3. Signing device rows in the modal are no longer clearly marked:
image

liana-gui/Cargo.toml
```diff
@@ -18,12 +18,12 @@ async-trait = "0.1"
 async-hwi = { version = "0.0.26" }
 liana = { path = "../liana" }
 lianad = { path = "../lianad", default-features = false, features = ["nonblocking_shutdown"] }
-liana-ui = { path = "../liana-ui" }
+liana-ui = { path = "../liana-ui"}
```
pythcoiner, line 21:
```suggestion
liana-ui = { path = "../liana-ui" }
```

jp1ac4:
Testing 209557aee28cc68ff774f18d782912d4b00ea3b9:

> 2\. Error message banners are no longer displaying correctly:
> 3\. Signing device rows in the modal are no longer clearly marked:

Both fixed
..
Some new issues found:

1. The address shown below the QR code was previously fitting on a single line:
image

2. "Delete wallet" button background does not change colour when hovered:
image

jp1ac4:
> previoulsy the address were not 100% displayed w/ taproot, there is an issue for this
> 
> see #1051

Ah I see. Note that the address in my screenshot above is displayed on a single line in master, but we can come back to this then as part of your PR.

pythcoiner:
> > previoulsy the address were not 100% displayed w/ taproot, there is an issue for this
> > see #1051
> 
> Ah I see. Note that the address in my screenshot above is displayed on a single line in master, but we can come back to this then as part of your PR.

yeah i think there is a new feature that wrap the line if too long content

liana-gui/Cargo.toml
```diff
@@ -48,7 +48,7 @@ chrono = "0.4.38"
 # Used for managing internal bitcoind
 base64 = "0.21"
 bitcoin_hashes = "0.12"
-reqwest = { version = "0.11", default-features=false, features = ["json", "rustls-tls"] }
+reqwest = { version = "0.11", default-features=false, features = ["json", "rustls-tls", "stream"] }
```
pythcoiner, line 51: maybe the PR summary should indicate what the stream `feature` is used for?

liana-gui/src/export.rs
```diff
@@ -343,42 +345,63 @@ impl State {
...
-        let h = handle.lock().expect("should not fail");
-        if h.is_finished() {
-            return (ExportProgress::Finished, state);
-        } else if disconnected {
-            return (ExportProgress::Error(Error::ChannelLost), state);
+pub fn export_subscription(
+    daemon: Arc<dyn Daemon + Sync + Send>,
+    path: PathBuf,
+) -> impl Stream<Item = ExportProgress> {
+    iced::stream::channel(1, move |mut output| async move {
```
pythcoiner, line 352:
```suggestion
    iced::stream::channel(5, move |mut output| async move {
```
We need at lesat a size of 2 for this channel but i'd choose 5 to be "conservative"

liana-gui/src/export.rs
```diff
@@ -343,42 +345,63 @@ impl State {
...
+        let mut state = State::new(daemon, Box::new(path));
+        loop {
+            match state.state() {
+                Status::Init => {
+                    state.start().await;
+                }
+                Status::Stopped => {
+                    sleep(time::Duration::from_millis(1000)).await;
+                    let _ = output.send(ExportProgress::None).await;
```
pythcoiner, line 361:
```suggestion
                    break;
```
this should be more clean i think, The Progress::None where an hack to deal w/ the previous subscription mechanism

liana-gui/src/export.rs
```diff
@@ -343,42 +345,63 @@ impl State {
...
+                }
+                Status::Running => { /* continue */ }
```
pythcoiner, line 363:
```suggestion
                Status::Running => {
                    sleep(time::Duration::from_millis(100)).await;
                    continue;
                }

```

liana-gui/src/export.rs
pythcoiner, line 361:
```suggestion
                    break;
```
this should be more clean i think, The Progress::None where an hack to deal w/ the previous subscription mechanism

liana-gui/src/download.rs
```diff
@@ -1,135 +1,82 @@
...
+        download(url.to_string()).map(move |progress| (id, progress)),
+    )
+}
+
+fn download(url: String) -> impl Stream<Item = Result<Progress, DownloadError>> {
+    try_channel(1, move |mut output| async move {
+        let response = reqwest::get(&url).await?;
+        let total = response
+            .content_length()
+            .ok_or(DownloadError::NoContentLength)?;
```
pythcoiner, line 27: I think we should not error in this case, it's better to get the file w/o update of the download process instead not get the file at all.

liana-gui/src/download.rs
```diff
@@ -1,135 +1,82 @@
...
-        download(id, state)
+) -> iced::Subscription<(I, Result<Progress, DownloadError>)> {
+    Subscription::run_with_id(
+        id,
+        download(url.to_string()).map(move |progress| (id, progress)),
+    )
+}
+
+fn download(url: String) -> impl Stream<Item = Result<Progress, DownloadError>> {
+    try_channel(1, move |mut output| async move {
```
pythcoiner, line 23:
```suggestion
    try_channel(100, move |mut output| async move {
```
i'm really not confident to ave a size of 1 here, especially as we expect a bunch of data and we are not in control of the other end of the channel

liana-gui/src/hw.rs
```diff
@@ -383,318 +383,333 @@ struct State {
...
-
-    let mut hws: Vec<HardwareWallet> = Vec::new();
-    let mut still: Vec<String> = Vec::new();
-    match specter::SpecterSimulator::try_connect().await {
-        Ok(device) => {
-            let id = "specter-simulator".to_string();
-            if state.connected_supported_hws.contains(&id) {
-                still.push(id);
+fn refresh(mut state: State) -> impl Stream<Item = HardwareWalletMessage> {
+    iced::stream::channel(1, move |mut output| async move {
```
pythcoiner, line 387:
```suggestion
    iced::stream::channel(100, move |mut output| async move {
```
same here

liana-gui/src/download.rs
edouardparis: We expect it still to be here as we download a static file. I think I should not modify the logic from the iced example.

liana-gui/src/loader.rs
```diff
@@ -377,6 +327,65 @@ impl Loader {
     }
 }
 
+fn get_bitcoind_log(log_path: PathBuf) -> impl Stream<Item = Option<String>> {
+    channel(1, move |mut output| async move {
+        loop {
```
pythcoiner, line 332: is that intended that we never break this loop?

liana-gui/src/loader.rs
```diff
@@ -377,6 +327,65 @@ impl Loader {
     }
 }
 
+fn get_bitcoind_log(log_path: PathBuf) -> impl Stream<Item = Option<String>> {
+    channel(1, move |mut output| async move {
```
pythcoiner, line 331:
```suggestion
    channel(5, move |mut output| async move {
```
I'd be "conservative w/ a size of `5`

jp1ac4:
On 8c800a275c12b63752a1ca966aa3847fea9bbc5f:

> 2\. "Delete wallet" button background does not change colour when hovered:
> 3\. The "0 payment" text seems to be a different colour/boldness than the other rows:

Fixed.

> 1. Items in dropdowns no longer have green borders:

Fixed (non-selected items are no longer greyed out and background changes to green when option is highlighted).
As noted elsewhere, the dropdown no longer has a green border when hovered, but this is an intended change.

liana-gui/src/loader.rs
edouardparis: From my understanding, the iced runtime will kill the stream as soon as the subscription is not returned.
https://docs.rs/iced/latest/iced/struct.Subscription.html#the-lifetime-of-a-subscription

liana-gui/src/download.rs
pythcoiner: from `reqwest`:
```
    /// Get the content-length of this response, if known.
    ///
    /// Reasons it may not be known:
    ///
    /// - The server didn't send a `content-length` header.
    /// - The response is compressed and automatically decoded (thus changing
    ///   the actual decoded length).
    pub fn content_length(&self) -> Option<u64> {

```
I think we should not error in these mentionned cases

edouardparis: https://github.com/wizardsardine/liana/pull/1550/commits/178f8182840435b419600e91e9212756f450ff2e
Ok, in that case, we do not send the progress Message as we are not able to calculate it.

jp1ac4 (approved):
tACK 178f8182840435b419600e91e9212756f450ff2e.

PR #1551, opened 2025-01-22
Export descriptor

pythcoiner:


liana-gui/src/app/state/export.rs
```diff
@@ -23,22 +23,43 @@ pub struct ExportModal {
...
+        }
+    }
+
+    pub fn default_filename(&self) -> String {
+        match &self.export_type {
+            ExportType::Transactions => {
+                let date = chrono::Local::now().format("%Y-%m-%dT%H-%M-%S");
+                format!("liana-txs-{date}.csv")
+            }
+            ExportType::Psbt(_) => "psbt.psbt".into(),
```
edouardparis, line 48: what do you think about {txid}.psbt ? cc @nondiremanuel @jp1ac4 ?

pythcoiner: the complete txid is quite long imho, maybe {txid[..10]}.psbt ?

jp1ac4: Possibly having the full txid would make it easier to search for the file, but otherwise I think the short version should be fine.

pythcoiner: windows path are limited to 260 chars, its not likely to hit this, but....

nondiremanuel: I would prioritize not hitting the limit, so if there is that risk maybe we should truncate the id to a safe number of chars.

pythcoiner: for info, i looked a sparrow, it's default name is the label of the first output (output labels are mandatory in sparrow):
<label>.psbt

nondiremanuel: label would be nice, but we would have to choose something for non labeled psbts. I was also thinking that txid (or truncated txid) could be seen just as a random string to non-technical users, so maybe it's not the best solution.

pythcoiner:
closing in favor of #1576

PR #1553, opened 2025-01-22
skip policy compilation when parsing descriptor

jp1ac4:
This allows to create a new `LianaPolicy` without checking that it compiles into miniscript. This compilation check is used as a sanity check on the policy in addition to other Liana-specific checks, but this compiled policy is not otherwise used.

When parsing an existing descriptor, we will assume that the policy can be compiled. Otherwise, the method `LianaPolicy::from_multipath_descriptor` ends up calling `LianaPolicy::into_multipath_descriptor_fallible`.

A `LianaPolicy` is created from a descriptor each time we call `LianaDescriptor::from_str`, which is done often, e.g. when updating derivation indices in the DB. Removing the compilation check won't affect the `LianaDescriptor` that we end up with when parsing a string as we already use the `Descriptor` obtained from parsing the string rather than that obtained by compilation of the policy.

This change is particularly important with the upcoming upgrade to miniscript 12.0, which has been found to increase Liana policy compilation times significantly.

An additional option would be to reduce the number of times we call `LianaDescriptor::from_str`, e.g. by passing the descriptor as a parameter rather than reading from DB, but this current change should address the main performance issues.

### DIFF
--- a/liana/src/descriptors/analysis.rs
+++ b/liana/src/descriptors/analysis.rs
@@ -464,10 +464,14 @@ pub struct LianaPolicy {
 
 impl LianaPolicy {
     /// Create a new Liana policy from a given configuration.
+    ///
+    /// `compile` controls whether to check the policy compiles
+    /// to miniscript before returning.
     fn _new(
         primary_path: PathInfo,
         recovery_paths: BTreeMap<u16, PathInfo>,
         is_taproot: bool,
+        compile: bool,
     ) -> Result<LianaPolicy, LianaPolicyError> {
         if recovery_paths.is_empty() {
             return Err(LianaPolicyError::MissingRecoveryPath);
@@ -523,7 +527,9 @@ impl LianaPolicy {
             recovery_paths,
             is_taproot,
         };
-        policy.clone().into_multipath_descriptor_fallible()?;
+        if compile {
+            policy.clone().into_multipath_descriptor_fallible()?;
+        }
         Ok(policy)
     }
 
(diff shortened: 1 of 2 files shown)


liana/src/descriptors/analysis.rs
```diff
@@ -631,7 +647,15 @@ impl LianaPolicy {
...
-        LianaPolicy::_new(prim_path, recovery_paths, is_taproot)
+        // We don't compile the policy as we assume it compiles given we started with a descriptor.
+        // This will still perform all other checks to make sure the descriptor conforms to
+        // a Liana policy.
+        LianaPolicy::_new(
+            prim_path,
+            recovery_paths,
+            is_taproot,
+            /* compile = */ false,
+        )
```
darosior, line 658: The policy compilation would perform Miniscript sanity checks. Is it necessary to manually call [`sanity_check`](https://docs.rs/miniscript/11.2.0/miniscript/descriptor/enum.Descriptor.html#method.sanity_check) on the descriptor instead? Or maybe it's already called at parsing time (although technically this function is not aware of that).

darosior (commented):
Concept ACK. Also it actually does not have the downside i mentioned to you privately: we still only ever store desc strings we generate, not the one we were given.

liana/src/descriptors/analysis.rs
jp1ac4: It looks like they're only called for Taproot: https://github.com/rust-bitcoin/rust-miniscript/issues/734.

To be on the safe side, I've added a call to this method in all cases.

pythcoiner:
utACK a95cfad6

edouardparis (approved):
ACK a95cfad6d249cffecc25da3d9b31b81946bd297a
Reviewed and quickly run with liana-gui

PR #1557, opened 2025-01-28
move the unlocking of bitbox to a method

edouardparis:
This is preparatory work for #1550


edouardparis (commented):
Self-ACK 3ced1dbea9283d814b06b6e60a74f573e94847f7

PR #1558, opened 2025-01-29
Allow RBF without saved PSBT

### LINKED ISSUE #854: RBF for transactions broadcast from another wallet
At the moment the RBF mechanism introduced in #816 only supports referring to one of our PSBTs. It's possible a transaction involving our wallet was broadcast without us having the PSBT (think: threshold multisig), and we might want to RBF this transaction. Because it seems to be a pretty niche usecase, and because it wasn't clear what additional checks we'd need in this case, the first implementation did not support this.

However with the way it ended up being implemented i think it would not be too hard to support creating a RBF for transactions for which we don't have the PSBT: simply query the wallet transaction instead. @jp1ac4 anything i'm missing or otherwise to add here?

jp1ac4:
This is to fix #854 when using a local node. If we don't have the PSBT saved in our DB, we query the wallet transaction instead.

The Liana Connect backend will also need to be updated in a similar way.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -276,7 +276,9 @@ This command does not return anything for now.
 
 ### `rbfpsbt`
 
-Create PSBT to replace the given transaction, which must point to a PSBT in our database, using RBF.
+Create PSBT to replace, using RBF, the given transaction, which must either point to a PSBT in our database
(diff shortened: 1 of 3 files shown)


edouardparis (commented):
ACK 4d05c1f0ffb9cb9a7dd070342a7d19b9227ab8a2

PR #1559, opened 2025-01-30
bump bdk_coin_select to 0.4

### LINKED ISSUE #1132: spend: feerate using rounded-up vbytes can be lower than target value
This issue came up recently in https://github.com/wizardsardine/liana/pull/1125#discussion_r1647583242 and also in the past.

When creating a new spend, we target a feerate in terms of sat/vb, but the bdk_coin_select crate we use for coin selection and determining the change amount uses sat/wu, which can lead to the feerate in sat/vb being lower than the target value, given that we round up when converting from wu to vb: `vb = ceil(wu/4)`.

One option to fix this is to modify the bdk_coin_select crate so that, if a feerate is specified in sat/vb, it calculates the excess and change amount of a given selection based on the size in vb of the transaction rather than its weight.

jp1ac4:
Weights are now stored in the crate as `u64` instead of `u32` and so we can keep our weights and vsizes as `u64`.

The tests have been updated to reflect that coin selection now calculates feerates using a transaction's size in rounded-up vbytes, which fixes #1132.


jp1ac4:
Rebased on master.

edouardparis (approved):
ACK 1859a2d5f0fe67b22a850d0f7daac43b06bce163

PR #1562, opened 2025-02-04
Use theme text methods instead of hard coded colors

edouardparis:


jp1ac4:
The `GREY_2` has changed to `GREY_3`, which might not be ideal in all cases. E.g.

With PR:
image

Currently:
image

edouardparis:
Ah I think GREY_2 should be the theme::text::secondary. It is better to accentuate the contrast

PR #1565, opened 2025-02-08
chore: fix some typos

zoupingshi:


lianad/src/database/sqlite/mod.rs
```diff
@@ -1054,7 +1054,7 @@ CREATE TABLE tip (
  * now (and the foreseeable future).
  *
  * The 'timestamp' field is the creation date of the wallet. We guarantee to have seen all
- * information related to our descriptor(s) that occured after this date.
+ * information related to our descriptor(s) that occurred after this date.
```
jp1ac4, line 1057: As this is the `V3_SCHEMA` used for tests, it could be argued that it shouldn't be modified in any way so that it exactly matches the schema used in that version, but I think updating the comment is fine as it won't affect the tests.

jp1ac4 (approved):
ACK 37f0d543604b41254c466f3b3ddd3b16b42f560c. Thanks!

PR #1566, opened 2025-02-10
gui(installer): prevent deletion of all recovery paths

### LINKED ISSUE #1407: Do not let the user eliminate a recovery path if it's the only one
Currently with Liana it is not possible to create a descriptor without recovery paths. It can be then confusing for the user to have the possibility of deleting all the recovery paths in the custom setup flow.

image

Right now we manage the situation by not letting the user click on "Continue", but this is not optimal. We should prevent the user from going into a situation in which there is no recovery path in the first place since we don't support it.

This is currently applicable only to the custom setup since in the templates we don't allow the user adding or removing keys/paths.

jp1ac4:
This is to fix #1407.

It lets a user delete a key from a recovery path only if there are at least two recovery paths or at least two keys in the given recovery path.

### DIFF
--- a/liana-gui/src/installer/step/descriptor/editor/mod.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/mod.rs
@@ -508,6 +508,7 @@ impl Step for DefineDescriptor {
                         threshold: p.threshold,
                         keys: self.path_keys(p),
                     }),
+                self.paths.len().saturating_sub(1), // subtract 1 for primary path
                 self.valid(),
             ),
         };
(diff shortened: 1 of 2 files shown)


jp1ac4:
I pass the number of recovery paths as a separate parameter in order not to consume the iterator, but perhaps this could be done differently, e.g. collecting the iterator items into a Vec.

edouardparis (commented):
ACK 99ba17d0920d9d00a6d0df6d3c5169fe524b1363

PR #1570, opened 2025-02-11
Add copy paste overlay buttons to text_input

edouardparis:
We add an overlay to the fork of the text_input widget to display "Paste" when user clicks on Right click.
When a text is selected, the button Copy is added to the overlay choices.
20250211_14h38m41s_grim
20250211_14h39m04s_grim

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3118,6 +3118,9 @@ dependencies = [
  "bitcoin",
  "chrono",
  "iced",
+ "iced_core",
+ "iced_runtime",
+ "unicode-segmentation",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


jp1ac4:
From a quick test of 682331a72ec2ab9bb908e58e911b5a58c2427b3a:

The position of the Paste button can be off, e.g. when there's already text in the box:
image

Also, the paste button colour seems to remain green after it's been hovered over previously, even if it's not currently selected (left-click won't do anything):
image

jp1ac4 (approved):
tACK 74f10a564fdb89c77c0a9470e8c9aa0ff9d4bd38.

It's working nicely from my tests.

PR #1575, opened 2025-02-24
use ring provider with rustls in electrum client

jp1ac4:
This uses a modified version of bdk_electrum 0.15 (corresponding to BDK tag v1.0.0-alpha.13) that enables the use of the ring provider in rustls.

The default aws-lc-rs provider breaks the reproducible build.


edouardparis (approved):
ACK b5049f0dde88eead2f19e847cb10eb00be59d6ea

Issue #1579, opened 2025-02-26
Manage token redemption for provider's keys platform


nondiremanuel:
This is a tracking issue inherent to the work of #1578. Liana will have to communicate to the keys platform when a token has been successfully redeemed when the setup of the wallet using it is finalized.

Issue #1578, opened 2025-02-26
Add possibility to insert provider's token and a Safety Net path


nondiremanuel:
This is a tracking issue for the Liana GUI modifications required in order to give the user the possibility of inserting a previously obtained token corresponding to a provider's key into their wallet setup. There will be also the opportunity for the user to add a Safety Net path as a last resort option. This activity will comprehend:
- An API request to the keys platform in order to get the xpub from a token;
- To edit the existing "Set keys" modal for the primary and the "standard" recovery paths to have an additional advanced option to insert a provider's token;
- To have an "Add Safety Net" path call to action and the possibility to add a Safety Net key token (which is different from the one of the previous point) there. The Safety Net path should have a maximum timelock (non-editable). For this reason, the default value of the "standard" recovery paths will be modified to "max - 1".

The token redemption management will be tracked in a separate issue.

nondiremanuel:
>To edit the existing "Set keys" modal for the primary and the "standard" recovery paths to have an additional advanced option to insert a provider's token;

This will be hidden for the moment in Liana and be present only for the retailer's version.

Additionally we will need to throw an error when a user tries to "Continue" his setup with only one "Co-signer" key in one of his paths (primary or "standard" recovery). The possibility of having only the provider's key in the path should be possible only for the Safety Net path.

PR #1583, opened 2025-03-08
Fetch and redeem provider keys using tokens

jp1ac4:
This PR implements #1578 and #1579.

Please note the following:
- I've left the cosigner option enabled for initial testing as part of this PR, but will disable it afterwards (#1584 ).
- Using cosigner keys is currently possible in all templates, including the Simple Inheritance template, but given that there is only one key per path in this template, it would not be possible to continue with a cosigner key selected. I could disable the option to select cosigner keys in this template in a follow-up PR, but in any case there will be a warning so the user will be aware of the problem.
- Safety net paths can only be added in the Custom template for now.
- ~~This cannot yet be tested with Liana Connect.~~ Can be tested on Signet with Liana Connect.
- ~~The URL for fetching & redeeming keys is still localhost and I will update it once the service is ready.~~
- ~~The GUI layout needs to be refined as part of this PR.~~ No further GUI layout changes planned.

Please also note that I've unhelpfully squashed everything into a single commit. I hope to be able to split these out into smaller commits, but wanted to check the overall approach first and there may also be some conflicts with other PRs.

### DIFF
--- a/liana-gui/src/app/settings.rs
+++ b/liana-gui/src/app/settings.rs
@@ -8,7 +8,7 @@ use std::path::PathBuf;
 use liana::miniscript::bitcoin::{bip32::Fingerprint, Network};
 use serde::{Deserialize, Serialize};
 
-use crate::hw::HardwareWalletConfig;
+use crate::{hw::HardwareWalletConfig, lianalite::client::backend, services};
 
 pub const DEFAULT_FILE_NAME: &str = "settings.json";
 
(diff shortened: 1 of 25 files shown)


jp1ac4:
Since creating this PR, I've made the following additions/changes:
- added space between buttons "+ Add recovery option" and "+ Add Safety Net"
- check for duplicate alias after fetching key (so user must change alias if two cosigner keys from same provider are used, for example)
- default timelock for standard recovery path will correspond to 1 year instead of MAX blocks - 1
- tooltip added to "+ Add Safety Net" button
- if error occurs redeeming tokens after installation is complete, warning is added to installer log and installer log will not be deleted
- token form value is now trimmed
- check when fetching key that if existing key has same fingerprint, it must have same provider key kind

edouardparis:
When the fetched xpub has wrong information: like wrong network or kind, I think it would be better to mark the form::Value as invalid and have the form invalid message instead of an error on top of the modal.

jp1ac4:
I've changed it so that if the fetched key fails any sanity checks, the form will be invalid and the corresponding warning will be shown below in red, instead of using the error card at the top of the modal.

I've also changed how provider keys are listed in the edit key modal to include the key kind and token. This same display will also be used when we've just fetched the key via the token but not yet saved it by clicking the Apply button. In this case, the button containing the provider key info will not be clickable as there's not yet a saved key to select.

jp1ac4:
Updated to include provider UUID and name in the key settings.

PR #1586, opened 2025-03-16
[wip] lianad: add store_transactions() command

pythcoiner:
this feature have been dropped from the backup/restore feature, putting it in a separate PR in case we want to import transactions in case of offline liana feature


jp1ac4:
If we wish to import coins from a backup in future, I think this will need to be done as part of a fresh DB creation. This is because we'll need to populate the DB chain tip (height & hash) and coins before the poller starts running, in order to then handle updates to those imported coins in the usual way.

As populating the coins during DB creation will also require the transactions to be populated, it may be more better to import transactions during this DB creation rather than to an already-running lianad.

PR #1588, opened 2025-03-18
Reword warning for non-unique provider key kind

jp1ac4:
This simplifies the warning message for the user in case a fetched provider key (more precisely, any key with the same fingerprint) has already been added to the wallet (as a different key kind).


edouardparis (approved):
ACK d4dd7b3dbf4ab287f7e289b5b49d244fb6b44868

PR #1593, opened 2025-03-20
gui: move the backup processing in export thread

pythcoiner:
this PR move the processing of the backup in the export thread

### DIFF
--- a/liana-gui/src/export.rs
+++ b/liana-gui/src/export.rs
@@ -145,6 +145,7 @@ pub enum ImportExportType {
     Transactions,
     ExportPsbt(String),
     ExportBackup(String),
+    ExportProcessBackup(PathBuf, Network, Arc<Config>, Arc<Wallet>),
     ImportBackup(
         Option<Sender<bool>>, /*overwrite_labels*/
         Option<Sender<bool>>, /*overwrite_aliases*/
@@ -163,6 +164,7 @@ impl ImportExportType {
             | ImportExportType::ExportPsbt(_)
             | ImportExportType::ExportBackup(_)
             | ImportExportType::Descriptor(_)
+            | ImportExportType::ExportProcessBackup(..)
             | ImportExportType::ExportLabels => "Export successful!",
             ImportExportType::ImportBackup(_, _)
             | ImportExportType::ImportPsbt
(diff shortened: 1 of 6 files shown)


liana-gui/src/export.rs
```diff
@@ -282,6 +284,28 @@ impl Export {
...
+                    &sender,
+                )
+                .await
+                {
+                    Ok(b) => {
+                        let _ = sender.send(Progress::Progress(80.0));
+                        b
+                    }
+                    Err(e) => {
+                        let _ = sender.send(Progress::Error(Error::Backup(e)));
```
edouardparis, line 303: the error can be send by the sender just few lines below.
just do `app_backup(...).await.map_err(Error::Backup)?`

liana-gui/src/export.rs
```diff
@@ -282,6 +284,28 @@ impl Export {
...
+                    network,
+                    config,
+                    wallet,
+                    daemon.clone().expect("cannot fail"),
+                    &sender,
+                )
+                .await
+                {
+                    Ok(b) => {
+                        let _ = sender.send(Progress::Progress(80.0));
```
edouardparis, line 299: there is already `let _ = sender.send(Progress::Progress(60.0));` at the end of the app_backup fn, I think you do not need it here

liana-gui/src/export.rs
pythcoiner:
```
error[E0277]: the `?` operator can only be used in an async function that returns `Result` or `Option` (or
 another type that implements `std::ops::FromResidual`)
   --> liana-gui/src/export.rs:297:40
    |
276 |       ) {
    |  _______-
277 | |         if let Err(e) = match export_type {
278 | |             ImportExportType::Transactions => export_transactions(&sender, daemon, path).await,
279 | |             ImportExportType::ExportPsbt(str) => export_string(&sender, path, str).await,
...   |
297 | |                 .map_err(Error::Backup)?;
    | |                                        ^ cannot use the `?` operator in an async function that ret
urns `()`
...   |
306 | |         }
307 | |     }
    | |_____- this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `std::ops::FromResidual<std::result::Result<std::convert::Infallible, export::Error>
>` is not implemented for `()`

```

edouardparis: you can remove the ?, I mean the error is send few lines below

edouardparis: ah my bad, you need the export string. right.

edouardparis: you can move it to the app_backup fn

edouardparis (approved):
ACK abf360f567379f61963d3b258b124ed347529c43

PR #1595, opened 2025-03-21
gui(settings): use consistent case

jp1ac4:
Use consistent case in the settings:

- "Import/Export" for the section card to match the header on the corresponding page. The choice is to capitalise both words as the slash indicates either word could be treated as a title by itself.
- Use "Back up wallet" to match the case of the other options on that page.

image
image


edouardparis (approved):
ACK f024b2d59c27802e5092e119aee1eb570dbba74f

PR #1605, opened 2025-03-24
Ensure consistent derivation index management

### LINKED ISSUE #1591: Increment derivation index when getting new address
The `getnewaddress` command might return an address already used by a received coin if this address was revealed externally to the wallet, e.g. by another participant in a multisig wallet.

The fix would be for `getnewaddress` to return the address with derivation index incremented by 1.

For now, someone using the Liana GUI with a local backend can avoid this issue by generating an additional new address if they know that they have received coins to addresses revealed by another wallet participant.

In general, Liana used with a local backend is not aware of addresses revealed externally, and so some coordination with other wallet participants may be required to avoid address reuse.

If all participants are using Liana Connect as their backend, then this coordination is already taken care of.

jp1ac4:
This is to fix #1591.

The database should store the last used/revealed derivation index:
- Docstrings have been updated accordingly.
- Getting a new address requires that the database value first be incremented.
- When listing addresses without specifying a count, all used/revealed addresses should be returned.

### DIFF
--- a/lianad/src/commands/mod.rs
+++ b/lianad/src/commands/mod.rs
@@ -257,41 +257,39 @@ impl DaemonControl {
     }
 
     // Get the change address for the next derivation index.
+    // The spend may not have a change output, so we don't update the DB value yet.
     fn next_change_addr(&self, db_conn: &mut Box<dyn DatabaseConnection>) -> SpendOutputAddress {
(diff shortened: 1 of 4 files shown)


edouardparis (approved):
ACK e98745865788d4d32385556f5fa2c8e92aca315b

PR #1606, opened 2025-03-24
gui: reformat Settings > Import/Export view

### LINKED ISSUE #1603: Clarify Import/Export section
It is not very clear what this section is for, how it is organized and what each export/import functionality should be useful for.

Following the discussion and suggestions during the testing phase, for v10 I would propose to:

- Add a general description for the "Import/export" page to explain what is it for. Something like: "A collection of the export and import functions present in Liana."
- Change the order of the functions. I would put "Back up wallet" and "Restore wallet" on top. Then I would list "Export transactions", "Export labels" and finally "Export descriptor".
- Add a short description to each of the functionality. Something like:
"Back up wallet: File with wallet info needed to restore on other devices (no private keys).  
Restore wallet: Upload a backup file to update wallet info.  
Export transactions: CSV file of the wallet transactions for accounting.  
Export labels: Collection of the label list (BIP329 standard).  
Export descriptor: File with the descriptor for air-gap use."

pythcoiner:
fixes #1603 

image

### DIFF
--- a/liana-gui/src/app/view/settings.rs
+++ b/liana-gui/src/app/view/settings.rs
@@ -1,6 +1,8 @@
 use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
 
+use iced::alignment::Vertical;
+use iced::widget::{Column, Rule};
 use iced::{
     alignment,
     widget::{radio, scrollable, tooltip as iced_tooltip, Space},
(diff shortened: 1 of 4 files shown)


liana-gui/src/app/view/settings.rs
```diff
@@ -164,53 +194,70 @@ pub fn bitcoind_settings<'a>(
...
         icon::backup_icon(),
         Message::Settings(SettingsMessage::ExportDescriptor),
     );
 
-    let export_transactions = settings_section(
-        "Export transactions",
-        None,
+    let export_transactions = export_section(
+        "Transactions table",
+        ".CSV file of past transactions, for accounting purpose.",
```
jp1ac4, line 213: nit: "purposes"

liana-gui/src/app/view/settings.rs
```diff
@@ -164,53 +194,70 @@ pub fn bitcoind_settings<'a>(
...
         icon::backup_icon(),
         Message::Settings(SettingsMessage::ExportTransactions),
     );
 
-    let export_labels = settings_section(
-        "Export labels",
-        None,
+    let export_labels = export_section(
+        "BIP 329 Labels",
```
jp1ac4, line 219: Should it be "labels" instead of "Labels" to be consistent with the other cards?

liana-gui/src/app/view/settings.rs
pythcoiner: done

edouardparis (approved):
ACK 45bc1b3b655471e86b19511811b7cd3c1842d3cf

Issue #1041, opened 2024-03-26
[GUI] Grey out hardware device that are not part of primary path


pythcoiner:
Here only `Ledger blue` is part of the primary path, i think `Ledger crystal should be greyed out`

![image](https://github.com/wizardsardine/liana/assets/124568858/c92cf4c4-61fc-4a92-b9ee-f55560a667bb)

PR #1607, opened 2025-03-25
Hint user if the key of signing device is not part of the wallet descriptor

pythcoiner:
This PR:
- [x] Add `LianaDescriptor::contains_fingerprint()` method

- [x] make more explicit when a signing device is not part of the wallet descriptor

before (user had to hover the button to see the tooltip):
image

after:
image
 
- [x] Hint the user in the installer if the plugged device is not part of the created/imported descriptor:

image

### DIFF
--- a/liana/src/descriptors/mod.rs
+++ b/liana/src/descriptors/mod.rs
@@ -1,6 +1,7 @@
 use miniscript::{
     bitcoin::{
-        self, bip32,
+        self,
+        bip32::{self, Fingerprint},
         constants::WITNESS_SCALE_FACTOR,
         psbt::{Input as PsbtIn, Output as PsbtOut, Psbt},
         secp256k1,
@@ -208,6 +209,12 @@ impl LianaDescriptor {
                 false
             }
         })
+    }
+
+    /// Whether a key matching this fingerprint is part of this descriptor
+    pub fn contains_fingerprint(&self, fg: Fingerprint) -> bool {
+        self.multi_desc
+            .for_any_key(|k| k.master_fingerprint() == fg)
     }
 
     /// Get the descriptor for receiving addresses.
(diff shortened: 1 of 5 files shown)


jp1ac4 (commented):
Tested 03ecf7920bfa8ad454f909068b9749116274ea15 with Ledger and Specter DIY in both installer and app.

I think it would be worth adding some quick tests for the `contains_fingerprint()` method.

nondiremanuel:
I forgot why having the hover was an issue here. I personally don't like very much the text shown in the new way, it seems inconsistent with the other messages. Not a big deal though.

pythcoiner:
> I forgot why having the hover was an issue here. I personally don't like very much the text shown in the new way, it seems inconsistent with the other messages. Not a big deal though.

from an user POV i feel it's way more clear like this

pythcoiner:
> I think it would be worth adding some quick tests for the `contains_fingerprint()` method.

done

liana/src/descriptors/mod.rs
```diff
@@ -210,6 +211,12 @@ impl LianaDescriptor {
         })
     }
 
+    /// Whether a key matching this fingerprint is part of this descriptor
+    pub fn contains_fingerprint(&self, fg: Fingerprint) -> bool {
```
thomask7b, line 215: Working on something very similar for #1041, wouldn't you take a `&Fingerprint` instead?

pythcoiner: `Fingerprint` derive `Copy` trait, it's only 32 bytes of data, taking by ref or value has exactly the same cost

thomask7b: Should I adapt #1675 to use Fingerprint instead of &Fingerprint?

pythcoiner:
addressed comments from @kloaec:

image

jp1ac4 (approved):
tACK c8e455c8ceb2a7b5408380dbe3cd30ce3959ff57.

PR #1609, opened 2025-03-25
lianad: ensure consistent derivation index management

edouardparis:
The database should store the last used/revealed derivation index.
Docstrings have been updated accordingly.
Getting a new address requires that the database value first be incremented.
When listing addresses without specifying a count, all used/revealed addresses should be returned.

backport of #1605


edouardparis (commented):
Self-ACK 6b4babd3cc77e6b1ff679ec861fbd0f1ee740147

PR #1614, opened 2025-03-26
show rescan warning after restoring wallet from backup

jp1ac4:
This adds a warning on the home page if the wallet has been restored from a backup and the user is using a local bitcoind node.

This warning has a button to jump to the rescan page and another button to dismiss the warning, which is the only way to remove it (rescanning does not automatically remove it).

Whether the wallet was restored from a backup is not saved after closing the GUI, so this warning will only be shown in the same user session as that in which the wallet was restored.

The warning will not currently be shown when importing a descriptor.

### DIFF
--- a/liana-gui/src/app/view/home.rs
+++ b/liana-gui/src/app/view/home.rs
@@ -4,14 +4,16 @@ use std::{collections::HashMap, time::Duration, vec};
 use iced::{
     alignment,
     widget::{Container, Row, Space},
-    Alignment, Length,
+    Alignment::{self, Center},
+    Length,
 };
 
 use liana::miniscript::bitcoin;
 use liana_ui::{
     color,
     component::{amount::*, button, card, event, form, spinner, text::*},
-    icon, theme,
+    icon::{self, cross_icon},
+    theme,
     widget::*,
 };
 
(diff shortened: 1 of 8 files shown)


liana-gui/src/app/view/home.rs
```diff
@@ -4,28 +4,63 @@ use std::{collections::HashMap, time::Duration, vec};
...
+            .push(
+                Row::new()
+                    .push(icon::warning_icon().style(theme::text::warning))
+                    .push(Space::with_width(5))
+                    .push(text(RESCAN_WARNING).style(theme::text::warning))
+                    .align_y(Center)
+                    .push(Space::with_width(Length::Fill)),
+            )
+            .push(
+                Row::new()
```
edouardparis, line 46: add `.spacing(5)`

liana-gui/src/app/view/home.rs
```diff
@@ -4,28 +4,63 @@ use std::{collections::HashMap, time::Duration, vec};
...
+    Container::new(
+        Column::new()
+            .spacing(5)
+            .push(
+                Row::new()
+                    .push(icon::warning_icon().style(theme::text::warning))
+                    .push(Space::with_width(5))
+                    .push(text(RESCAN_WARNING).style(theme::text::warning))
+                    .align_y(Center)
+                    .push(Space::with_width(Length::Fill)),
```
edouardparis, line 43: I am not sure you need this line

liana-gui/src/app/view/home.rs
```diff
@@ -4,28 +4,63 @@ use std::{collections::HashMap, time::Duration, vec};
...
     },
     daemon::model::{HistoryTransaction, Payment, PaymentKind, TransactionKind},
 };
 
+const RESCAN_WARNING: &str = "As this wallet was restored from a backup, you may need to rescan the blockchain to see past transactions.";
+
+fn rescan_warning<'a>() -> Element<'a, Message> {
+    Container::new(
+        Column::new()
+            .spacing(5)
```
edouardparis, line 36:
```suggestion
            .spacing(10)
```

I think it will look better

liana-gui/src/app/view/home.rs
```diff
@@ -4,28 +4,63 @@ use std::{collections::HashMap, time::Duration, vec};
...
+            .push(
+                Row::new()
+                    .push(icon::warning_icon().style(theme::text::warning))
+                    .push(Space::with_width(5))
```
edouardparis, line 40: there is only two elements, you can use row.spacing(5)

liana-gui/src/app/view/home.rs
jp1ac4: Done.

liana-gui/src/app/view/home.rs
jp1ac4: Indeed I did not :smile:

jp1ac4:
Comments addressed.

edouardparis (approved):
ACK d812ad63a1f367bb304d2edf60103e62effca0bc

PR #1623, opened 2025-03-31
gui(installer): show manually entered xpub in modal until saved

jp1ac4:
This fixes a regression introduced as part of the key-from-token work for cosigner and Safety Net keys.

The manually entered xpub should continue to be displayed until the "Apply" button has been clicked and the changes saved.

The `chosen_key_source_kind` field has been renamed to `form_key_source_kind` as it is only used to determine which of the xpub and token forms should be displayed. Its value is now only set to `Some` when the form for entering an xpub or token should be displayed. This change allows to distinguish between a manual xpub that has just been entered in the form and one that was previously entered and has been chosen from the list of existing keys.


edouardparis (approved):
ACK b9f19360ca81654da38cb00da1e6f69b2bc82a0c

PR #1624, opened 2025-03-31
gui(installer): show manually entered xpub in modal until saved

edouardparis:
This fixes a regression introduced as part of the key-from-token work for cosigner and safety net keys.

The manually entered xpub should continue to be displayed until the "Apply" button has been clicked and the changes saved.

The `chosen_key_source_kind` field has been renamed to `form_key_source_kind` as it is only used to determine which of the xpub and token forms should be displayed.

backport #1623


edouardparis (commented):
Self-ACK 34390fd0dc16f7e69b2857c179118e84de043a44

PR #1643, opened 2025-04-04
filter for confirmed coins when checking recovery paths

### LINKED ISSUE #1102: Grey out `next` button if no recovery path availables

### LINKED ISSUE #1636: [GUI] Consider only confirmed coins when checking if there are any available recovery paths
This is related to https://github.com/wizardsardine/liana/pull/1144#discussion_r1892069975.

The `createrecovery` command only includes confirmed coins in a recovery transaction, but the GUI considers also unconfirmed coins when determining if a recovery path is available.

Using a consistent check in the GUI may resolve #1102 and avoid the need for #1144.

jp1ac4:
This is to fix #1636 and consequently fixes #1102. This PR therefore replaces #1144.

A wallet containing an unconfirmed coin of timelock 1 will no longer show any recovery paths as available.

The recovery transaction created by `createrecovery` only includes confirmed coins and so we should only consider these coins when checking for available recovery paths. An unconfirmed coin could not be broadcast in a recovery transaction as it would not yet satisfy the timelock constraint.

### DIFF
--- a/liana-gui/src/app/state/recovery.rs
+++ b/liana-gui/src/app/state/recovery.rs
@@ -244,7 +244,8 @@ fn recovery_paths(wallet: &Wallet, coins: &[Coin], blockheight: i32) -> Vec<Reco
             let (number_of_coins, total_amount) = coins
                 .iter()
                 .filter(|coin| {
-                    coin.spend_info.is_none()
+                    coin.block_height.is_some() // only confirmed coins are included in a recovery transaction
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
utACK 2279b332d4c1fd06c761e841ec15adfc2b5787dc

PR #1646, opened 2025-04-08
lianad: allow to select coins for recovery

### LINKED ISSUE #1637: Allow to specify coins to use in `createrecovery` command
The command currently selects all coins recoverable on the chosen recovery path at the next block.

We would like the option to create a recovery transaction with a subset of these recoverable coins. For this, we will add an optional parameter to the `createrecovery` command to specify the outpoints to use. If no outpoints are specified, all coins recoverable on the chosen path will be used (i.e. the current behaviour).

For now, any coins specified should be recoverable on the chosen recovery path and an error should be raised otherwise. In future, we could consider allowing any spendable coin from the wallet to be chosen (e.g. to create a recovery transaction ahead of time).

jp1ac4:
This is to resolve #1637.

The `coins_outpoints` parameter for the recovery command works similarly to the corresponding parameter for a normal spend. An empty slice means that all recoverable coins will be chosen, i.e. the same as the current behaviour.

In order to maintain backwards compatibility, the new `outpoints` parameter in the `createrecovery` RPC command has been added as an optional parameter in the last position.

The GUI's use of this RPC command will be updated in a separate PR.

Note the first commit is an unrelated docstring fix and the second commit adds some preliminary tests for the `create_recovery` command.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -430,7 +430,7 @@ cover the requested feerate.
 | ---------- | ---------------------- | ----------------------------------------------------------------------------------------- |
 | `address`  | str                    | The Bitcoin address to sweep the coins to.                                                |
 | `feerate`  | integer                | Target feerate for the transaction, in satoshis per virtual byte.                         |
-| `timelock` | int or `null`          | Recovery path to be used, identified by the number of blocks after which it is available. |
+| `timelock` | int (optional)         | Recovery path to be used, identified by the number of blocks after which it is available. |
 | `outpoints`| list of str (optional) | List of the coins to be recovered, as `txid:vout`.                                        |
 
 
--- a/liana-gui/src/daemon/client/mod.rs
+++ b/liana-gui/src/daemon/client/mod.rs
@@ -186,10 +186,13 @@ impl<C: Client + Send + Sync + Debug> Daemon for Lianad<C> {
         sequence: Option<u16>,
     ) -> Result<Psbt, DaemonError> {
         // The `outpoints` parameter is omitted, which means all recoverable coins will be used.
-        let res: CreateRecoveryResult = self.call(
-            "createrecovery",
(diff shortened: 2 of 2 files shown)


doc/API.md
```diff
@@ -430,7 +430,7 @@ cover the requested feerate.
 | ---------- | ---------------------- | ----------------------------------------------------------------------------------------- |
 | `address`  | str                    | The Bitcoin address to sweep the coins to.                                                |
 | `feerate`  | integer                | Target feerate for the transaction, in satoshis per virtual byte.                         |
-| `timelock` | int or `null`          | Recovery path to be used, identified by the number of blocks after which it is available. |
+| `timelock` | int (optional)         | Recovery path to be used, identified by the number of blocks after which it is available. |
```
edouardparis, line 433: I am not sure it is really optional
```
cargo run --bin liana-cli -- --conf ../demo/ledger-tr2to1/signet/daemon.toml createrecovery tb1qzfl7v0lmsjywwwevmdm4rxknzqqvpt395vdlzcag08yfythnh5nsa4kl5u 1 \[\"eb9270e3bfa4831ec7775eb9a6be67a450080777d0e9d7cdc7111e719b00b039:1\"\]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.29s
     Running `target/debug/liana-cli --conf ../demo/ledger-tr2to1/signet/daemon.toml createrecovery tb1qzfl7v0lmsjywwwevmdm4rxknzqqvpt395vdlzcag08yfythnh5nsa4kl5u 1 '["eb9270e3bfa4831ec7775eb9a6be67a450080777d0e9d7cdc7111e719b00b039:1"]'`
{
  "error": {
    "code": -32602,
    "message": "Invalid params: Invalid 'timelock' parameter."
  }
}
```

edouardparis (approved):
ACK 2527bcea2eef753fb776e1618c7d4b9b79801a3c

Issue #1647, opened 2025-04-09
Add tooltip to "Recovery" menu option


nondiremanuel:
As a possible UI improvement, if we can include it in the cycle, we could add a tooltip to the whole "Recovery" menu option button with the same text we have now on the info tooltip ("i"):

![Image](https://github.com/user-attachments/assets/14a5e8af-7420-4b70-9a24-62e68541bb6c)

This is not essential since we will have a description in the recovery page itself, but it may still be a good way to avoid a click to the user.

PR #1648, opened 2025-04-09
feat: move recovery from settings to own menu

### LINKED ISSUE #1640: [GUI] Move Recovery page to own menu option
To improve visibility, the Recovery page should be available from the left-hand menu rather than part of the Settings page.

As the Recovery feature is not part of regular wallet usage, this menu option should be placed in the bottom left corner above "Settings".

This issue does not involve any changes to the page content.

jp1ac4:
This is to resolve #1640.

The tooltip has been removed for now as the menu buttons don't currently support adding an icon with tooltip, but a tooltip can be re-added later if required (#1647).

### DIFF
--- a/liana-gui/src/app/view/mod.rs
+++ b/liana-gui/src/app/view/mod.rs
@@ -26,7 +26,8 @@ use liana_ui::{
     color,
     component::{button, text::*},
     icon::{
-        coins_icon, cross_icon, history_icon, home_icon, receive_icon, send_icon, settings_icon,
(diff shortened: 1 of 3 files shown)


jp1ac4:
Recovery in bottom left corner above "Settings":
image
image

No longer listed in Settings:
image

edouardparis (approved):
ACK df4528dc767a39fb979579c7cafca1fd49fa3d49

Issue #1482, opened 2024-11-22
[GUI] Rename "Testnet" to "Testnet 3"


kloaec:
We should rename Testnet to Testnet 3.
Following this change, we should make sure this is done in a clean way, allowing for a future Testnet 4 choice in parallel.

PR #1650, opened 2025-04-13
Add testnet4 support

### LINKED ISSUE #1119: support `testnet4`
`testnet3` is broken, it would be good to have `testnet4` support

thomask7b:
This should closes #1119.

I've seen #1482 but as within the bitcoin crate Testnet3 is called Testnet I'd wait for your confirmation that within Liana you want to call it testnet3 instead of testnet. But it can also be done in another PR.

### DIFF
--- a/liana-gui/src/app/state/settings/bitcoind.rs
+++ b/liana-gui/src/app/state/settings/bitcoind.rs
@@ -579,6 +579,14 @@ impl RescanSetting {
                                 date
                             }
                         }
+                        Network::Testnet4 => {
+                            if date < TESTNET4_GENESIS_BLOCK_TIMESTAMP {
+                                info!("Date {} prior to genesis block, using genesis block timestamp {}", date, TESTNET4_GENESIS_BLOCK_TIMESTAMP);
+                                TESTNET4_GENESIS_BLOCK_TIMESTAMP
+                            } else {
+                                date
+                            }
+                        }
                         Network::Signet => {
                             if date < SIGNET_GENESIS_BLOCK_TIMESTAMP {
                                 info!("Date {} prior to genesis block, using genesis block timestamp {}", date, SIGNET_GENESIS_BLOCK_TIMESTAMP);
@@ -642,4 +650,5 @@ impl RescanSetting {
 /// Use bitcoin-cli getblock $(bitcoin-cli getblockhash 0) | jq .time
 const MAINNET_GENESIS_BLOCK_TIMESTAMP: i64 = 1231006505;
 const TESTNET3_GENESIS_BLOCK_TIMESTAMP: i64 = 1296688602;
+const TESTNET4_GENESIS_BLOCK_TIMESTAMP: i64 = 1659741361;
 const SIGNET_GENESIS_BLOCK_TIMESTAMP: i64 = 1598918400;
--- a/liana-gui/src/launcher.rs
+++ b/liana-gui/src/launcher.rs
@@ -16,9 +16,10 @@ use lianad::config::ConfigError;
 
 use crate::{app, installer::UserFlow};
 
-const NETWORKS: [Network; 4] = [
+const NETWORKS: [Network; 5] = [
     Network::Bitcoin,
     Network::Testnet,
+    Network::Testnet4,
     Network::Signet,
     Network::Regtest,
 ];
@@ -214,6 +215,7 @@ impl Launcher {
                                                                 Network::Bitcoin => "Bitcoin",
                                                                 Network::Signet => "Signet",
                                                                 Network::Testnet => "Testnet",
+                                                                Network::Testnet4 => "Testnet4",
                                                                 Network::Regtest => "Regtest",
                                                                 _ => "",
                                                             }
--- a/liana-gui/src/installer/step/descriptor/editor/key.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/key.rs
@@ -564,6 +564,10 @@ mod tests {
             "48'/1'/0'/2'"
         );
         assert_eq!(
+            default_derivation_path(Network::Testnet4).to_string(),
+            "48'/1'/0'/2'"
+        );
+        assert_eq!(
             default_derivation_path(Network::Signet).to_string(),
             "48'/1'/0'/2'"
         );
--- a/liana-gui/src/installer/step/node/bitcoind.rs
+++ b/liana-gui/src/installer/step/node/bitcoind.rs
@@ -102,7 +102,7 @@ impl Download {
 /// Default prune value used by internal bitcoind.
 pub const PRUNE_DEFAULT: u32 = 15_000;
 /// Default ports used by bitcoind across all networks.
-pub const BITCOIND_DEFAULT_PORTS: [u16; 8] = [8332, 8333, 18332, 18333, 18443, 18444, 38332, 38333];
+pub const BITCOIND_DEFAULT_PORTS: [u16; 10] = [8332, 8333, 18332, 18333, 18443, 18444, 48332, 48333, 38332, 38333];
 
 #[derive(Debug)]
 pub enum InstallState {
@@ -242,6 +242,7 @@ fn bitcoind_default_address(network: &Network) -> String {
     match network {
         Network::Bitcoin => "127.0.0.1:8332".to_string(),
         Network::Testnet => "127.0.0.1:18332".to_string(),
+        Network::Testnet4 => "127.0.0.1:48332".to_string(),
         Network::Regtest => "127.0.0.1:18443".to_string(),
         Network::Signet => "127.0.0.1:38332".to_string(),
         _ => "127.0.0.1:8332".to_string(),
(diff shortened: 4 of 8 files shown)


nondiremanuel:
Thank you very much for the contribution! If we are able to find some review time, we will try to include this change to the next Liana release.

pythcoiner:
comment from @jp1ac4 in our internal channeel:
> It might be worth adding a comment to USAGE.md about the min bitcoind version required for testnet4

pythcoiner:
also following an internal discussion, I think the user should be offered only Testnet4 if available, as we already have too many choices for networks.

Note: we should not drop support for Testnet3 that are already installed.

jp1ac4:
I think the idea from https://github.com/wizardsardine/liana/pull/1650#issuecomment-2849912648 was to show "testnet" (Testnet3) only for opening an existing wallet on that network, but otherwise to offer only testnet4 as an option for new wallets. So we'd need to check for existing Testnet3 wallets in order to know whether to include "testnet" in the list.

But this logic might be affected by the changes relating to #1703 so it's probably worth waiting for that to be merged before proceeding here.

liana-gui/src/launcher.rs
```diff
@@ -405,6 +410,37 @@ fn wallets_list_item(
...
+/// Returns the list of displayed networks.
+///
+/// `Testnet4` is excluded if the Bitcoin Core version is less than 28.0.
+/// `Testnet` is excluded if no wallet exists, `Testnet4` is prefered.
+fn displayed_networks(dir: &LianaDirectory) -> Vec<Network> {
+    let mut networks = NETWORKS.to_vec();
+
+    networks.retain(|&n| match n {
+        Network::Testnet4 => is_version_at_least_28_0() && dir.path().join(n.to_string()).exists(),
+        Network::Testnet => dir.path().join(n.to_string()).exists(),
```
jp1ac4: nit: you could use `dir.network_directory(n).path().exists()` here. (The existing `Launcher::new()` logic could be updated similarly.)

liana-gui/src/launcher.rs
```diff
@@ -405,6 +410,37 @@ fn wallets_list_item(
...
 
+/// Returns the list of displayed networks.
+///
+/// `Testnet4` is excluded if the Bitcoin Core version is less than 28.0.
+/// `Testnet` is excluded if no wallet exists, `Testnet4` is prefered.
+fn displayed_networks(dir: &LianaDirectory) -> Vec<Network> {
+    let mut networks = NETWORKS.to_vec();
+
+    networks.retain(|&n| match n {
+        Network::Testnet4 => is_version_at_least_28_0() && dir.path().join(n.to_string()).exists(),
```
jp1ac4: I think we can show testnet4 in all cases. Possibly there will be an error when connecting to the node, but we won't know the user's bitcoind version until trying to connect.

Even for managed bitcoind, the latest version (currently 29.0) will be downloaded in the installer when creating a testnet4 wallet, and any other managed bitcoind versions already downloaded will work with the user's existing wallets on other networks.

liana-gui/src/launcher.rs
```diff
@@ -405,6 +410,37 @@ fn wallets_list_item(
...
+        Network::Testnet => dir.path().join(n.to_string()).exists(),
+        _ => true,
+    });
+
+    if !networks
```
jp1ac4: I'm not sure if this logic is needed. Isn't it enough to check if the testnet3 network directory exists (or perhaps also that the settings file exists therein) and to show testnet3 only in this case?

liana-gui/src/launcher.rs
thomasballivet: Don't you think that we should not show testnet4 if `is_version_at_least_28_0()` is not true?

jp1ac4: In principle yes, but the `VERSION` only tells us the Liana-managed bitcoind version that will be downloaded for new wallets. So perhaps this check could be moved to the node selection step in the installer to decide whether to offer a Liana-managed bitcoind to the user, e.g. if the user chooses testnet4 and the managed bitcoind version is earlier than 28.0, then we don't show the managed bitcoind option. But I think we can assume that the `VERSION` for new wallets won't be lowered from the current value of 29.0, and so we can assume that the managed bitcoind version is recent enough for testnet4 when the user is installing a new wallet.

thomasballivet: Ok I get the point now.
I'm hesitating too but I do think that if we keep it simple the second option is better and even if I agree on the first part I don't think this specific case deserve this adaptation to this installer.
But let me know what's the best option for you

jp1ac4: I would remove the version check from the launcher for now and I don't think any version checks are required in the installer at this stage.

thomasballivet: done

doc/USAGE.md
```diff
@@ -81,8 +81,9 @@ which is currently possible through the Bitcoin Core daemon (`bitcoind`) or an E
 
 The chosen Bitcoin backend must be available while Liana is running.
 
-If using `bitcoind`, it must be running on your machine for the desired network (mainnet, signet, testnet or regtest)
-and may be pruned (note this may affect block chain rescans) up to the maximum (around 550MB of blocks).
+If using `bitcoind`, it must be running on your machine for the desired network (mainnet, signet, testnet, testnet4 or
+regtest). Note that testnet4 will only be available if Bitcoin Core version 28.0 or later is used. It may be pruned 
+(this may affect block chain rescans) up to the maximum (around 550MB of blocks).
```
jp1ac4, line 86: This needs to be modified to reflect the updated logic.

liana-gui/src/launcher.rs
```diff
@@ -405,6 +409,21 @@ fn wallets_list_item(
...
 
+/// Returns the list of displayed networks.
+///
+/// `Testnet4` is displayed if the Bitcoin Core version is at least 28.0.
+/// `Testnet` is displayed if no wallet exists, or if Bitcoin Core version is lower than 28.0.
+fn displayed_networks(dir: &LianaDirectory) -> Vec<Network> {
+    let mut networks = NETWORKS.to_vec();
+
+    networks.retain(|&n| match n {
+        Network::Testnet => dir.network_directory(n).path().exists(),
```
jp1ac4: Deleting a testnet3 wallet may leave the network directory in place, so I think we need to check for the settings file to detect if there are still any wallets, similarly to `Launcher::new()`:
```diff
diff --git a/liana-gui/src/launcher.rs b/liana-gui/src/launcher.rs
index c38fdb9a..f1097ccf 100644
--- a/liana-gui/src/launcher.rs
+++ b/liana-gui/src/launcher.rs
@@ -417,7 +417,11 @@ fn displayed_networks(dir: &LianaDirectory) -> Vec<Network> {
     let mut networks = NETWORKS.to_vec();
 
     networks.retain(|&n| match n {
-        Network::Testnet => dir.network_directory(n).path().exists(),
+        Network::Testnet => dir
+            .network_directory(n)
+            .path()
+            .join(settings::SETTINGS_FILE_NAME)
+            .exists(),
         _ => true,
     });

```

Note you might be able to extract this logic to a common function and refactor `Launcher::new()` to also use it.

doc/USAGE.md
thomasballivet: Even if it is displayed on the UI testnet4 won't work if the bitcoin core version is < 28.0, are you sure we should update this?

thomasballivet: but at launcher.rs:413 I did the change

jp1ac4: Ah sorry, I misread it. This is fine, no change needed :)

liana-gui/src/app/state/settings/bitcoind.rs
```diff
@@ -660,4 +668,5 @@ impl RescanSetting {
 /// Use bitcoin-cli getblock $(bitcoin-cli getblockhash 0) | jq .time
 const MAINNET_GENESIS_BLOCK_TIMESTAMP: i64 = 1231006505;
 const TESTNET3_GENESIS_BLOCK_TIMESTAMP: i64 = 1296688602;
+const TESTNET4_GENESIS_BLOCK_TIMESTAMP: i64 = 1659741361;
```
jp1ac4: I get 1714777860 for this value, which also seems to agree with https://mempool.space/testnet4/block/00000000da84f2bafbbc53dee25a72ae507ff4914b867c565be350b0da8bf043

liana-gui/src/installer/step/wallet_alias.rs
```diff
@@ -33,6 +33,7 @@ impl Step for WalletAlias {
                         Network::Bitcoin => "Bitcoin",
                         Network::Signet => "Signet",
                         Network::Testnet => "Testnet",
+                        Network::Testnet4 => "Testnet",
```
jp1ac4: I think we can use "Testnet4" here to match what we display in the dropdown and also what we do in `wallets_list_item()`.

Issue #1641, opened 2025-04-04
[GUI] Enable coin control on Recovery page


jp1ac4:
The Recovery page should allow the user to select which coins to use for a recovery transaction.

For this, the Recovery page should be similar to the Send page with some differences:
-  As is currently the case, a recovery transaction is a sweep to a single address, so there will be no "+ Add payment" button.
- The amount will be the maximum from selected coins and not editable by the user. It will be similar to ticking the MAX box for a regular spend, but there will be no MAX box visible on the Recovery page.
- There will be a drop down to select the recovery path to use (probably above the list of coins). When selected, all coins that can be spent on this recovery path (using the current logic for selecting recovery coins) will be displayed. All coins will be selected by default and the user can then deselect as required.
- If there are no recovery paths available, this will be stated and there will not be any other content on the page. TBD if the user should first select a recovery path from the drop down before seeing this message. In either case, there should not be any editable fields on the page if no coins are spendable.

PR #1658, opened 2025-04-16
feat: Add coin control to Recovery panel

jp1ac4:
This implements #1641.

The recovery process is treated as a regular spend, with an initial step for selecting the recovery path.

Further refactoring to `DefineSpend` could be applied in a follow-up to clearly distinguish between the three kinds of spends now handled:
1) regular primary path spend to external recipient(s)
2) self-send (using primary path)
3) recovery spend


jp1ac4:
Rebased on master.

Currently making some changes in relation to reloading steps in the recovery process. Will push updates in a short while.

jp1ac4:
I've updated how the recovery panel reloads in order to eliminate discrepancies between the coins loaded on different steps of the recovery panel as well as the tip height used for determining which coins are recoverable:
1) When the recovery panel is reloaded, we force a cache update and then use these cached values to update the state. This ensures consistency between the coins and the tip height. Previously, we only reloaded the coins and so might have needed to wait for the cached tip height to update. Note this does not affect how regular spends are reloaded.
2) For a recovery spend, when we navigate to the next or previous step, this new step loads the latest cached data to ensure consistency with the previously selected step. Otherwise, it was possible that the recovery path selection step would not correspond to the coins available for selection.

Other changes made were:
1) For a recovery, we keep track of whether the user has edited the coins selection in order to know whether any newly available coins should be selected on a reload event.
2) If the user changes the chosen recovery path, the next step is reset and any previous state is cleared.

jp1ac4:
I've updated how the spend panel is reloaded so that it no longer uses the cache. The cache will only be used for the initial creation when starting the GUI.

Instead, the spend panel's reload method will get both the latest coins and the tip height and pass these to the current step, as well as storing them for use when moving between steps.

liana-gui/src/app/state/spend/mod.rs
```diff
@@ -20,24 +21,76 @@ use crate::{
 
 pub struct CreateSpendPanel {
     draft: step::TransactionDraft,
+    /// The timelock of the recovery path to use for spending.
+    ///
+    /// If the primary path will be used, this value will remain as `None`.
+    /// Otherwise, its value should always be set to a recovery path,
+    /// which may change from one to another.
+    recovery_timelock: Option<u16>,
```
edouardparis, line 29: What if we move the targeted recovery_timelock to the step::TransactionDraft ?
it seems that both are linked and moved in the same methods.

jp1ac4:
Rebased on master.

Currently refactoring as per https://github.com/wizardsardine/liana/pull/1658#discussion_r2055397918.

liana-gui/src/app/state/spend/mod.rs
jp1ac4: This has now been done.

jp1ac4:
Moved `recovery_timelock` to `TransactionDraft` as per https://github.com/wizardsardine/liana/pull/1658#discussion_r2055397918.

edouardparis (approved):
ACK c7bbbd854ccfb8f2dccb6e7f63b6f4c982cd76c1

PR #1661, opened 2025-04-17
fix: cache coins when first loading GUI with Liana Connect

jp1ac4:
With this PR, coins will be cached when first loading a GUI that is using Liana Connect as the backend.

The cache refreshes less often for Liana Connect (every 2 mins), which means there can be a discrepancy between the coins loaded on a page and those in the cache, especially when first loading the GUI with no coins cached.

This can be seen when choosing to refresh a coin in the Coins panel. After selecting the coin and the refresh option, the Self-Send page is displayed with coins listed, but the coin that was previously selected in the Coins panel won't be selected here unless the cache has already refreshed. By caching coins at the start, it makes such a scenario less likely to occur when using Liana Connect.

### DIFF
--- a/liana-gui/src/app/cache.rs
+++ b/liana-gui/src/app/cache.rs
@@ -1,6 +1,11 @@
-use crate::daemon::model::Coin;
+use crate::daemon::{
+    model::{Coin, ListCoinsResult},
+    Daemon, DaemonError,
+};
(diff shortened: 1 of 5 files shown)


edouardparis (approved):
ACK 417a949037b7c33895e34081d9acd23ddd4a10f8

PR #1665, opened 2025-04-22
Remove the link between gui config and lianad config

edouardparis:
This commit is part of preparatory work to support multiple wallets.

This commit introduces two breaking change:

The fields daemon_config_path and daemon_rpc_path are removed. The GUI will deduce these values from their expected location in the root data directory.

Because this link is removed, the gui flag --conf is not useful anymore as it cannot then find the location of the root data directory.


jp1ac4 (approved):
ACK 19f5132ee8439ef3d4d27dd7254104d60c8540af.

Tested with new & existing wallets using local & remote backends.

PR #1666, opened 2025-04-23
Prepare datadir for multiple wallet support

edouardparis:
A new field data_directory is introduced in lianad configuration, it resolves directly to the path of the wallet datadir without the per network reasoning. The previous data_dir field is kept for backward compatibility for already generated configuration file.


jp1ac4 (approved):
ACK 61bc0ffbc85748437da0155ea25a5b741306aef2.

Tested with new & existing wallets using local & remote backends.

Issue #1664, opened 2025-04-22
[GUI] Recovery panel small fixes & improvements


jp1ac4:
These are some small UI fixes & improvements for the Recovery panel:
- change "N recovery paths will be available at the next block" to "N recovery paths are available" / "1 recovery path is available"
- show fingerprint alias to the right-hand side rather than below to improve visibility
- reload fingerprint aliases in case they've been edited in settings
- add explanatory text "Send your funds to another wallet's address if you have lost access to your primary spending path." or similar on the page where a recovery path is selected
- show "can recover 1 coin" instead of "can recover 1 coins"

PR #1667, opened 2025-04-23
fix: ensure spend (& recovery) panel steps use latest fingerprint aliases

jp1ac4:
This fixes one of the items from #1664, but it's a little different from the other items there so I felt a separate PR would be better.

This fixes an issue for the Recovery panel that was introduced by #1658, but which was already present in the Send panel. If, while creating a regular spend, the user edits the fingerprint aliases, these updated aliases would not have shown in the signing info of the draft PSBT. For a recovery, this issue arises on the first step when selecting a recovery path, where the aliases are shown for each key required on that path.

The issue arises because the `wallet` field of individual steps was not being updated.

Now, the `wallet` will be stored by `CreateSpendPanel` and used by `load()` when moving between steps. Furthermore, the `WalletUpdated` message will be used to update the current step directly.

### DIFF
--- a/liana-gui/src/app/state/spend/mod.rs
+++ b/liana-gui/src/app/state/spend/mod.rs
@@ -185,8 +185,11 @@ impl State for CreateSpendPanel {
     fn reload(
         &mut self,
         daemon: Arc<dyn Daemon + Sync + Send>,
-        _wallet: Arc<Wallet>,
+        wallet: Arc<Wallet>,
     ) -> Task<Message> {
+        for step in self.steps.iter_mut() {
+            step.reload_wallet(wallet.clone());
+        }
         let daemon1 = daemon.clone();
         let daemon2 = daemon.clone();
         let coin_statuses_1 = if self.draft.is_recovery() {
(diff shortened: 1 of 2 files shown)


jp1ac4:
Note there's an alternative approach from @thomask7b in https://github.com/wizardsardine/liana/pull/1669/commits/745bc01cf3d3f3b26ee4c7a375de7a8ac40f7784 where the recovery panel (and possibly others) would be updated upon a `WalletUpdated` message. I considered something like that, but felt we might then update panels that are not being used. On the other hand, reloading the wallet each time we open a panel might be too much if we don't expect the wallet to be updated often.

liana-gui/src/app/state/spend/mod.rs
```diff
@@ -231,6 +246,9 @@ impl State for CreateSpendPanel {
                 },
                 Message::Labels,
             ),
+            Task::perform(async move { wallet }, |wallet| {
```
edouardparis, line 249: I think you do not need to trigger a new Task, you can add a new method `reload_wallet` to the steps, and call it for all of them in the reload method of the CreateSpendPanel.

jp1ac4: This is now done and was much simpler :relaxed:

edouardparis (approved):
utACK 7d2a1d4d315819ad18c040c7ee3e2830b40d8d6f

PR #1668, opened 2025-04-23
Small fixes and improvements to the Recovery panel UI

jp1ac4:
This fixes and implements the items from #1664, except for the alias refresh, which is covered by #1667.

### DIFF
--- a/liana-gui/src/app/view/recovery.rs
+++ b/liana-gui/src/app/view/recovery.rs
@@ -118,7 +118,7 @@ pub fn recovery_path_view<'a>(
                                                 .padding(5)
                                                 .style(theme::pill::simple),
                                             liana_ui::widget::Text::new(fg.to_string()),
-                                            tooltip::Position::Bottom,
+                                            tooltip::Position::Top,
                                         )
                                         .style(theme::card::simple),
                                     )
--- a/liana-gui/src/app/view/recovery.rs
+++ b/liana-gui/src/app/view/recovery.rs
@@ -34,33 +34,40 @@ pub fn recovery<'a>(
     warning: Option<&Error>,
 ) -> Element<'a, Message> {
     let no_recovery_paths = recovery_paths.is_empty();
+    const INFO_TEXT: &str = "Recover your funds by sending them to another wallet if you have lost access to your primary spending path.";
     dashboard(
         &Menu::Recovery,
         cache,
         warning,
         Column::new()
             .push(Container::new(h3("Recovery")).width(Length::Fill))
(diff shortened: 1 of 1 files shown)


jp1ac4:
Here are a couple of screenshots:

image
image

liana-gui/src/app/view/recovery.rs
```diff
@@ -34,28 +34,40 @@ pub fn recovery<'a>(
     warning: Option<&Error>,
 ) -> Element<'a, Message> {
     let no_recovery_paths = recovery_paths.is_empty();
+    const INFO_TEXT: &str = "Recover your funds by sending them to another wallet if you have lost access to your primary spending path.";
```
thomask7b, line 37: Since it will probably not be reused I'd suggest to put directly the string at line 44, it would makes this view more readable.
In contrast, I'd introduce `nb_recovery_paths` to avoid multiple calls on `recovery_paths.len()`.

jp1ac4: That's true, it was mainly because putting the long string directly below would stop `cargo fmt` working correctly.

An alternative that still avoids the `cargo fmt` issue would be
```diff
diff --git a/liana-gui/src/app/view/recovery.rs b/liana-gui/src/app/view/recovery.rs
index e866f2dc..28007a19 100644
--- a/liana-gui/src/app/view/recovery.rs
+++ b/liana-gui/src/app/view/recovery.rs
@@ -34,14 +34,17 @@ pub fn recovery<'a>(
     warning: Option<&Error>,
 ) -> Element<'a, Message> {
     let no_recovery_paths = recovery_paths.is_empty();
-    const INFO_TEXT: &str = "Recover your funds by sending them to another wallet if you have lost access to your primary spending path.";
     dashboard(
         &Menu::Recovery,
         cache,
         warning,
         Column::new()
             .push(Container::new(h3("Recovery")).width(Length::Fill))
-            .push(Container::new(text(INFO_TEXT)))
+            .push(Container::new(text(
+                "Recover your funds by sending them \
+                to another wallet if you have lost access \
+                to your primary spending path.",
+            )))
             .push(Space::with_height(Length::Fixed(20.0)))
             .push(
                 Container::new(
```

I'm not sure if there's a strong advantage to one over the other.

Regarding `recovery_paths.len()`, as there are only a couple of calls, I think it's fine to keep it as it is. Also, I don't think there's much overhead to calling `len()` as the length is stored by the vector object. But I'm not sure what the best practice is in such cases.

edouardparis (approved):
utACK 356301c79c6f621e8329287f70a112ddbb16d43d

Issue #1656, opened 2025-04-16
Bump managed bitcoind version to 29.0


nondiremanuel:
Tracking issue for the activity of bumping the bitcoin core version in the liana-managed bitcoind backend option.

PR #1669, opened 2025-04-24
Small UI fixes

### LINKED ISSUE #1664: [GUI] Recovery panel small fixes & improvements

thomask7b:
I did this to fix #1664 and then I saw #1667 and #1668.
I pushed anyway so you can see I had a different approach for the wallet update (fingerprint aliases), I did it in a separate commit so you can check. It was still in progress, I'd have prefer to batch the panel updates with `Task::batch(commands)` like it has been done above view/spend/mod.rs from 349 to 359.
And also so you know that if you want assign me on some issues you're most welcome.

Capture d’écran du 2025-04-23 20-47-53

Capture d’écran du 2025-04-24 03-28-24

### DIFF
--- a/liana-gui/src/app/mod.rs
+++ b/liana-gui/src/app/mod.rs
@@ -368,6 +368,11 @@ impl App {
             }
             Message::WalletUpdated(Ok(wallet)) => {
                 self.wallet = wallet.clone();
+                let _ = self.panels.recovery.update(
+                    self.daemon.clone(),
+                    &self.cache,
+                    Message::WalletUpdated(Ok(wallet.clone())),
+                );
                 self.panels.current_mut().update(
                     self.daemon.clone(),
                     &self.cache,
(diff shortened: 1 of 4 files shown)


nondiremanuel:
Thanks @thomask7b for contributing and proposing to tackle some issues for v11. Here you can find a short list of small UI/UX improvements that we would be very happy to include:

- https://github.com/wizardsardine/liana/issues/1656
- https://github.com/wizardsardine/liana/issues/1670
- https://github.com/wizardsardine/liana/issues/1041

Feel free to tackle one of those or ask if you see others that you might be interested in. It could be helpful if you comment on the issue you are going to work on first so that we don't duplicate the work. Thanks again!

PR #1672, opened 2025-04-24
Fix sliders background color

### LINKED ISSUE #1670: Fix color contrast for timelock duration setting
The color contrast between the disabled part of the selection bar and the background is not enough. The bar results invisible. This is probably subsequent to the iced upgrade.

Image

We should increase the contrast in order for it to be more visible.

thomask7b:
You can see here the black background for the slider : 

Capture d’écran du 2025-04-24 14-39-48

Closes #1670

### DIFF
--- a/liana-ui/src/theme/palette.rs
+++ b/liana-ui/src/theme/palette.rs
@@ -497,7 +497,7 @@ impl std::default::Default for Palette {
                 background: color::GREEN,
                 border: color::GREEN,
                 rail_border: None,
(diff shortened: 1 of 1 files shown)


nondiremanuel:
> Something like that?
> 
> Capture d’écran du 2025-04-24 16-35-43

Yes, it's definitely better imo. Thanks!

pythcoiner:
utACK 7664567

PR #1674, opened 2025-04-24
Use type wrapper instead of PathBuf for directories

edouardparis:
With the multiplicity of directories, liana directory, network directory and in the future per wallet lianad directories, passing a simple type PathBuf does not transmit to the contributors the information about which directory each module works with.

This commit introduces wrappers around PathBuf
to give this information and link parent directories to their child directories.


jp1ac4 (approved):
ACK bb7777878ea639def09a50f52fdd338467e8a87d.

Tested with new and existing wallets, including using managed bitcoind.

PR #1675, opened 2025-04-25
Disable signing devices that should not or cannot spend

### LINKED ISSUE #1041: [GUI] Grey out hardware device that are not part of primary path

thomask7b:
Inspired by #1607 this shoud closes #1041.

Notice that I added the wanted logic also for the recovery spending. That means you can only select a signing device related to the primary path for usual spend. But for recovery spending you can only select a signing device related to the recovery_path.

You can see here both examples :

Capture d’écran du 2025-04-25 17-32-39

Capture d’écran du 2025-04-25 17-30-53

### DIFF
--- a/liana-gui/src/app/state/psbt.rs
+++ b/liana-gui/src/app/state/psbt.rs
@@ -187,6 +187,7 @@ impl PsbtState {
                     cache.datadir_path.clone(),
                     cache.network,
                     self.saved,
+                    self.tx.recovery_timelock(),
                 );
                 let cmd = modal.load(daemon);
                 self.modal = Some(PsbtModal::Sign(modal));
@@ -438,6 +439,7 @@ pub struct SignModal {
     signed: HashSet<Fingerprint>,
     is_saved: bool,
     display_modal: bool,
+    recovery_timelock: Option<u16>,
 }
 
 impl SignModal {
@@ -447,6 +449,7 @@ impl SignModal {
         datadir_path: LianaDirectory,
         network: Network,
         is_saved: bool,
+        recovery_timelock: Option<u16>,
     ) -> Self {
         Self {
             signing: HashSet::new(),
@@ -456,6 +459,7 @@ impl SignModal {
             signed,
             is_saved,
             display_modal: true,
+            recovery_timelock,
         }
     }
 }
@@ -565,6 +569,7 @@ impl Modal for SignModal {
                 view::psbt::sign_action(
                     self.error.as_ref(),
                     &self.hws.list,
+                    &self.wallet.main_descriptor,
                     self.wallet.signer.as_ref().map(|s| s.fingerprint()),
                     self.wallet
                         .signer
@@ -572,6 +577,7 @@ impl Modal for SignModal {
                         .and_then(|signer| self.wallet.keys_aliases.get(&signer.fingerprint)),
                     &self.signed,
                     &self.signing,
+                    self.recovery_timelock,
                 ),
             )
             .on_blur(Some(view::Message::Spend(view::SpendTxMessage::Cancel)))
--- a/liana-gui/src/app/view/hw.rs
+++ b/liana-gui/src/app/view/hw.rs
@@ -13,6 +13,7 @@ pub fn hw_list_view(
     hw: &HardwareWallet,
     signed: bool,
     signing: bool,
+    can_sign: bool,
 ) -> Element<Message> {
     let mut bttn = Button::new(match hw {
         HardwareWallet::Supported {
@@ -40,6 +41,8 @@ pub fn hw_list_view(
                     alias.as_ref(),
                     "The wallet descriptor is not registered on the device.\n You can register it in the settings.",
                 )
+            } else if !can_sign {
+                hw::disabled_hardware_wallet(kind, version.as_ref(), fingerprint, "This signing device is not part of this spending path.")
             } else {
                 hw::supported_hardware_wallet(kind, version.as_ref(), fingerprint, alias.as_ref())
             }
@@ -71,7 +74,7 @@ pub fn hw_list_view(
     })
     .style(theme::button::secondary)
     .width(Length::Fill);
-    if !signing {
+    if can_sign && !signing {
         if let HardwareWallet::Supported { registered, .. } = hw {
             if *registered != Some(false) {
                 bttn = bttn.on_press(Message::SelectHardwareWallet(i));
--- a/liana/src/descriptors/analysis.rs
+++ b/liana/src/descriptors/analysis.rs
@@ -9,6 +9,7 @@ use miniscript::{
     RelLockTime, ScriptContext, Threshold,
 };
 
+use miniscript::bitcoin::bip32::Fingerprint;
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     convert::TryFrom,
@@ -367,6 +368,11 @@ impl PathInfo {
                 .map_err(|e| LianaPolicyError::InvalidPolicy(miniscript::Error::Threshold(e)))?,
             ),
         })
+    }
+
+    /// Determine whether the fingerprint is part of this path.
+    pub fn contains_fingerprint(&self, fingerprint: Fingerprint) -> bool {
+        self.thresh_origins().1.contains_key(&fingerprint)
     }
 }
 
--- a/liana/src/descriptors/mod.rs
+++ b/liana/src/descriptors/mod.rs
@@ -217,6 +217,39 @@ impl LianaDescriptor {
             .for_any_key(|k| k.master_fingerprint() == fg)
     }
 
+    /// Determine whether the fingerprint is part of a specific path of this descriptor.
+    /// If recovery_timelock is None, checks in the primary path.
+    /// If recovery_timelock is Some(timelock), checks in the recovery path with specified timelock.
+    pub fn contains_fingerprint_in_path(
+        &self,
+        fingerprint: Fingerprint,
+        recovery_timelock: Option<u16>,
+    ) -> bool {
+        match recovery_timelock {
+            None => self.contains_fingerprint_in_primary_path(fingerprint),
+            Some(timelock) => self.contains_fingerprint_in_recovery_path(fingerprint, timelock),
+        }
+    }
+
+    /// Determine whether the fingerprint is part of the primary path of this descriptor.
+    fn contains_fingerprint_in_primary_path(&self, fingerprint: Fingerprint) -> bool {
+        self.policy().primary_path.contains_fingerprint(fingerprint)
+    }
+
+    /// Determine whether the fingerprint is part of the recovery path of this descriptor for the
+    /// specified timelock.
+    fn contains_fingerprint_in_recovery_path(
+        &self,
+        fingerprint: Fingerprint,
+        recovery_timelock: u16,
+    ) -> bool {
+        self.policy()
+            .recovery_paths
+            .get(&recovery_timelock)
+            .map(|path_info| path_info.contains_fingerprint(fingerprint))
+            .unwrap_or(false)
+    }
+
     /// Get the descriptor for receiving addresses.
     pub fn receive_descriptor(&self) -> &SinglePathLianaDesc {
         &self.receive_desc
@@ -751,9 +784,9 @@ impl DerivedSinglePathLianaDesc {
 mod tests {
     use super::*;
 
-    use bitcoin::{hashes::Hash, Sequence};
-
     use crate::signer::HotSigner;
+    use bitcoin::{hashes::Hash, Sequence};
+    use miniscript::bitcoin::bip32::Fingerprint;
 
     fn random_desc_key(
         secp: &secp256k1::Secp256k1<impl secp256k1::Signing>,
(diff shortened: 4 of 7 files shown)


liana/src/descriptors/mod.rs
```diff
@@ -588,6 +588,25 @@ impl LianaDescriptor {
             .checked_div(witness_factor)
             .unwrap()
     }
+
+    /// Determine whether the fingerprint is part of the primary path of this descriptor.
+    pub fn is_part_of_primary_path(&self, fingerprint: &Fingerprint) -> bool {
```
jp1ac4, line 593: I would rename this as the `is_part_of_` suggests the descriptor is part of the fingerprint rather than the other way round. Perhaps `contains_fingerprint_in_primary_path`, and similarly  `_recovery_paths` for the method below.

liana/src/descriptors/mod.rs
```diff
@@ -588,6 +588,25 @@ impl LianaDescriptor {
...
+        self.policy()
+            .primary_path
+            .thresh_origins()
+            .1
+            .contains_key(fingerprint)
```
jp1ac4, line 598: You could add this logic as a method on the `PathInfo` struct.

jp1ac4 (commented):
I've added some initial comments regarding the new methods, but overall they're looking good. Thanks also for adding those tests.

There will likely be some changes needed in how we display those devices that can't sign but we'll need to come back to you on that.

thomask7b:
Sure, I'll take into account your comments, I fully agree with them.
Let me know when you know more about the display.

nondiremanuel:
Thanks for the PR! 

I personally don't like having the option in red, which could scare the user. I would prefer just having it disabled, with possibly a message, coherently with #1607. Also, is the green appearing only on hover (as I would expect) or every time?

thomask7b:
So the behavior you want is a normal component but nothing should happen on hover or on click, right? 
Ok I can add a message instead.
Yes on this example the mouse is hover the green one.

thomask7b:
You can see here the display updated :
Capture d’écran du 2025-04-29 23-20-27
I have been using the same display as in #1607 updating the view for the hot wallet. I updated it here because it has not been impacted by #1607.

That said here you can see below the old message that we coud adapt to precise the spending path (like by passing a is_recovery_tx to `unrelated_hardware_wallet`). I didn't adapt in this PR to avoid conflicts in case it's merged before #1607.
Otherwise I would have to take here the changes of #1607.
Let me know what flow would you prefer.
Capture d’écran du 2025-04-29 23-19-20

nondiremanuel:
> You can see here the display updated : Capture d’écran du 2025-04-29 23-20-27 I used the same display as in #1607 updating the view for the hot wallet. I updated it here because it has not been impacted by #1607.
> 
> That said here you can see below the old message that we coud adapt to precise the spending path (like by passing a `is_recovery_tx` to `unrelated_hardware_wallet` for example). I didn't adapt in this PR to avoid conflicts in case it's merged before #1607. Otherwise I would have to take here the changes of #1607. Let me know what flow would you prefer. Capture d’écran du 2025-04-29 23-19-20

Looks good to me, thanks! In terms of what goes in first, I let the others comment according to the review and test process.

liana/src/descriptors/mod.rs
```diff
@@ -588,6 +588,19 @@ impl LianaDescriptor {
...
             .unwrap()
     }
+
+    /// Determine whether the fingerprint is part of the primary path of this descriptor.
+    pub fn contains_fingerprint_in_primary_path(&self, fingerprint: &Fingerprint) -> bool {
+        self.policy().primary_path.contains(fingerprint)
+    }
+
+    /// Determine whether the fingerprint is part of the recovery path of this descriptor.
+    pub fn contains_fingerprint_in_recovery_path(&self, fingerprint: &Fingerprint) -> bool {
```
pythcoiner, line 598: I think it's not enought to check if the fingerprint is in `any` recovery path, we should pass the targeted recovery path timelock as argument of this functon, else it can still be confusing for an user that try to sign w/ a key from a different recovery path than the expected one

thomask7b: Yes I understand I will adapt

pythcoiner:
I'll review after #1607 merged

thomask7b:
Since #1607 has been merged, the text used in `unrelated_hardware_wallet` : "This signing device is not related to this Liana wallet." was not relevant for the cases of this PR. So I adapted this a bit. Just saying to make the re-review easier.

Capture d’écran du 2025-05-06 22-13-26

Capture d’écran du 2025-05-06 22-18-22

liana-gui/src/app/state/psbt.rs
```diff
@@ -92,10 +92,16 @@ pub struct PsbtState {
     pub warning: Option<Error>,
     pub labels_edited: LabelsEdited,
     pub modal: Option<PsbtModal>,
+    pub recovery_timelock: Option<u16>,
```
jp1ac4, line 95: I think you could avoid an extra field here by instead adding a method `SpendTx::recovery_timelock()`.

This method would need to check each input of the `SpendTx`'s `psbt.unsigned_tx` and see if its sequence is for a relative timelock that is height-based. You could take the max such timelock across all inputs (although we expect they would be the same).

I think you could use this: https://docs.rs/bitcoin/0.32.5/bitcoin/blockdata/transaction/struct.Sequence.html#method.to_relative_lock_time.

liana/src/descriptors/mod.rs
```diff
@@ -2210,13 +2229,77 @@ mod tests {
...
+            descr.contains_fingerprint_in_primary_path(&Fingerprint::from_str("aabb0011").unwrap())
+        );
+        assert!(!descr
+            .contains_fingerprint_in_primary_path(&Fingerprint::from_str("bbcc2233").unwrap()));
+        assert!(!descr
+            .contains_fingerprint_in_primary_path(&Fingerprint::from_str("ddeeff00").unwrap()));
+    }
+
+    #[test]
+    fn descriptor_contains_fingerprint_in_recovery_path_multi() {
```
jp1ac4, line 2262: It may be good to add a test for a descriptor with more than one recovery path to make sure the correct path is selected, using different fingerprints in each path (as you've done in the subsequent test).

liana-gui/src/app/view/hw.rs
```diff
@@ -13,6 +13,7 @@ pub fn hw_list_view(
     hw: &HardwareWallet,
     signed: bool,
     signing: bool,
+    is_enabled: bool,
```
jp1ac4, line 16: I'm thinking that `can_sign` might be a better name for this parameter, the same as your function. Or was there a reason you didn't use that name here?

liana/src/descriptors/analysis.rs
```diff
@@ -368,6 +369,11 @@ impl PathInfo {
             ),
         })
     }
+
+    /// Determine whether the fingerprint is part of this path.
+    pub fn contains(&self, fingerprint: &Fingerprint) -> bool {
```
jp1ac4, line 374: I think `contains_fingerprint` would be a clearer name here and more consistent the other methods, especially if we want to add any other `contains_` methods to `PathInfo` in the future.

jp1ac4 (commented):
Thanks for the work here, it's looking good.

I've added a few comments for your consideration.

For future reference, you might have been able to split some of the changes into smaller, self-contained commits, e.g. some intermediate steps for refactoring, but I'm not always good at doing that myself so it's fine :smile:

liana/src/descriptors/analysis.rs
thomask7b: Yes sure (i'm coming from Java /Kotlin and we can have same named methods if the signature is different, that's why I did this mistake ^^)

liana-gui/src/app/view/hw.rs
thomask7b: Just that in my mind it was clearer to say "if it can sign then it's enabled" but yes `can_sign` is also very clear

liana/src/descriptors/mod.rs
thomask7b: Yes it's right, I'm going to cover those cases

liana-gui/src/app/state/psbt.rs
thomask7b: I can look into it if you really think this way is better but for now I don't understand why recalculate this data when we can just take it from the actually selected spending path?

jp1ac4: It was to reduce the number of fields and function parameters, but perhaps you're right that it's easier to store/pass this information directly. So you can leave it as it is for now.

liana/src/descriptors/mod.rs
```diff
@@ -595,6 +595,25 @@ impl LianaDescriptor {
             .checked_div(witness_factor)
             .unwrap()
     }
+
+    /// Determine whether the fingerprint is part of the primary path of this descriptor.
+    pub fn contains_fingerprint_in_primary_path(&self, fingerprint: &Fingerprint) -> bool {
```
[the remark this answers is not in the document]
jp1ac4: Another thought is that it might be more convenient to use in the GUI if we have instead a single method:
```
contains_finerprint_in_path(&self, fingerprint: Fingerprint, recovery_timelock: Option<u16>)
```
where if `recovery_timelock` is `None`, it checks the primary path. What do you think?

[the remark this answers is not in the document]
thomask7b: I don't like to adapt the library, or code that could be shared for many use cases, to fit one specific GUI needs. And I think it's clearer with two specific methods each doing exactly what they have to do.
But at the same time what you say is very true because the risk is to repeat this logic in many view usages and we have nothing between the view and the library to handle this logic that I implemented in `can_sign` and that you're suggesting to put here.
So yes, I think it's better, but I'd keep the two previous methods as private to keep it very clear when reading the code. Let me know what you think

liana-gui/src/app/state/psbts.rs
```diff
@@ -119,7 +120,7 @@ impl State for PsbtsPanel {
             }
             Message::View(view::Message::Select(i)) => {
                 if let Some(tx) = self.spend_txs.get(i) {
-                    let tx = psbt::PsbtState::new(self.wallet.clone(), tx.clone(), true);
+                    let tx = psbt::PsbtState::new(self.wallet.clone(), tx.clone(), true, None);
```
jp1ac4, line 123: There might be recovery PSBTs listed on the PSBTs panel. You could try using `tx.sigs` to get recovery info similarly to how we determine whether to show the "Recovery" badge:

https://github.com/wizardsardine/liana/blob/619ea6923f7af64e5a56b73dca095811378a9c50/liana-gui/src/app/view/psbts.rs#L105

jp1ac4: Looking into it further, `tx.sigs` will contain those recovery paths that are within the PSBT input sequence value, i.e. path timelock <= PSBT input sequence, since any such path can be used with this sequence value.

So we could either assume the PSBT input sequence is exactly the spending path that should be used or otherwise allow the user to sign with any path that is within the sequence. Probably the latter is more consistent with current bahaviour. Wdyt @edouardparis @pythcoiner ?

In this case, I think the logic in general for this PR would need to be modified slightly so that we don't check the signing device against a single recovery path but rather against all those contained in `sigs.recovery_paths()`.

jp1ac4: After further discussion, we feel that the UX will be better if there's only one recovery path available to sign with. So in this case, I think we can use here the max timelock in `sigs.recovery_paths()`. The rest of the PR can stay the same.

jp1ac4: Thanks for testing those scenarios. The recovery paths in `tx.sigs` are populated as follows:
https://github.com/wizardsardine/liana/blob/b4062b28a9036ef0faf56d15a06e8768568d85ef/liana/src/descriptors/mod.rs#L392-L402

It includes all recovery paths with a timelock <= the timelock used for the PSBT, so taking the max will give us the one used for the PSBT.

liana-gui/src/app/state/psbts.rs
```diff
@@ -119,7 +120,18 @@ impl State for PsbtsPanel {
             }
             Message::View(view::Message::Select(i)) => {
                 if let Some(tx) = self.spend_txs.get(i) {
-                    let tx = psbt::PsbtState::new(self.wallet.clone(), tx.clone(), true);
+                    let recovery_timelock = tx
+                        .sigs
+                        .recovery_paths()
+                        .iter()
+                        .max_by(|(a, _), (b, _)| a.cmp(b))
+                        .map(|path| *path.0);
```
jp1ac4, line 128: It's a little simpler to do:
```diff
diff --git a/liana-gui/src/app/state/psbts.rs b/liana-gui/src/app/state/psbts.rs
index e46c16ce..6b459524 100644
--- a/liana-gui/src/app/state/psbts.rs
+++ b/liana-gui/src/app/state/psbts.rs
@@ -120,12 +120,7 @@ impl State for PsbtsPanel {
             }
             Message::View(view::Message::Select(i)) => {
                 if let Some(tx) = self.spend_txs.get(i) {
-                    let recovery_timelock = tx
-                        .sigs
-                        .recovery_paths()
-                        .iter()
-                        .max_by(|(a, _), (b, _)| a.cmp(b))
-                        .map(|path| *path.0);
+                    let recovery_timelock = tx.sigs.recovery_paths().keys().max().cloned();
                     let tx = psbt::PsbtState::new(
                         self.wallet.clone(),
                         tx.clone(),
```

jp1ac4 (commented):
Given the need to set `recovery_timelock` using `spend_tx.sigs` in a few places, and given that the logic is fairly simple (just taking the max timelock in `sigs.recovery_paths()`), I think it would be preferable and less error-prone to add a `recovery_timelock()` method on `SpendTx` using this logic rather than an additional field in `PsbtState`. In either case, I think the proper place for this field/method is `SpendTx` rather than `PsbtState`.

pythcoiner:
> Given the need to set `recovery_timelock` using `spend_tx.sigs` in a few places,

I dont think it's the right way to do it, see my [previous comment](https://github.com/wizardsardine/liana/pull/1675#discussion_r2092641592)

liana-gui/src/app/state/psbts.rs
thomask7b: Indeed simpler, thank  you for for this and the analysis for the recovery timelock. I understand better now why you wanted to remove it from PsbtState previously. And yes it makes sense, it's a better approach

liana-gui/src/app/state/psbts.rs
thomask7b: Thank you for the explanation I understand better now

PR #1685, opened 2025-04-30
gui: store SignModal state and allow to open export modal after signing

### LINKED ISSUE #1681: Issue with exporting / importing a PSBT after signing
If I upload / create a psbt and sign it, the button to export it (but also the one to import) don't work, which would be the expected flow in a collaborative signing flow:

https://github.com/user-attachments/assets/0daf771f-0b38-46e1-afc2-70648b9b1fc9

pythcoiner:
This PR fixes #1681:
- `PsbtState::modal` was not properly reset after signing modal closed
- When `PsbtState::modal` is of type `SignModal` we store it's state, else we lost track of which key already signed.
- `SignModal::display_modal` is not useful anymore, as we store the modal state, and only "hidding" the modal not let display another modal (export modal for instance)

### DIFF
--- a/liana-gui/src/app/state/psbt.rs
+++ b/liana-gui/src/app/state/psbt.rs
@@ -92,6 +92,9 @@ pub struct PsbtState {
     pub warning: Option<Error>,
     pub labels_edited: LabelsEdited,
     pub modal: Option<PsbtModal>,
+    // NOTE: sign_modal is used to store state of SignModal
+    // when another modal is selected
+    pub sign_modal: Option<PsbtModal>,
 }
 
 impl PsbtState {
@@ -102,6 +105,7 @@ impl PsbtState {
             labels_edited: LabelsEdited::default(),
             warning: None,
             modal: None,
+            sign_modal: None,
             tx,
             saved,
         }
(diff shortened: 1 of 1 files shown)


liana-gui/src/app/state/psbt.rs
```diff
@@ -165,19 +169,20 @@ impl PsbtState {
...
-                if let Some(PsbtModal::Sign(SignModal { display_modal, .. })) = &mut self.modal {
-                    *display_modal = false;
+                if matches!(self.modal, Some(PsbtModal::Sign(_))) {
+                    // store SignModal state
+                    self.sign_modal = self.modal.take();
+                    self.modal = None;
                     return Task::none();
                 }
-
-                self.modal = None;
```
jp1ac4, line 173: I think this line is still needed to close other modals, e.g. the broadcast modal.

pythcoiner: good catch, adressed & also dropped the unuseful `return Task::none();`

pythcoiner:
rebased

jp1ac4 (approved):
tACK 7281e6352967009419a493ccd3825887001afa14.

PR #1691, opened 2025-05-02
refac liana-gui: network flag redirects to launcher

edouardparis:
Adding the network flag to liana-gui command opens now the launcher with the network set instead of trying to run the wallet in the network directory.

This is preparatory work for the multiple wallet feature.


edouardparis (commented):
Self-ACK c4715acb047b48e882102e38b67ed9bb37816306

PR #1695, opened 2025-05-05
gui: be smart with window size at launch

### LINKED ISSUE #1546: [GUI] Default size of Liana should allow avoiding to scroll most screens
The default size of Liana application window (not the minimum, which should be unchanged) should be big enough for the main pages of the installer and wallet itself to not be scrolled. For example, with the current size the introduction pages of the templates in the installer and the "Send" page need often to be scrolled. The default size should avoid this as much as possible (even if for the Send page this depends on the number of coins present, which would suggest a review of its structure).

pythcoiner:
this PR fixes #1546:
 - at window startup we maximized the window size (that's the only way to know the area usable by our window)
 - if the height >= 1200, we resize the window at 1200x950 (we don't have to guess size of the task/title bar as we get the "usable area" from previous step
 - if the height < 1200 we rezise to the iced default window size
 
 from my test, maximazing + resize isn't visible for user, it appear that the window is directly open to its final size

### DIFF
--- a/liana-gui/src/main.rs
+++ b/liana-gui/src/main.rs
@@ -10,6 +10,7 @@ use iced::{
     event::{self, Event},
     keyboard,
     widget::{focus_next, focus_previous},
+    window::{self},
     Settings, Size, Subscription, Task,
 };
 use tracing::{error, info};
@@ -90,6 +91,8 @@ pub struct GUI {
     logger: Logger,
     // if set up, it overrides the level filter of the logger.
     log_level: Option<LevelFilter>,
+    id: Option<window::Id>,
+    window_init: Option<bool>,
 }
 
 enum State {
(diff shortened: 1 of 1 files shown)


liana-gui/src/main.rs
```diff
@@ -142,21 +147,58 @@ impl GUI {
...
         )
     }
 
     fn update(&mut self, message: Message) -> Task<Message> {
         match (&mut self.state, message) {
+            (_, Message::WindowSize(monitor_size)) => match (&self.window_init, &self.id) {
+                (Some(false), Some(id)) => {
+                    self.window_init = Some(true);
+                    let mut batch = vec![window::maximize(*id, false)];
+                    if monitor_size.height >= 1200.0 {
```
edouardparis, line 174: should it be monitor_size.width instead ?

pythcoiner:
closing this as the window maximising + resize seems visible on macOS & windows

PR #1696, opened 2025-05-05
feat: asynchronous management of internal bitcoind

edouardparis:


liana-gui/src/node/bitcoind.rs
```diff
@@ -498,22 +526,92 @@ impl Bitcoind {
...
     /// Stop (internal) bitcoind.
-    pub fn stop(&self) {
-        stop_bitcoind(&self.config);
+    pub fn stop(self) {
+        match self.lock.delete() {
+            Err(e) => {
+                tracing::error!("Failed to release bitcoind lock: {}", e);
+            }
+            Ok(false) => {
+                info!("Other process are using internal bitcoind. Process lock is deleted");
```
jp1ac4, line 535: nits:
- "Other processes are..."
- I think "Process lock has been deleted" reads better.

jp1ac4 (approved):
ACK 49251b50e56ecc4a9fd5d8624dc9f7ca3257b504.

Tested with new & existing wallets using managed bitcoind. I launched two GUI instances for the same wallet and was able to confirm that bitcoind was only stopped when closing the second instance (regardless of the order of opening).

pythcoiner:
not yet look at the code, are locks cleaned by the panic handler if needed?

pythcoiner:
image

my connexion is quite bad at the moment this screen can be displayed for ~20-30 secs, I feel it miss the information that the GUI is not frozen, maybe replace `Login` by `Connecting...` so the user can expect to wait?

edouardparis (commented):
Self-ACK 8c0c23e688b02a73393a7172b3f41bfaa2a9c143

PR #1699, opened 2025-05-06
gui: Improve visibility primary button text & fixes

### LINKED ISSUE #1697: Improve visibility of primary button text
The text inside primary buttons is not visible enough, at least on my screen. I believe we should improve primary button's style, for example by making the text field more thick, to improve readibility. This is especially important after merging #1693 

Attached, an example of how it looks now:

Image

thomask7b:
The first commit fixes #1697.
The second set "Go to replacement" as primary.
The third make "Generate address" become secondary after generation.
Tell me if I should squash.

image

Capture d’écran du 2025-05-06 20-47-25

I've tried to make the icon thicker but doesn't work the same way. Maybe using a different font or an image would help to fixes this.

### DIFF
--- a/liana-gui/src/app/view/receive.rs
+++ b/liana-gui/src/app/view/receive.rs
@@ -43,10 +43,15 @@ pub fn receive<'a>(
             Row::new()
                 .align_y(Alignment::Center)
                 .push(Container::new(h3("Receive")).width(Length::Fill))
-                .push(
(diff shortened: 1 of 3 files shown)


jp1ac4:
This needs a rebase on master but otherwise is looking good.

pythcoiner:
tACK d2c6f9bb

PR #1701, opened 2025-05-07
refactor: response has not succeed and add missing checks

### LINKED ISSUE #1679: Improve some Liana Connect error messages
There are some requests we make to Liana Connect for which we do not check the response status before attempting to parse the response, including for `create_spend_tx` and `create_recovery`.

We should go through all requests and make sure to check the response status before attempting to parse the response (see `broadcast_spend_tx` for an example).

thomask7b:
This should resolves #1679


liana-gui/src/services/connect/client/backend/mod.rs
```diff
@@ -1237,3 +1199,18 @@ fn spend_tx_from_api(
...
 }
+
+#[async_trait]
+pub trait ResponseExt {
+    async fn to_http_error<T>(self) -> Result<T, DaemonError>;
+}
+
+#[async_trait]
+impl ResponseExt for Response {
+    async fn to_http_error<T>(self) -> Result<T, DaemonError> {
```
jp1ac4, line 1210: This is a nice approach. Perhaps you could even change this method to something like `maybe_return_http_error()` and include the `if !response.status().is_success() { }` logic here? Then you'd only need `response.maybe_return_http_error().await?;` in each API method.

thomask7b: Thanks!
I've tried to do so but had difficulties with the borrow checker on `response`.
I'll try again and let you know if i'm not able to make it work

thomask7b: Done

liana-gui/src/services/connect/client/backend/mod.rs
```diff
@@ -1237,3 +1117,20 @@ fn spend_tx_from_api(
...
+
+#[async_trait]
+impl ResponseExt for Response {
+    async fn map_json_or_error<T: DeserializeOwned + Send>(self) -> Result<T, DaemonError> {
+        let status = self.status();
+        if !status.is_success() {
+            return Err(DaemonError::Http(Some(status.into()), self.text().await?));
+        }
+
+        self.json::<T>().await.map_err(DaemonError::from)
```
jp1ac4, line 1134: Possibly this step is causing the error you posted in Discord in some cases:
image

You might need a separate function to check the status (that either returns an error or the response itself), and then deserializing to JSON can be done in a separate step.

jp1ac4: Otherwise, the approach looks very nice.

thomask7b: Thanks!
Yes I didn't get that the issue comes from this change. Thank you for the suggestion I'm going to have a look into it

jp1ac4: For now, if you'll be returning either the response or an error, it's probably enough to use a regular function rather than defining a trait if that simplifies things.

thomask7b: I managed to fix using 2 functions, one to check the status one to map the json

PR #1705, opened 2025-05-15
Installer: let user choose derivation path

### LINKED ISSUE #1700: Add Account (i.e. derivation path) selection when setting a key from a hw device
It would be very useful for users to be able to select a different derivation path for the key imported from a hardware device in case they use it in other configurations. This would improve security and privacy practices, especially with the introduction of the multiple wallets functionality. In the current "Set keys" modal, the less invasive way of introducing it would probably be by having a little button inside the bigger button used to select the hardware device. This would show the "Account" (as for Sparrow) that we use by default (e.g. #0) and give the possibility to select a different one in a drop down menu. The menu would need to show somehow also the derivation path (e.g. 48'/0'/0'/2' for Account #0) for advanced users (it could also be on hover).

The process would look something like:

Selection step:

Image

Drop down menu:

Image

When selecting the same key in a different path, as for the alias, the Account number will automatically be the same selected the first time (we could either not show the option or disable it).

pythcoiner:
This PR closes #1700.

### DIFF
--- a/liana-gui/src/installer/step/descriptor/editor/key.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/key.rs
@@ -1,9 +1,9 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 
 use iced::{Subscription, Task};
-use liana::miniscript::bitcoin::bip32::Xpub;
+use liana::miniscript::bitcoin::bip32::{ChildNumber, Xpub};
 use liana::miniscript::{
     bitcoin::{
         bip32::{DerivationPath, Fingerprint},
@@ -89,6 +89,7 @@ pub struct EditXpubModal {
     hot_signer_fingerprint: Fingerprint,
     chosen_signer: Option<Key>,
     modal: Option<ExportModal>,
+    accounts: HashMap<Fingerprint, ChildNumber>,
 }
 
 impl EditXpubModal {
@@ -104,6 +105,10 @@ impl EditXpubModal {
         hot_signer_fingerprint: Fingerprint,
         keys: Vec<Key>,
     ) -> Self {
+        let accounts = keys
+            .iter()
+            .filter_map(|k| k.account.map(|acc| (k.fingerprint, acc)))
+            .collect();
         Self {
             device_must_support_tapminiscript,
             path_kind,
@@ -139,6 +144,7 @@ impl EditXpubModal {
             hot_signer,
             duplicate_master_fg: false,
             modal: None,
+            accounts,
         }
     }
 
@@ -158,6 +164,10 @@ impl super::DescriptorEditModal for EditXpubModal {
         self.duplicate_master_fg = false;
         self.error = None;
         match message {
+            Message::SelectAccount(fg, index) => {
+                self.accounts.insert(fg, index);
+                return Task::none();
+            }
             Message::Select(i) => {
                 if let Some(HardwareWallet::Supported {
                     device,
@@ -170,6 +180,11 @@ impl super::DescriptorEditModal for EditXpubModal {
                     self.processing = true;
                     self.form_key_source_kind = None;
                     let device_version = version.clone();
+                    let account = self
+                        .accounts
+                        .get(fingerprint)
+                        .copied()
+                        .unwrap_or(ChildNumber::from_hardened_idx(0).expect("hardcoded"));
                     let fingerprint = *fingerprint;
                     let device_kind = *kind;
                     let device_cloned = device.clone();
@@ -181,10 +196,11 @@ impl super::DescriptorEditModal for EditXpubModal {
                                 device_kind,
                                 fingerprint,
                                 network,
-                                get_extended_pubkey(device_cloned, fingerprint, network).await,
+                                get_extended_pubkey(device_cloned, fingerprint, network, account)
+                                    .await,
                             )
                         },
-                        |(device_version, device_kind, fingerprint, network, res)| {
+                        move |(device_version, device_kind, fingerprint, network, res)| {
                             Message::DefineDescriptor(message::DefineDescriptor::KeyModal(
                                 message::ImportKeyModal::FetchedKey(match res {
                                     Err(e) => Err(e),
@@ -198,6 +214,7 @@ impl super::DescriptorEditModal for EditXpubModal {
                                                 fingerprint,
                                                 name: "".to_string(),
                                                 key,
+                                                account: Some(account),
                                             })
                                         } else {
                                             Err(Error::Unexpected(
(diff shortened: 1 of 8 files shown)


pythcoiner:
actual state:
image

image

edouardparis:
I would not use the warning icon but the tooltip one (`icon::tooltip_icon`) and move it on the right of 
`Import an extended public key by selecting a signing device`.

Also I do not understand why incrementing account index makes the derivation path not standard.

pythcoiner:
> Also I do not understand why incrementing account index makes the derivation path not standard.

cc @bigspider

bigspider:
Not sure what do you mean? Account numbers for the various supported BIPs are considered standard by the Ledger app if not larger than 100.

pythcoiner:
> Not sure what do you mean? Account numbers for the various supported BIPs are considered standard by the Ledger app if not larger than 100.

I failed to export an xpub w/ m/48'/1'/`1'`/2'

image

and success if `display` == true

pythcoiner:
> I would not use the warning icon but the tooltip one (`icon::tooltip_icon`) and move it on the right of
> `Import an extended public key by selecting a signing device`.

image

pythcoiner:
fixed my mistake, the index stored was Unhardened

liana-ui/src/component/hw.rs
```diff
@@ -59,6 +60,63 @@ pub fn supported_hardware_wallet<'a, T: 'a, K: Display, V: Display, F: Display>(
     .padding(10)
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct Account {
+    pub index: u32,
```
jp1ac4, line 65: nit: it may be worth renaming as `hardened_index` to avoid doubt, or otherwise using `ChildNumber`.

liana-gui/src/installer/view/mod.rs
```diff
@@ -480,7 +484,13 @@ pub fn hardware_wallet_xpubs<'a>(
             if processing {
                 hw::processing_hardware_wallet(kind, version.as_ref(), fingerprint, alias.as_ref())
             } else {
-                hw::supported_hardware_wallet(kind, version.as_ref(), fingerprint, alias.as_ref())
+                hw::supported_hardware_wallet_with_account(
+                    kind,
+                    version.as_ref(),
+                    *fingerprint,
+                    alias.as_ref(),
+                    accounts.get(fingerprint).cloned().map(|c| c.into()),
```
jp1ac4, line 492: I think this `c.into()` conversion is not quite working as required. I think the function `supported_hardware_wallet_with_account` is expecting the hardened index starting from 0, whereas this is giving the index starting from 2**31.

liana-ui/src/component/hw.rs
pythcoiner: addressed

liana-gui/src/installer/view/mod.rs
```diff
@@ -1699,8 +1719,16 @@ pub fn hw_list_view<'a>(
                     alias.as_ref(),
                     "Device firmware version does not support taproot miniscript",
                 )
+            } else if let Some(accounts) = accounts {
+                hw::supported_hardware_wallet_with_account(
+                    kind,
+                    version.as_ref(),
+                    *fingerprint,
+                    alias.as_ref(),
+                    accounts.get(fingerprint).cloned(),
```
jp1ac4, line 1728: nit: could also use `.copied()`

liana-gui/src/installer/step/descriptor/editor/key.rs
```diff
@@ -576,14 +589,32 @@ pub fn default_derivation_path(network: Network) -> DerivationPath {
     .unwrap()
 }
 
+pub fn derivation_path(network: Network, account: ChildNumber) -> DerivationPath {
+    assert!(account.is_hardened());
```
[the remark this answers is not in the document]
jp1ac4: nit: I feel it might be slightly better to return an `Option` or `Result` here and let the caller unwrap it.

liana-ui/src/component/hw.rs
```diff
@@ -59,6 +60,77 @@ pub fn supported_hardware_wallet<'a, T: 'a, K: Display, V: Display, F: Display>(
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct Account {
+    pub hardened_index: ChildNumber,
+    pub fingerprint: Fingerprint,
+}
+
+impl Account {
+    pub fn new(index: ChildNumber, fingerprint: Fingerprint) -> Self {
+        Self {
+            hardened_index: index,
```
jp1ac4, line 72: nit: possibly there should be a check that the `index` argument is hardened given the field name, or otherwise I think you can rename the field back to `index`, which should be fine now that the type is `ChildNumber`.

liana-gui/src/installer/view/editor/mod.rs
```diff
@@ -353,6 +353,18 @@ pub fn edit_key_modal<'a>(
...
+        .push(Space::with_height(5))
+        .push(widget::tooltip::Tooltip::new(
+            icon::tooltip_icon(),
+            "Switch account if you already use the same hardware in other configurations",
+            widget::tooltip::Position::Bottom,
+        ));
+    let source = Row::new()
+        .push(p1_regular("Select the source of your key").bold())
+        .push(Space::with_width(10))
+        .push(info)
```
jp1ac4, line 366: nit: I feel this tooltip would be more visible if it were next to the pick list within the HW card. Same applies for "Share Xpubs" below.

jp1ac4 (approved):
tACK 7d6fce8a1df6c23f30f0dff412efa427bf200fa2.

Tested with Ledger and Specter DIY. Added comments for some nits.

liana-gui/src/installer/view/editor/mod.rs
pythcoiner: this have been request by @edouardparis previously and ACK'd by @nondiremanuel iirc

pythcoiner:
> nit: I feel it might be slightly better to return an Option or Result here and let the caller unwrap it.

I don't think it worth bother w/ Option/Result here, it's intended to catch an error at implementation time not at run time

jp1ac4 (approved):
tACK 386c14ca8d49299636b5b956efc77f1439c076af.

liana-ui/src/component/hw.rs
pythcoiner: `hardened_index` => `index`

pythcoiner:
added display of account after xpub fetched:

image

pythcoiner:
> If I choose the same key in a different path, the selected account is shown as 0 regardless of the selection made in the other path. So I think the account parameter is not being set in all places. Also, but less importantly, in the key list view, when viewing HWs, the account is not shown

fixed these 2 issues

pythcoiner:
> the "Register descriptor" step is showing the "Account #0"

fixed this issue

jp1ac4 (approved):
tACK ce8e1f3c3eed5a6e991307f8740e54cb0680e5bc.

edouardparis (approved):
ACK ce8e1f3c3eed5a6e991307f8740e54cb0680e5bc

PR #1704, opened 2025-05-13
feat: List unused addresses on Receive panel

jp1ac4:
This implements #809 by listing unused addresses on the Receive panel.

PR #1707, opened 2025-05-15
[GUI] some fixes for unused addresses list

jp1ac4:
This PR consists of two fixes following #1704:
1) don't show the address with index 0 in the list of unused addresses. This is because a new Liana wallet starts generating new addresses from index 1.
2) hide the option to view unused addresses if there aren't any (such as for a new wallet).


pythcoiner:
tACK efd0c32e

PR #1709, opened 2025-05-16
fix: use cookie file auth for new wallets with managed bitcoind

### LINKED ISSUE #1708: New wallet with local bitcoind managed by liana should use cookie instead of authrpc
Since the introduction of bitcoind clients monitoring with the `./liana/bitcoind/locks` directory, multiple liana processes are aware if they should start or stop the liana managed bitcoind. 
We can reintroduce again the use of the .cookie file for authentication for new wallet.

jp1ac4:
This fixes #1708.

There would be no way to get the RPC password of an existing managed bitcoind installation without looking in other daemon.toml files.

An existing rpcauth field in the bitcoin.conf file will remain unchanged so that an existing wallet using username/password authentication will continue to work.

### DIFF
--- a/liana-gui/src/installer/mod.rs
+++ b/liana-gui/src/installer/mod.rs
@@ -12,7 +12,7 @@ use liana_ui::{
     component::network_banner,
     widget::{Column, Element},
 };
-use lianad::config::Config;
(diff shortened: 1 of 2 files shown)


edouardparis (approved):
ACK 63ef4838a705a6f9187cf607a4e320b71f374557

PR #1713, opened 2025-05-19
Update wallet alias

edouardparis:
We had a new `wallet_alias` to settings file.
This field is displayed in the launcher when wallets are listed and in the window title when the app is running the wallet.

The field is also store in liana-connect wallet.metadata field to facilitate import of wallet from Liana-Connect.
If the field was modified on Liana-Connect, then next time the user open the wallet in the gui, the settings field is updated.

User can modify the wallet alias in the Settings > Wallet section.

### DIFF
--- a/liana-gui/src/app/state/settings/wallet.rs
+++ b/liana-gui/src/app/state/settings/wallet.rs
@@ -29,6 +29,7 @@ use crate::{
     dir::LianaDirectory,
     export::{ImportExportMessage, ImportExportType},
     hw::{HardwareWallet, HardwareWalletConfig, HardwareWallets},
+    services::connect::client::backend::api::WALLET_ALIAS_MAXIMUM_LENGTH,
 };
 
 enum Modal {
@@ -168,7 +169,7 @@ impl State for WalletSettingsState {
             Message::View(view::Message::Settings(view::SettingsMessage::WalletAliasEdited(
                 alias,
             ))) => {
-                self.wallet_alias.valid = alias.len() < 64;
+                self.wallet_alias.valid = alias.len() < WALLET_ALIAS_MAXIMUM_LENGTH;
                 self.wallet_alias.value = alias;
                 Task::none()
             }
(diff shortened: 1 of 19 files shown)


liana-gui/src/installer/view/mod.rs
```diff
@@ -2141,6 +2146,43 @@ pub const REMOTE_BACKEND_DESC: &str = "Use our service to instantly be ready to
...
 
+pub fn wallet_alias<'a>(
+    progress: (usize, usize),
+    email: Option<&'a str>,
+    wallet_alias: &form::Value<String>,
+) -> Element<'a, Message> {
+    layout(
+        progress,
+        email,
+        "Give a alias to your wallet",
```
jp1ac4, line 2157: "an alias"
nit: I have a preference for "Give your wallet an alias" but either is fine.

liana-gui/src/installer/view/mod.rs
```diff
@@ -2141,6 +2146,43 @@ pub const REMOTE_BACKEND_DESC: &str = "Use our service to instantly be ready to
...
+                Column::new()
+                    .spacing(20)
+                    .push(p1_bold("Wallet alias:"))
+                    .push(
+                        form::Form::new("Wallet alias", wallet_alias, Message::WalletAliasEdited)
+                            .warning("Wallet alias is too long.")
+                            .size(text::P1_SIZE)
+                            .padding(10),
+                    )
+                    .push(p2_regular("You will be able to change it later")),
```
jp1ac4, line 2169: Perhaps worth mentioning it can be changed in the Settings > Wallet page?

liana-gui/src/app/state/settings/wallet.rs
```diff
@@ -159,6 +165,13 @@ impl State for WalletSettingsState {
                     Task::none()
                 }
             }
+            Message::View(view::Message::Settings(view::SettingsMessage::WalletAliasEdited(
+                alias,
+            ))) => {
+                self.wallet_alias.valid = alias.len() < 64;
```
jp1ac4, line 171: As the `64` is also used in the installer, could it be stored in a variable somewhere and used in both places?

jp1ac4 (commented):
This is working nicely with both local and remote backends.

Just a few comments.

PR #1716, opened 2025-05-21
fix: use scrollbar for node settings

jp1ac4:
This fixes an issue when viewing a long node setting value, such as the cookie path, on a reduced window size.

image


edouardparis (approved):
utACK 01ceb237f8446a31653bc294142a3a1397128373

PR #1717, opened 2025-05-21
fix settings: missing update_aliases for backup restore

edouardparis:
While restoring a backup at the import/export panel, the message to update the wallet keys aliases was not handled and although the aliases are stored in the settings file, the wallet was not "reloaded" through the Message::WalletUpdated result. The consequence was while opening the wallet section, users found the aliases not updated by the backup (even if the file was updated).

The change in the update_aliases is to update the wallet alias only if the function require the change by passing
`Some(alias)` in arg.


jp1ac4 (approved):
utACK 654be987f2da61ed47310c3d30720126b64aa292.

PR #1718, opened 2025-05-21
Clear descriptor cache

### LINKED ISSUE #1715: Add "Clear" button in the descriptor setup step
We should have a "Clear" button in the descriptor setup ("Set keys") step. This button should let the user reset the screen and start from scratch (including clearing the cache). This is particularly useful with v11 since we added the account selection which is not editable other than deleting the specific key for all paths. Since if such key is the only key in the primary path we don't allow to delete it, the user should be able to do it by clearing the template.

The "Clear" button will be placed on the same line of the "Continue" one but on the opposite (left) side. Some vertical space will be added to better separate it from the buttons above.

Image

pythcoiner:
closes #1715 

image

image

image

### DIFF
--- a/liana-gui/src/installer/descriptor.rs
+++ b/liana-gui/src/installer/descriptor.rs
@@ -1,5 +1,8 @@
 use async_hwi::{DeviceKind, Version};
-use liana::miniscript::{bitcoin::bip32::Fingerprint, descriptor::DescriptorPublicKey};
+use liana::miniscript::{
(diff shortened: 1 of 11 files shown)


pythcoiner:
updated button spacing/size:

image

jp1ac4 (approved):
tACK a51704fcc66fcfa7e5bc133eaf2f7c406ee19e8b.

PR #1720, opened 2025-05-23
show partial address in collapsed view

jp1ac4:
To avoid the need for a scrollbar when showing previous addresses in the collapsed view, we will now show the first and last 16 characters only.

image
image


edouardparis (approved):
utACK 78a261da5a5e92f3589403760858105026e6278d

PR #1721, opened 2025-05-23
Apply backup alias when importing backup in installer.

edouardparis:
- We add a default wallet alias if not value is set in the installer context.
- We add the wallet alias field to the backup file. 
  - While importing a backup in the installer, the backup wallet alias will be suggested to the user at the final step. 
  - While importing a backup in the wallet settings for an already installed wallet, we keep the current wallet alias and not override it by the backup one.

### DIFF
--- a/liana-gui/src/backup.rs
+++ b/liana-gui/src/backup.rs
@@ -50,6 +50,8 @@ fn now() -> u64 {
 pub struct Backup {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub alias: Option<String>,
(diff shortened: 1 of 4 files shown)


liana-gui/src/backup.rs
```diff
@@ -157,13 +160,15 @@ impl Backup {
         let keys = wallet.keys();
 
         let network_dir = datadir.network_directory(network);
+        let mut wallet_alias = Option::<String>::None;
```
jp1ac4, line 163: Could this value be taken from `wallet`?

edouardparis: ah right, I can directly take it from it

jp1ac4 (approved):
tACK 250ba8bb45d8b883772fe5e49e7b48f2a827306b.

PR #1729, opened 2025-05-27
fix update wallet alias in wallet settings

edouardparis:
L522 the wallet was populated with its new alias
and the check L567 was always false.

wallet_alias is None if we do not have to update it, so better pass the variable directly.


edouardparis (commented):
Self-ACK 546ac29587e831abbe79e66d23cac9683467a842

PR #1740, opened 2025-06-02
Smart screen size

### LINKED ISSUE #1546: [GUI] Default size of Liana should allow avoiding to scroll most screens

pythcoiner:
closes #1546 following last requirements:

> Currently the consensus has been on the following solution: resize the window according to a logic similar to https://github.com/wizardsardine/liana/pull/1695 at the first launch and then remember the last size for following launches of the application.

Summary of the feature introduced in this PR:
- The first time Liana launch, the window is maximized in order to "guess" the maximum usable size on the monitor, and then we apply [this logic](https://github.com/wizardsardine/liana/issues/1546#issuecomment-2750974470) for deciding the default size.
- The default size is recorded in `global_settings.json` (so this aplly to ALL wallets)
- On nexts Liana launch, the window is resized to the value contained in `global_settings.json`

I added something not clearly defined: if the user resize its screen, we reapply the last screen size at startup.


liana-gui/src/app/settings.rs
```diff
@@ -386,8 +387,63 @@ pub mod global {
...
+    }
+
+    #[derive(Debug, Deserialize, Serialize, Default)]
+    pub struct GlobalSettings {
         pub bitbox: Option<BitboxSettings>,
+        pub window_config: Option<WindowConfig>,
+    }
+
+    impl GlobalSettings {
+        pub fn datadir(global_datadir: &LianaDirectory) -> PathBuf {
```
jp1ac4: Should this have a different name as it returns the path of the settings file rather than a directory?

Also, should it be a method on `GlobalSettings` or just a function in the `global` module given that it doesn't depend on `GlobalSettings`? Or perhaps it could be a method on `LianaDirectory` instead.

liana-gui/src/main.rs
```diff
@@ -97,6 +101,9 @@ pub struct GUI {
     logger: Logger,
     // if set up, it overrides the level filter of the logger.
     log_level: Option<LevelFilter>,
+    id: Option<window::Id>,
```
jp1ac4: nit: Could this be renamed to `window_id`?

jp1ac4 (commented):
Tested 507508e2fe31143a8cecd8b2734f9e14afb8f8a7 on Ubuntu. The initial window maximisation is noticeable. Subsequently, the previous window size is remembered.

Left some comments for consideration.

liana-gui/src/app/settings.rs
pythcoiner: indeed `datadir()` is not a good naming, maybe just call it `path()` is better?

but I think it should remain a (static) method of `GlobalSettings` as it returns the path where to fetch the `GlobalSettings` serialization

jp1ac4: I think `path()` is better.

liana-gui/src/app/settings.rs
```diff
@@ -386,8 +387,63 @@ pub mod global {
...
+        }
+
+        pub fn load_window_config(path: &PathBuf) -> Option<WindowConfig> {
+            Self::load(path)
+                .ok()
+                .flatten()
+                .and_then(|s| s.window_config)
+        }
+
+        pub fn update_window_config(
```
jp1ac4: This method doesn't seem to be used anywhere currently.

jp1ac4 (commented):
Did you consider saving the screen size to file only when closing the GUI? Otherwise, there could be writes for all intermediate window sizes when resizing.

pythcoiner:
> Did you consider saving the screen size to file only when closing the GUI? Otherwise, there could be writes for all intermediate window sizes when resizing.

I dont see any issue to update the state for intermediate sizes?

jp1ac4:
> I dont see any issue to update the state for intermediate sizes?

It's probably fine, but if it's possible to write to the file just once without changing too much code, I think that would be preferable.

liana-gui/src/app/settings.rs
pythcoiner: adressed

liana-gui/src/app/settings.rs
pythcoiner: adressed, I dont understand why clippy were not yelling...

liana-gui/src/main.rs
```diff
@@ -150,21 +171,115 @@ impl GUI {
...
+                            let path = GlobalSettings::path(self.state.datadir_path());
+                            let mut settings = GlobalSettings::load(&path)
+                                .ok()
+                                .flatten()
+                                .unwrap_or_default();
+                            settings.window_config = Some(WindowConfig {
+                                width: *width,
+                                height: *height,
+                            });
+                            let _ = settings.to_file(&path);
```
edouardparis: I agree that we do not need to show any error to the user, but I think at least we should log them.

pythcoiner:
> I agree with Michael, I prefer we only write once the file.
> 
> > by doing this I'm pretty sure we have an edge case were we can overwrite the bitbox pairing code: the hw logic do not access the context, it use the file to store the state, that's why I've done this way...
> 
> You can add a update method for the global settings method that uses the same logic that the new one for the settings.file, it adds a recovery lock and make sure the change is encapsulated in a read/write workflow: https://github.com/wizardsardine/liana/blob/master/liana-gui/src/app/settings.rs#L48
> 
> It is better also to be consistent and we are not in a hurry.

adressed

liana-gui/Cargo.toml
```diff
@@ -53,6 +53,7 @@ bitcoin_hashes = "0.12"
 reqwest = { version = "0.11", default-features=false, features = ["json", "rustls-tls", "stream"] }
 rust-ini = "0.19.0"
 rfd = "0.15.1"
+fs2 = "0.4.3"
```
edouardparis, line 57: we already have a lib for advisory lock on write.
https://github.com/wizardsardine/liana/blob/master/liana-gui/src/app/settings.rs#L66

pythcoiner: `NoiseConfig::read_config()` & `NoiseConfig::store_config()` cannot call async functions...

edouardparis:
Sorry it seems that the never the config is written when I am starting liana-gui with a new directory.

pythcoiner:
> Sorry it seems that the never the config is written when I am starting liana-gui with a new directory.
what do you mean by "with a new directory"? you mean `empty`?

edouardparis:
Yes and it seems from a ctrl+f that this function is never used:
https://github.com/wizardsardine/liana/pull/1740/files#diff-028568692e3034ffe6d0f318c0d7800a7b980395783715af50b16685a804cb41R416

pythcoiner:
yes my bad, i should no have been carrefull while rebasing, some logic have been wiped

pythcoiner:
fixed

edouardparis:
ACK 1740

PR #1745, opened 2025-06-12
fix: use correct path to daemon.toml in settings

edouardparis:
backport of #1744


edouardparis (commented):
Self-ACK d16dbf8d4b6c71947f628df37b8a75e011fdb7fb

PR #1749, opened 2025-06-16
refac: remove patch field from Version

edouardparis:
The Version struct is used across the software
to display the release version. A release version
does not include patch number as we increment
only the major and the minor during a new release.

For very small patch that does not change the
API and does not require a new release, then the crates Cargo.toml version patch number will be incremented.


jp1ac4 (approved):
tACK fef70879b9507e0429a8a0a5211ba0547b68cc92.

This also means the `version` field of the `getinfo` response will follow the SimVer format as documented in https://github.com/wizardsardine/liana/blob/master/doc/API.md#getinfo.

PR #1750, opened 2025-06-17
refac: remove patch field from Version

edouardparis:
backport of #1749 

The Version struct is used across the software
to display the release version. A release version
does not include patch number as we increment
only the major and the minor during a new release.

For very small patch that does not change the
API and does not require a new release, then the crates Cargo.toml version patch number will be incremented.


edouardparis (commented):
Self-ACK 66beb1397f45712f7146655c2f6cbb53f278fc24

PR #1756, opened 2025-06-23
Add an option for add a descriptor from file/backup for liana-connect

### LINKED ISSUE #1751: Include restore via backup file using Liana Connect in the add an existing Liana wallet flow
A user can currently just enter the descriptor, we should let them have the possibility to upload a backup file:

Image

The section should become:
```

**Load a wallet from backup file / descriptor**
Upload the wallet backup file or paste its descriptor to add it

"Import descriptor" button

or

"Descriptor"
(paste area)
```

This work is important since we won't have the backend switch in the settings but we will indicate the user how to restore easily the wallet with a new backend/node option. Using the file is much simpler, easy to understand and we should standardize it for all backend options.

pythcoiner:
This PR closes #1751:
 - add the option to import a descriptor from a file.
 - add a spacer at the bottom of the screen.
 - add the possibility to import also a descriptor from a backup file.

### DIFF
--- a/liana-gui/src/backup.rs
+++ b/liana-gui/src/backup.rs
@@ -1,7 +1,10 @@
 use chrono::{Duration, Utc};
-use liana::miniscript::{
-    self,
-    bitcoin::{bip32::Fingerprint, Network, Txid},
+use liana::{
+    descriptors::LianaDescriptor,
+    miniscript::{
+        self,
+        bitcoin::{bip32::Fingerprint, Network, Txid},
+    },
 };
 use lianad::{
     bip329,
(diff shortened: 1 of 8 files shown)


liana-gui/src/main.rs
```diff
@@ -164,6 +164,7 @@ impl GUI {
     }
 
     fn update(&mut self, message: Message) -> Task<Message> {
+        log::info!("{message:?}");
```
edouardparis: Forgot to remove it

pythcoiner: cleaned

edouardparis (approved):
ACK 31798b7e4317125e671e2d52e28e9ddc16815bf0

PR #1759, opened 2025-06-28
chore: fix some minor issues in the comments

cloudclaim:
fix some minor issues in the comments

### DIFF
--- a/liana-gui/src/services/connect/client/backend/api.rs
+++ b/liana-gui/src/services/connect/client/backend/api.rs
@@ -257,7 +257,7 @@ pub struct Transaction {
     pub raw: bitcoin::Transaction,
     pub inputs: Vec<Input>,
     pub outputs: Vec<Output>,
-    /// If the transaction has multiple incoming or ougoing payment.
+    /// If the transaction has multiple incoming or outgoing payments.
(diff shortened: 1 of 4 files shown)


liana-gui/src/services/connect/client/backend/api.rs
```diff
@@ -257,7 +257,7 @@ pub struct Transaction {
     pub raw: bitcoin::Transaction,
     pub inputs: Vec<Input>,
     pub outputs: Vec<Output>,
-    /// If the transaction has multiple incoming or ougoing payment.
+    /// If the transaction has multiple incoming or outgoing payment.
```
jp1ac4: Could you please also change "payment" -> "payments".

jp1ac4 (commented):
Thanks very much. There's one additional small change that would be nice to include.

liana-gui/src/services/connect/client/backend/api.rs
cloudclaim: Of course. Modified.

Please review again.

jp1ac4 (approved):
ACK ac1043a1e09ec1570c489d537fbcff3ebb4c7152.

Thanks for including the additional fix.

PR #1760, opened 2025-06-29
Improve timelock duration selection and display

thomask7b:
The first commit adds monthly steps to the timelock selection slider and adapts time estimation precision.
The second is a tiny GUI improvement.

Here few examples : 

Capture d’écran du 2025-06-29 19-22-16
Capture d’écran du 2025-06-29 19-22-12
Capture d’écran du 2025-06-29 19-22-01
Capture d’écran du 2025-06-29 19-21-56
Capture d’écran du 2025-06-29 19-21-34

### DIFF
--- a/liana-gui/src/installer/view/editor/mod.rs
+++ b/liana-gui/src/installer/view/editor/mod.rs
@@ -524,7 +524,7 @@ fn example_xpub(network: Network) -> String {
 }
 
 /// returns y,m,d,h,m
-pub fn duration_from_sequence(sequence: u16) -> (u32, u32, u32, u32, u32) {
+fn duration_from_sequence(sequence: u16) -> (u32, u32, u32, u32, u32) {
     let mut n_minutes = sequence as u32 * 10;
     let n_years = n_minutes / 525960;
     n_minutes -= n_years * 525960;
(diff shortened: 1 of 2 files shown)


thomask7b:
This example is probably more interesting, I puted manually 31000 :

Capture d’écran du 2025-06-30 17-02-50

thomask7b:
Thank you for your feeback.
How it is now? I just updated

liana-gui/src/installer/view/editor/mod.rs
```diff
@@ -585,18 +593,21 @@ pub fn edit_sequence_modal<'a>(sequence: &form::Value<String>) -> Element<'a, Me
...
+                    },
+                ))
                 .push(
                     Container::new(
                         slider(1..=u16::MAX, sequence, |v| {
                             Message::DefineDescriptor(
                                 message::DefineDescriptor::ThresholdSequenceModal(
-                                    message::ThresholdSequenceModal::SequenceEdited(v.to_string()),
+                                    message::ThresholdSequenceModal::SequenceEdited(
+                                        (if v > 1 && v != u16::MAX { v - 1 } else { v })
```
jp1ac4, line 629: What was the reason for this change?

thomasballivet: It makes the intermediate steps at the appropriate value. For example the first step is one but the second will be 4384 instead of 4383 without this line. So because we start at 1 we need to minus 1 each intermediate value. It's is just to have the most accurate value to represent n months.
But maybe it's overelaborate and the tiny difference can be acceptable

jp1ac4: OK thanks for explaining that. I think it would be worth adding a brief comment in the code.

thomask7b:
Here is an example : 

Capture d’écran du 2025-07-02 15-48-56

nondiremanuel:
> Thanks for the review @jp1ac4 . As I think the following is relevant I already processed the changes, within another commit because there is a refactoring and waiting for the confirmation of @nondiremanuel .
> 
> > @nondiremanuel Should we use the same time granularity in this view:
> > image

Yes, it's definitely needed, thanks for noticing and implementing.

jp1ac4 (approved):
tACK 749be2ec9eac16f2596a7bdf57f5ecbb82ab5465. Thanks!

PR #1766, opened 2025-07-04
fix: skip syncing screen if node is not bitcoind

### LINKED ISSUE #1256: [GUI] Loader - Change wording depending on node type
When loading the GUI, the message shown contains "Bitcoin Core is synchronising the blockchain".

We should either use generic wording or something that depends on the node type (Bitcoin Core / Electrum server).

jp1ac4:
Fixes #1256.

The loading screen showing the blockchain sync progress will skipped if the node type is not Bitcoin Core.


edouardparis (approved):
utACK ba96bb6c063641debcbe2baf2c0f8e52e133f89a

PR #1767, opened 2025-07-06
liana-ui: add warning to form::Value

pythcoiner:
This PR add a `warning` field to `form::Value`, so the warning can be accessed also from outside the view scope. This field will be "inherited" by the `form::Form`.

This PR "closes" a TODO in `EditXpubModal`, but I decided to go with a separate `warning` field as it can be useful to display a warning even if a value is considered valid.

Note: as this change do not modify the usage from within the view scope, usages of `form::Form` have not been updated.


edouardparis:
utACK d9c540ad1442bc7787fe33b69b2e60b62e17a650
Looks good to me and I agree with the idea.

PR #1772, opened 2025-07-09
fix(poller): only update last used index for affected keychain

### LINKED ISSUE #1333: poller: check if coin is change when setting DB's next derivation index
See https://github.com/wizardsardine/liana/pull/1222#discussion_r1740264547.

We can check if the received coin is change or not so that we only set the appropriate next derivation index.

(I can't think of any reason not to make this change)

jp1ac4:
This fixes #1333 by only setting the last used index of the affected keychain.


edouardparis (approved):
utACK cb60089ace858772446ba82dcb76fc082a13242e

PR #1780, opened 2025-07-21
fix: a wallet with remote auth does not use managed bitcoind

jp1ac4:
This is a follow-up to #1746 to address the case where the wallet uses Liana Connect, in which case `start_internal_bitcoind` is set to `null` in the settings file.


edouardparis (approved):
utACK d86d6852b93013df15e914f43a8e631dd7cbe084

PR #1790, opened 2025-07-24
fix: change wording loader go back button

### LINKED ISSUE #1789: Change button text when Liana fails to start
The "Use another Bitcoin network" text is not valid anymore, especially with multiple wallets. We should just have "Back" instead:

Image

edouardparis:
close #1789


edouardparis (commented):
Self-ACK e52b3f192e5b9925ae5baf737cc82b7801bd9d37

PR #1792, opened 2025-07-24
fix: change wording loader go back button

### LINKED ISSUE #1789: Change button text when Liana fails to start

edouardparis:
close #1789


edouardparis:
Self-ACK 7cf9ebf9a477b355caf64d684f1a0da7331f5e30

PR #1794, opened 2025-07-24
Add message that wallet was deleted to wallet member

### LINKED ISSUE #1793: Add an error when trying to load a wallet deleted by the owner/creator on Liana Connect
Currently, when a participant of a wallet tries to load a Liana Connect wallet deleted for all members by its creator, he gets stuck in a perpetual "Liana Connect" loading screen. We should add an error with a message explaining the situation to the user. Something like: 
`This wallet was deleted by its creator for all participants and cannot be opened.`
`To access it again, restore it using a backup file or the wallet descriptor.`

This should be valid for all situations in which a Liana Connect wallet is locally present but deleted on the Liana Connect DB.

https://github.com/user-attachments/assets/396a64a7-53ae-4863-b711-9ea4c45e30fe

edouardparis:
close #1793

### DIFF
--- a/liana-gui/src/services/connect/login.rs
+++ b/liana-gui/src/services/connect/login.rs
@@ -129,6 +129,7 @@ pub struct LianaLiteLogin {
 pub enum ConnectionStep {
     CheckingAuthFile,
     CheckEmail,
+    WalletDoesNotExist,
     EnterOtp {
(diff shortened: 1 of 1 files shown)


pythcoiner:
tACK 6273a806

Issue #1798, opened 2025-07-31
Optionally show current fiat price in Home


nondiremanuel:
We should give the user the option to see how much their balance corresponds in fiat terms, as basically any wallet does. The fiat balance should be in parenthesis close to the BTC one, also for the unconfirmed part.
The fiat price option should be configurable in a new Settings > General section, in which the user will be able to enable/disable such feature and select the currency among the available. At this stage, we will only take the price from one source and we will show in this page which is it. When/if we will add more, we will probably give the possibility to the user to choose which to use (this is part of a separate issue).

See [here](https://miro.com/app/board/uXjVKvwKmIs=/?moveToWidget=3458764635984364636&cot=14) for some mockups.

Since we don't want to compromise privacy for those users that want to keep full self-sovereignity, I would propose that:
- Fiat price option is "ON" by default only for users that use Liana Connect as a backend (they already expressed a preference towards usability);
- It is off by default for other backend options.

For the latter case, we could have a banner / notification telling them that they can activate it, redirecting to the settings (not part of this issue, I will open a separate one).

jp1ac4:
We could also add a step to the Installer to enable/disable fiat prices when creating a new wallet.

jp1ac4:
For Liana Connect users, I suppose we would default to USD?

PR #1808, opened 2025-08-11
refactor: update cache fields according to source

jp1ac4:
This PR is a preparatory step towards #1798.

If we want to cache values from different sources and update frequencies, e.g. when adding fiat prices, I think it will be convenient to group and update them independently of each other.

This PR adds a `DaemonCache` struct within `Cache` for storing and updating dynamic daemon values only.

The `UpdateDaemonCache` message replaces `UpdateCache` to update these daemon cache values only.

The new `CacheUpdated` message will be called when any fields within the cache have been updated, e.g. daemon values, fiat price values, etc.

### DIFF
--- a/liana-gui/src/app/cache.rs
+++ b/liana-gui/src/app/cache.rs
@@ -13,14 +13,9 @@ use std::sync::Arc;
 pub struct Cache {
     pub datadir_path: LianaDirectory,
     pub network: Network,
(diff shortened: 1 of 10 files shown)


edouardparis (approved):
utACK dff3abe0bd2cfe7cd64a1faaae0109196092f3cd

PR #1811, opened 2025-08-12
Some preparatory refactoring

jp1ac4:
This PR includes the following refactoring that will be useful as part of #1798:

1. Move `deser_fromstr` function to a new serde utils module.
2. Add a utils function to get the current time (as a `Duration` since the epoch).
3. Move the `ResponseExt` trait to a new http module and use it to deduplicate logic.

Note for (3) that I removed the `json_or_error` trait method from the `ResponseExt` trait. This is because this can be
achieved fairly simply by using `Ok(...?)`. Trying to include it in the http module with a generic error type could then lead to type annotations being needed when using it due to several error types implementing `From<reqwest::Error>`. I felt it was therefore simpler to use the `Ok(...?)` approach.


edouardparis (approved):
utACK df2d45b2664b47759329666b36e5907e35fcb16b

PR #1830, opened 2025-09-01
installer: switch order of key/devices in the key selection modal

### LINKED ISSUE #1827: Switch position of the first two sections in the "set key source" modal
Differently to my first mockup, I think the "Detected hardware" should be on top of everything (while currently it's the "Already used sources" on top), since if I plug in a hardware it is probably the most likely option I need to select.

Image

pythcoiner:
closes #1827

image


edouardparis (approved):
ACK 0b11d36cfcebe941d30fd6ba832580afdc3f042f

Issue #1819, opened 2025-08-28
Make fiat price cache & setting global?


jp1ac4:
#1815 added a wallet-specific fiat price cache & setting, with each wallet making its own price API requests.

A user with multiple wallet tabs open might then send multiple requests for the same fiat price in parallel. An alternative would be to send one request per source/currency pair configured across all open wallets and store the values in some sort of global cache, which would then be filtered down to the relevant wallets.

Furthermore, rather than requiring the user to configure the fiat price setting per wallet, there could be a global setting for a single source/currency pair. The only downside here would be in case a user wanted a different currency in some wallets.

nondiremanuel:
As we were discussing, we can keep the option per wallet and possibly find a way to optimize the calls using a cross-wallet cache or something.

PR #1833, opened 2025-09-02
store fiat prices in new global cache

jp1ac4:
This PR addresses #1819 by adding a new global cache to store fiat prices.

This global cache will be used to avoid making multiple API requests for the same currency in case several tabs are open. It will also store the list of available currencies for a given price source to avoid multiple requests when accessing the settings page.

The fiat price settings and iced subscriptions are still handled by each individual wallet.

When a new price is fetched, it will be sent to all tabs that require it.


jp1ac4:
I will replace this with another PR following the new global tick logic.

PR #1834, opened 2025-09-02
Add global tick subscription

edouardparis:
In order to reduce back pressure with multiple pane open, only one subscription is providing a tick every 1 second. It is up to each state of each panel to handle this tick.

For example the app state will compare the tick instant to the last time it had updated the daemon cache to know if it has to update it again.

### DIFF
--- a/liana-gui/src/app/mod.rs
+++ b/liana-gui/src/app/mod.rs
@@ -14,7 +14,7 @@ use std::io::Write;
 use std::sync::Arc;
 use std::time::Duration;
 
-use iced::{clipboard, time, Subscription, Task};
+use iced::{clipboard, Subscription, Task};
 use tokio::runtime::Handle;
 use tracing::{error, info, warn};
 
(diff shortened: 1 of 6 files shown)


liana-gui/src/app/mod.rs
```diff
@@ -363,31 +332,71 @@ impl App {
...
+                Ok(DaemonCache {
+                    blockheight: info.block_height,
+                    coins: coins.coins,
+                    rescan_progress: info.rescan_progress,
+                    sync_progress: info.sync,
+                    last_poll_timestamp: info.last_poll_timestamp,
+                    last_tick: tick,
+                })
+            },
+            Message::UpdateDaemonCache,
```
jp1ac4, line 394: I don't think it's essential, but the `UpdateDaemonCache` handler could check that the `last_tick` value matches the expected value.

jp1ac4 (commented):
Tested and it's working as expected. There's an unused import causing clippy to fail.

jp1ac4 (approved):
tACK a6a1d85778e8aa996d074ebfd424840994f436d1.

PR #1837, opened 2025-09-03
use global tick for all fiat price requests

jp1ac4:
This follows on from #1835 to use the new global tick for handling all fiat price requests.

This allows us to combine and simplify the logic that controls how often regular updates are performed with that used to prevent multiple requests being sent, which is done by using `last_request` in both cases.

Explicit requests when starting the application and editing fiat price settings are no longer needed.

### DIFF
--- a/liana-gui/src/app/mod.rs
+++ b/liana-gui/src/app/mod.rs
@@ -35,7 +35,7 @@ use wallet::{sync_status, SyncStatus};
 
 use crate::{
     app::{
-        cache::{Cache, DaemonCache, FIAT_PRICE_UPDATE_INTERVAL_SECS},
+        cache::{Cache, DaemonCache, FIAT_PRICE_TTL},
         error::Error,
         menu::Menu,
         message::FiatMessage,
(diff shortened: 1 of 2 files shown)


jp1ac4:
Converted to draft to make a couple of small changes.

liana-gui/src/app/mod.rs
```diff
@@ -362,92 +327,85 @@ impl App {
...
+            {
+                let new_request = cache::FiatPriceRequest {
+                    source: sett.source,
+                    currency: sett.currency,
+                    instant: tick,
+                };
+                // Store the request immediately to prevent multiple requests being sent.
+                self.cache.fiat_price_cache.last_request = Some(new_request);
+                tasks.push(Task::perform(async move {}, |_| {
+                    Message::Fiat(FiatMessage::GetPrice)
```
edouardparis: I think we can directly do the task here no ? And remove FiatMessage::GetPrice

jp1ac4: This has now been done.

edouardparis (approved):
ACK 683659354084bb19514ea3f38da4c34f4f430d73

PR #1838, opened 2025-09-04
installer: grey out key if already used in this path

### LINKED ISSUE #1826: Missing error when selecting a device already in use in the same path
If I click on a device already used in the same path, the modal just closes without doing anything. I think we should have an error as we had with the old modal.

https://github.com/user-attachments/assets/14df7c62-fd58-44eb-9948-55187462794f

pythcoiner:
image

closes #1826


jp1ac4 (approved):
tACK cf9112c504c97f96fcd788fbbeb44893b9f45dbc.

PR #1839, opened 2025-09-04
store fiat prices in new global cache

jp1ac4:
This PR addresses https://github.com/wizardsardine/liana/issues/1819 by adding a new global cache to store fiat prices.

The `GUI` rather than individual tabs will now take care of triggering new price requests and processing results .


edouardparis (approved):
ACK f6acbdaaf8ae81eddbbee7392c9a9ead10773260

PR #1840, opened 2025-09-04
store currencies list in global cache

jp1ac4:
Instead of caching the available currencies in the settings page, this PR moves these lists to the global cache, which will avoid needing to request the currencies each time the user opens the settings page and will be especially useful in case the user is configuring multiple wallet tabs at the same time.


jp1ac4:
I don't think the failing test is related to this PR.

PR #1841, opened 2025-09-05
liana-sdk: deprecate into_multipath_descriptor() 

pythcoiner:
I think we should rename this method to make it more clear that we recompile.

### DIFF
--- a/liana/src/descriptors/analysis.rs
+++ b/liana/src/descriptors/analysis.rs
@@ -543,7 +543,7 @@ impl LianaPolicy {
             is_taproot,
         };
         if compile {
-            policy.clone().into_multipath_descriptor_fallible()?;
+            policy.clone().compile_multipath_descriptor_fallible()?;
         }
         Ok(policy)
     }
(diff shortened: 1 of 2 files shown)


liana/src/descriptors/analysis.rs
```diff
@@ -783,12 +783,20 @@ impl LianaPolicy {
     /// involves a Miniscript policy compilation: this function is **not deterministic**. If you
     /// are inferring a `LianaPolicy` from a descriptor, generating a descriptor from this
     /// `LianaPolicy` may not yield the same descriptor.
-    pub fn into_multipath_descriptor(
+    pub fn compile_multipath_descriptor(
         self,
     ) -> descriptor::Descriptor<descriptor::DescriptorPublicKey> {
         self.into_multipath_descriptor_fallible()
```
jp1ac4: Even though it's private, I think it would be good to also rename `into_multipath_descriptor_fallible`.

pythcoiner: yep!

pythcoiner: done, I've not keep the old fn w/ deprecation as it's private

jp1ac4 (approved):
ACK df6de383d6a67e7dc03f595975f11a377ea19fcb.

PR #1844, opened 2025-09-05
Improve key modal UI

### LINKED ISSUE #1825: Suggestion messages on devices in the set key modal are not aligned
See screenshot:

Image

I think it's related to the length of the device name

pythcoiner:
closes #1825

### DIFF
--- a/liana-gui/src/installer/step/descriptor/editor/key.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/key.rs
@@ -49,6 +49,7 @@ use crate::{
     signer::Signer,
 };
 
+const MAX_ALIAS_LEN: usize = 24;
 pub type FnMsg = fn() -> Message;
 
 pub fn new_multixkey_from_xpub(
@@ -766,9 +767,9 @@ impl SelectKeySource {
             );
             return Task::none();
         }
-        // TODO: which max length for an alias?
-        self.form_alias.valid = alias.len() < 30;
-        self.form_alias.value = alias;
+        if alias.chars().count() <= MAX_ALIAS_LEN {
+            self.form_alias.value = alias;
+        }
         Task::none()
     }
     fn on_account(&mut self, index: ChildNumber) -> Task<Message> {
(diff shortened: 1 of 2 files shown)


pythcoiner:
note: I tried to add `.wrapping(Wrapping::None)`, but it's just not effective:
image

pythcoiner:
> I think for the number of characters you would need `alias.chars().count()` as `len()` returns the number of bytes.

good catch (and weird rustism)

but I dont see our issue coming from there, as chars used in previous examples stay in "ASCII land" (1 byte), so in those example, `.len()` and `.count()` must output same values

fixed anyway

pythcoiner:
> That's not good. 13 chars seems too low of a limit. Maybe it's better to truncate with `(...)` at this point. Unless you have other solutions.

I dont think truncating is good either, for it to be effective we should truncate after the 11th char...

I think the better solution is to keep it as is, there is edge cases, but it's a iced limitation

liana-gui/src/installer/step/descriptor/editor/key.rs
```diff
@@ -766,9 +767,9 @@ impl SelectKeySource {
             );
             return Task::none();
         }
-        // TODO: which max length for an alias?
-        self.form_alias.valid = alias.len() < 30;
-        self.form_alias.value = alias;
+        if alias.chars().count() < MAX_ALIAS_LEN {
```
jp1ac4: It's perhaps a minor thing, but I would understand the max to be inclusive and so would expect `<=` .

pythcoiner: done: https://github.com/wizardsardine/liana/compare/d961e7fc2b0b5308f723d70aa4847e86af09ad04..cc0bb5ef470e23cfbf1196a8e578fee66d1cec8d

jp1ac4 (approved):
ACK cc0bb5ef470e23cfbf1196a8e578fee66d1cec8d.

PR #1845, opened 2025-09-05
Automatic and user triggered refresh of home information

edouardparis:
- An automatic refresh of the home balance is done every 10s unless the user is currently exploring the history by loading more pages. 
- When the user scroll to the very top, the refresh is automatically triggered
if the information was not loaded in the last 3s.

This PR is a less invasive change, the refresh occurs only when it matters for the user: when he is looking at the balance and the last payments.


jp1ac4 (approved):
tACK 90d398a033fa1b2f8e59694b73d3e3c8751cd690.

PR #1807, opened 2025-08-08
[WIP] Encrypted backup

pythcoiner:
- [x] Installer: import encrypted descriptor
- [x] Installer: export encrypted descriptor
- [x] Installer: import encrypted backup
- [x] Settings: export encrypted descriptor
- [x] Settings: export encrypted backup
- [x] Settings: import encrypted backup
- [x] Connect: import encrypted descriptor/backup

Closes #1803 & #1804

image

image

image

image

decryption can be tested by using the cli of [encrypted_descriptor](https://github.com/pythcoiner/encrypted_backup/tree/liana) on `liana` branch.
you can build the cli with

```
cargo build --release --features "cli miniscript_latest"
```

PR #1846, opened 2025-09-08
Encrypted descriptor

pythcoiner:
This is a lightweight version of #1807, where we do not encrypt for a backup export.


liana-gui/src/backup.rs
```diff
@@ -482,6 +451,28 @@ pub enum KeyType {
     ThirdParty,
 }
 
+impl ToPayload for Backup {
```
jp1ac4: I think this commit (297328797038d9770b064e47296482292634e8e0) can be omitted from this PR?

liana-gui/src/decrypt.rs
```diff
@@ -0,0 +1,655 @@
...
+                Ok(b) => Some(b),
+                Err(_) => {
+                    error = Some(Error::InvalidEncoding);
+                    None
+                }
+            } {
+            backup.get_derivation_paths().into_iter().collect()
+        } else {
+            let mut h = HashSet::new();
+            h.insert(DerivationPath::from_str("48'/0'/0'/2'").expect("hardcoded"));
```
jp1ac4: nit: Could you use the `default_derivation_path()` function here?

pythcoiner:
> I inserted it only in the "Wallet" section of the settings to distinguish between the concept of backup and import/exports material. So my idea is that it shouldn't be added here, but open to discuss. At the same time we have the plain text descriptor file among the exports, which seems to contradict this concept, but I don't think it should be in the wallet section since we should incentivize to use the encrypted version.

the plaintext descriptor option must be kept for register descriptor on airgap devices

liana-gui/src/app/view/settings/mod.rs
```diff
@@ -1039,6 +1027,12 @@ pub fn wallet_settings<'a>(
                 Row::new()
                     .spacing(10)
                     .push(Column::new().width(Length::Fill))
+                    .push(
+                        button::secondary(Some(icon::backup_icon()), "Back Up Descriptor")
```
jp1ac4: I would use "Back up descriptor" for case consistency with the other button.

liana-gui/src/export.rs
```diff
@@ -154,6 +160,8 @@ impl Display for Error {
                 f,
                 "Import failed. The PSBT either doesn't belong to the wallet or has already been spent."
             ),
+            Error::EncryptedBackup(e) => write!(f, "Fail to encrypt backup: {e:?}"),
+            Error::UnknownFormat => write!(f, "Format of the file unknow"),
```
jp1ac4: typo: "unknown"

liana-gui/src/export.rs
```diff
@@ -154,6 +160,8 @@ impl Display for Error {
                 f,
                 "Import failed. The PSBT either doesn't belong to the wallet or has already been spent."
             ),
+            Error::EncryptedBackup(e) => write!(f, "Fail to encrypt backup: {e:?}"),
```
jp1ac4: typo: "Failed"

liana-gui/src/decrypt.rs
```diff
@@ -0,0 +1,655 @@
...
+            import_xpub_error: None,
+            xpub: Value::default(),
+            xpub_busy: false,
+            mnemonic: Value::default(),
+            mnemonic_busy: false,
+            focus: Focus::None,
+            modal: None,
+        }
+    }
+    pub fn update(&mut self, msg: Decrypt) -> Task<installer::Message> {
```
jp1ac4: Would it be useful to return something more general that doesn't depend on `installer` here?

liana-gui/src/installer/prompt.rs
```diff
@@ -1,5 +1,12 @@
...
-pub const BACKUP_DESCRIPTOR_HELP: &str = "In Bitcoin, the coins are locked using a Script (related to the 'address'). In order to recover your funds you need to know both the Scripts you have participated in (your 'addresses'), and be able to sign a transaction that spends from those. For the ability to sign, you back up your private key, this is your mnemonic ('seed words'). For finding the coins that belong to you, you back up a template of your Script (your 'addresses'), this is your descriptor, included in your wallet backup file. However, note the descriptor need not be stored as securely as the private key. A thief who steals your descriptor but not your private key cannot steal your funds.";
+pub const BACKUP_DESCRIPTOR_MESSAGE: &str = "This backup is required to recover your funds.
+Click “Back Up Descriptor” to download an encrypted file of your wallet configuration and store it in safe, accessible places.
```
jp1ac4, line 2: Should this be changed to "encrypted file of your wallet descriptor"?

liana-gui/src/decrypt.rs
pythcoiner: for now only uses cases are with installer::Message, if needed we can later make it more generic

liana-gui/src/backup.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/7cc67366e3324b5e09ed0e126f28de1bc249d09d..182d6e1ce7f1032a7dbb970dd40a5d6e5fc1271b

liana-gui/src/decrypt.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/182d6e1ce7f1032a7dbb970dd40a5d6e5fc1271b..fa4f4cdc5d20ccb0734616275ec09e3a831df2f7

liana-gui/src/app/view/settings/mod.rs
pythcoiner: done:
https://github.com/wizardsardine/liana/compare/fa4f4cdc5d20ccb0734616275ec09e3a831df2f7..95726f2aec86a0d3e6b2c1d60aaa9ef2c99cd3e3

liana-gui/src/export.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/95726f2aec86a0d3e6b2c1d60aaa9ef2c99cd3e3..481e9627b3a92b44fe9bfc304a9499b9008d4aa4

liana-gui/src/export.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/481e9627b3a92b44fe9bfc304a9499b9008d4aa4..6c498101729b4e38691fc3eeb79f74a916c45b2b

liana-gui/src/installer/prompt.rs
pythcoiner: @nondiremanuel

liana/src/descriptors/mod.rs
```diff
@@ -266,6 +271,42 @@ impl LianaDescriptor {
             .expect("We never create a Liana descriptor with an invalid Liana policy.")
     }
 
+    /// Get the set of Xpubs that can sign (excluding unspendable key) in the form of an
+    /// DescriptorPublicKey::XPub with empty derivation path, in order to avoid duplicates.
+    pub fn keys(&self) -> Vec<DescriptorPublicKey> {
```
jp1ac4: I feel the method name is too general for what it does.

liana/src/descriptors/mod.rs
```diff
@@ -266,6 +271,42 @@ impl LianaDescriptor {
...
+        let mut keys = BTreeSet::new();
+        let nums = bip341_nums();
+        self.multi_desc.for_each_key(|k| {
+            if let DescriptorPublicKey::MultiXPub(multixkey) = k {
+                let key = DescriptorPublicKey::XPub(DescriptorXKey {
+                    origin: multixkey.origin.clone(),
+                    xkey: multixkey.xkey,
```
jp1ac4, line 283: Rather than adding dummy values, which could misleadingly suggest these are the values present in the descriptor itself, wouldn't it be clearer to return the `multikey.xkey` and then let the caller add these dummy values as required?

liana/src/descriptors/mod.rs
pythcoiner: I'm happy if you gimme a better one

liana/src/descriptors/mod.rs
pythcoiner: those are not dummies values, It's just about having a `DescriptorPublicKey` that contains Xpub + Origin, Xpub is not enough as it does not contains the origin and I expect we use the result of this function to list xpubs of the Bip388 Policy

this way I think we can directly call .to_string() 

- `DescriptorPublicKey::Single` is not suitable as it will not output a xpub
- `DescriptorPublicKey::MultiXpub` is not suitable as `DerivPaths` cannot be empty

liana-gui/src/installer/mod.rs
```diff
@@ -242,13 +241,31 @@ impl Installer {
...
+                        let mut task_2 = Task::none();
+                        if update {
+                            task_2 = self
+                                .steps
+                                .get_mut(self.current)
+                                .expect("There is always a step")
+                                .update(
+                                    &mut self.hws,
+                                    // We notify downstream that the the list have been updated
+                                    Message::HardwareWallets(HardwareWalletMessage::Update),
```
edouardparis: it is a nit:
I think HardwareWalletMessage should be reserved for internal use of the hws module.
I think a new Message::HardwareWalletsUpdated should be added to the installer Message instead as it is trigger by the installer.update() for its internal steps.

liana/src/descriptors/mod.rs
edouardparis: I am fine with it, maybe add a comment to say that is a vessel for the xkey.

pythcoiner: well it's not a vessel, it's litteraly how it's intended to be used:

this

```
let key_str = "[abcdef01/0/0/0]xpub6Eze7yAT3Y1wGrnzedCNVYDXUqa9NmHVWck5emBaTbXtURbe1NWZbK9bsz1TiVE7Cz341PMTfYgFw1KdLWdzcM1UMFTcdQfCYhhXZ2HJvTW";
let key = DescriptorPublicKey::from_str(key_str).unwrap();

```

will yield

```
XPub(
    DescriptorXKey {
        origin: Some(
            (
                0xabcdef01,
                0/0/0,
            ),
        ),
        xkey: Xpub {
            network: Main,
            depth: 4,
            parent_fingerprint: 0xc6c45ae6,
            child_number: Normal {
                index: 0,
            },
            public_key: PublicKey(
                4fcda97320d37e42ddd2c9adf5c5c5edf7ff9decb550c50ad2ace5ae3a16abb3b780ec63e8766dbaa3e323148cc0e6ee760e3a41b5a0b1229f5d69b59ce3a0a3,
            ),
            chain_code: 0xf98f1f14305e50df1ff23883210e9020e212bc62a790eff034e9f5036af844a1,
        },
        derivation_path: ,
        wildcard: None,
    },
)

```

I'm fine to add a comment, but from my POV make it more confusing

jp1ac4: Sorry I understand better now. I think a name like `definite_keys_without_deriv_path()` would make it clearer what is being returned (rust-miniscript uses `DefiniteDescriptorKey` for a DescriptorPublicKey without wildcard).

jp1ac4: Also, since we require an origin for all user-provided keys when creating a `LianaDescriptor` (as enforced in the constructors), an alternative check would be to exclude any descriptor keys that don't have an origin. Or simply to return all keys and let the caller check if it has an origin.

pythcoiner: > an alternative check would be to exclude any descriptor keys that don't have an origin

in which other case than the unspendable key we can have a key w/o origin?

pythcoiner: > (rust-miniscript uses DefiniteDescriptorKey for a DescriptorPublicKey without wildcard)

No in our case it's not a `DefiniteDescriptorKey`, my understanding of "definite" in rust-miniscript context is that the key is not deriveable anymore, while the key we generating here is the "root" key for derivation

liana-gui/src/installer/mod.rs
pythcoiner: done https://github.com/wizardsardine/liana/compare/da49af8b909ff17a1648e1e6a3ace85969734d58..3f11fe3c49fa9ef88a219cf43a2f0979061dd22c

liana/src/descriptors/mod.rs
pythcoiner: more I thought about it more I feel `.keys()` is the right wording: the result is just the keys from descriptor context:
- different multipath we use are a "hack" in order to derive a different signing key in each branch, "real" keys are represented by `DescriptorPublicKey::XPub`
- unspendable key is provably not a key

I think everything else than `.keys()` or `.spendable_keys()` will add more confusions

liana-gui/src/export.rs
```diff
@@ -603,6 +605,44 @@ pub async fn export_encrypted_descriptor(
 ) -> Result<(), Error> {
     let descriptor = descr.descriptor();
     let bytes = EncryptedBackup::new().set_payload(descriptor)?.encrypt()?;
+
```
edouardparis, line 608: I think you can add some `   send_progress!(sender, Progress(30.0));` across the method.
It will allows us to detect when the error EncryptionFailed is returned

liana/src/descriptors/mod.rs
jp1ac4: That's true. I feel `spendable_keys()` is slightly better as it makes clear that the unspendable key is not returned, but I think either way is fine.

liana/src/descriptors/mod.rs
jp1ac4: > in which other case than the unspendable key we can have a key w/o origin?

I meant it would be an equivalent check, since we expect a key not to have an origin if and only if it's unspendable, but I suppose it's more reliable to check for the NUMS directly.

liana-gui/src/export.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/6f797947507409f6bea177ba41fc1a895ab4a7a9..432d58ee4843b26f99e532af8ef85c5fda34bcd4

liana/src/descriptors/mod.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/432d58ee4843b26f99e532af8ef85c5fda34bcd4..e6d4a98acb6d57b6d149ebdcdb5dadc520980fe2

liana/src/descriptors/mod.rs
pythcoiner: > but I suppose it's more reliable to check for the NUMS directly.

yes especially if in the future we allow importing key w/o origin (watchonly?)

jp1ac4 (approved):
tACK b278ef298138de6e7d0bacb470ab3a9337863752.

Issue #1802, opened 2025-08-04
Fiat price indication in "Send" flow


nondiremanuel:
It would be very nice to have, for those who enabled the "Fiat price" setting, a way to see how much they are spending in fiat. It would be also good to have the possibility to set the fiat price in the "Spend" process and the BTC spend amount would reflect the conversion of that. 
An indication of the amount of the fee in fiat would be also good to have.

A high-level mockup of the possible flow is shown [here](https://miro.com/app/board/uXjVKvwKmIs=/?moveToWidget=3458764636375766497&cot=14).

In the Recovery flow we could still show the fiat amount, but both would be non-editable. We can show it in the same way with the boxes both disabled or in a different one according also to what's easier on the implementation POV.

PR #1848, opened 2025-09-09
feat: allow to view & edit fiat amounts on Send page

jp1ac4:
This PR implements #1802.

When the fiat price setting is enabled, the user will be able to do the following on the Send page:
- enter a fiat amount directly for a recipient, which will then be converted to BTC
- view the fiat value of an edited BTC amount
- view the exchange rate used
- view the fee's fiat value

Note that if the exchange rate changes in the background, the new exchange rate will be used:
- when converting any newly- or already-edited BTC amounts to fiat
- when converting any newly-edited fiat amounts to BTC
- for the displayed exchange rate
- for the fee's fiat value

Any BTC amounts that have already been converted from fiat values will not change upon an exchange rate update (and neither will the fiat amount that was entered by the user).


liana-gui/src/app/view/spend/mod.rs
```diff
@@ -187,6 +188,15 @@ pub fn create_spend_tx<'a>(
             .push(
                 Column::new()
                     .push(Column::with_children(recipients).spacing(10))
+                    .push_maybe(fiat_converter.map(|conv| {
```
jp1ac4: We should probably not show the exchange rate for a self-send as there are no recipient amounts to enter.

jp1ac4:
I've added a check for a change in either the fiat currency or the exchange rate, in both of which cases the fiat amount will be recalculated from the BTC amount, if any. In the latter case where the currency is the same but the exchange rate changes, I keep the fiat form value if it's an invalid amount so that the user can fix it (most likely the user is currently editing this field).

liana-gui/src/app/view/spend/mod.rs
jp1ac4: The exchange rate is now shown within the recipient card so will no longer be visible for a self-send.

jp1ac4:
As discussed, I've made the following changes:
- "Amount" form label changed to "Amount (BTC)"
- Show currency label before the text input field to avoid too much space between fiat amount and label
- Show exchange rate as a tooltip within "i" with additional details about the price source and update time

PR #1849, opened 2025-09-09
disable fiat prices in all cases where not explicitly enabled

jp1ac4:
It was decided for now not to enable fiat prices by default for Liana Connect mainnet users. Furthermore, discussions about a notification/modal upon opening the GUI are ongoing and so for now we can treat a missing setting the same way in all cases.


jp1ac4:
Rebased on master.

edouardparis (approved):
utACK a9d07cdfe3e92ccc297e82573e55906f57e2e626

PR #1852, opened 2025-09-11
Modify how fiat amounts are displayed

jp1ac4:
This PR makes small adjustments to how fiat amounts are displayed on Home and Send pages:
* smaller text size
* use "~" without parentheses
* more space between BTC and fiat amounts on home page

These changes are instead of those proposed in #1843.

### DIFF
--- a/liana-gui/src/app/view/home.rs
+++ b/liana-gui/src/app/view/home.rs
@@ -80,30 +80,33 @@ pub fn home_view<'a>(
         .push(h3("Balance"))
         .push(
             Column::new()
-                .push(if sync_status.is_synced() {
-                    Row::new()
-                        .align_y(Alignment::Center)
-                        .push(amount_with_size(balance, H1_SIZE))
-                        .push_maybe(fiat_balance.map(|fiat| {
-                            Row::new()
-                                .align_y(Alignment::Center)
-                                .push(Space::with_width(20))
-                                .push(fiat.to_text().size(H2_SIZE).color(color::GREY_2))
-                        }))
-                } else {
-                    Row::new().push(spinner::Carousel::new(
-                        Duration::from_millis(1000),
-                        vec![
-                            amount_with_size(balance, H1_SIZE),
-                            amount_with_size_and_colors(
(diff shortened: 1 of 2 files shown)


liana-gui/src/app/view/home.rs
```diff
@@ -82,20 +82,13 @@ pub fn home_view<'a>(
...
-                                    text(format!(
-                                        "({} {})",
-                                        fiat.to_formatted_string(),
-                                        fiat.currency()
-                                    ))
-                                    .size(H1_SIZE)
-                                    .color(color::GREY_2),
-                                )
+                                .push(Space::with_width(20))
+                                .push(fiat.to_text().size(H2_SIZE).color(color::GREY_2))
```
edouardparis: For responsive layout when the window size is reduced you should
add .wrap() and you will certainly have to add wrap() in the else clause

liana-gui/src/app/view/home.rs
```diff
@@ -150,16 +144,8 @@ pub fn home_view<'a>(
...
-                                            text(format!(
-                                                "({} {})",
-                                                fiat.to_formatted_string(),
-                                                fiat.currency()
-                                            ))
-                                            .size(H3_SIZE)
-                                            .color(color::GREY_3),
-                                        )
+                                        .push(Space::with_width(10)) // total spacing = 20 including row spacing
+                                        .push(fiat.to_text().size(H4_SIZE).color(color::GREY_3))
```
edouardparis, line 151: Same here, I think I did not try with unconfirmed amounts

liana-gui/src/app/view/home.rs
jp1ac4: I've added a `.wrap()` on the entire `Row` that covers both `if` and `else` branches.

liana-gui/src/app/view/home.rs
jp1ac4: I've added a `.wrap` on the whole `Row` here, and also on the Send page for the `Row` containing the feerate and fee.

PR #1853, opened 2025-09-12
gui: fix badge's tooltip display

thomask7b:
Sometimes visibility of tooltip could be improved, what do you think?

Capture d’écran du 2025-09-05 10-12-23

With borders and background :

Capture d’écran du 2025-09-05 10-02-55

### DIFF
--- a/liana-ui/src/component/badge.rs
+++ b/liana-ui/src/component/badge.rs
@@ -78,7 +78,9 @@ pub fn badge_pill<'a, T: 'a>(label: &'a str, tooltip: &'a str) -> Container<'a, 
                 .padding(10)
                 .center_x(Length::Shrink)
                 .style(theme::pill::simple),
-            tooltip,
+            Container::new(text::p1_regular(tooltip))
+                .padding(10)
+                .style(theme::card::simple),
             tooltip::Position::Top,
         )
     })


nondiremanuel:
Thanks for taking this. 
I'm not entirely convinced by all the different colors present in the screen, but I'm not sure how it should look like. I'll let others comment too.

liana-ui/src/component/badge.rs
```diff
@@ -78,7 +78,9 @@ pub fn badge_pill<'a, T: 'a>(label: &'a str, tooltip: &'a str) -> Container<'a,
                 .padding(10)
                 .center_x(Length::Shrink)
                 .style(theme::pill::simple),
-            tooltip,
+            Container::new(text::p2_regular(tooltip))
+                .padding(8)
+                .style(theme::card::warning),
```
jp1ac4: I think `theme::card::simple` would be more suitable here to avoid the warning colours, and this would also be consistent with some other tooltips. You could also use `padding(10)` here for consistency with other UI elements, unless there was a particular reason for using `8`?

thomasballivet: Just to not take too much space but I changed to 10 and simple

liana-ui/src/component/badge.rs
```diff
@@ -78,7 +78,9 @@ pub fn badge_pill<'a, T: 'a>(label: &'a str, tooltip: &'a str) -> Container<'a,
                 .padding(10)
                 .center_x(Length::Shrink)
                 .style(theme::pill::simple),
-            tooltip,
+            Container::new(text::p2_regular(tooltip))
```
jp1ac4: I think this should be `p1_regular` to maintain the current tooltip font size.

PR #1858, opened 2025-09-17
installer: do not allow duplicate key aliases

### LINKED ISSUE #1850: Two keys shouldn't be allowed to have the same alias
Likely introduced with the new set key modal work:

Image

While previously we had:

Image

We should restore the error behavior.

pythcoiner:
This PR add a check to detect if an alias is already in use for another key in both:
 - `SelectKeySource`:  
 
image

 - `EditKeyAlias`:
 
image

 
closes #1850


jp1ac4 (approved):
tACK ba103e425aa8337db9892c183dc746215a40419c.

PR #1872, opened 2025-09-23
fix: map insufficient funds error to success result for Liana Connect backend

### LINKED ISSUE #1871: Liana Connect: Handling insufficient funds and other errors during PSBT generation
This relates to all types of PSBT generation (spend + recovery + RBF) when using Liana Connect as a backend.

Due to improper error handling leading to unclear error messages (e.g. https://github.com/wizardsardine/liana/issues/1677#issue-3025021993), we added checks for HTTP success codes for the above API endpoints (see #1701). While this resolved the unclear error message, it inadvertently meant that an insufficient funds error was displayed just as any other error rather than being used to display the remaining amount to select as part of the spend creation flow.

So we need to keep the improved error handling while treating insufficient funds as a special case.

There seem to be two options to resolve this:
1) keep `check_success()` in the GUI to treat anything without success code as an error; change backend to return `StatusCode:OK` when there are insufficient funds; optionally remove unused `DraftPsbtResult::Error` variant from GUI.
2) remove `check_success()` from the GUI; add `DraftPsbtResult::Error` variant to the backend to be consistent with the GUI and use it to return all errors other than insufficient funds.

jp1ac4:
This fixes #1871 by correctly mapping the insufficient funds error to a success result and otherwise returning details of the unsuccessful response.

### DIFF
--- a/liana-gui/src/services/connect/client/backend/mod.rs
+++ b/liana-gui/src/services/connect/client/backend/mod.rs
@@ -797,9 +797,7 @@ impl Daemon for BackendWalletClient {
             })
             .send()
             .await?
-            .check_success()
-            .await?
-            .json()
+            .json() // no need to check success before parsing the json as it could be an error variant
             .await?;
 
         match res {
@@ -840,9 +838,7 @@ impl Daemon for BackendWalletClient {
             })
             .send()
             .await?
-            .check_success()
-            .await?
-            .json()
+            .json() // no need to check success before parsing the json as it could be an error variant
             .await?;
 
         match res {


liana-gui/src/services/connect/client/backend/api.rs
```diff
@@ -316,19 +316,13 @@ pub struct Psbt {
...
-    Error(DraftPsbtError),
 }
 
 #[derive(Clone, Deserialize)]
 pub struct InsufficientFundsInfo {
     pub missing: u64,
 }
 
-#[derive(Clone, Deserialize)]
-pub struct DraftPsbtError {
```
edouardparis, line 328: I thinking not using it and directly return the text body of the response will make the error uglier in the error banner (with json brackets).

jp1ac4: Ah yes, I was mistaken that this variant is not used. I'll drop this second commit as in any case it wasn't necessary to fix the insufficient funds bug.

PR #1900, opened 2025-10-16
Dust limit

### LINKED ISSUE #1865: Lower Liana "dust" limit
Our current "dust" limit for CREATING outputs is 5k sats. It does not reflect reality of the network.

I understand our scripts might require higher amounts than the usual dust limit in Core, but either we should calculate that on a per-user basis, or pick a more sane number than 5k sats.

It would be very useful for users to be able to modify this setting.

### LINKED ISSUE #1910: Throw "Invalid amount" error when "MAX" of selected coins is dust
When selecting "MAX" but the amount of the coin(s) selected is lower than dust, we currently require the user to add extra sats to reach the minimum to not be dust. Here an example with `dust limit` = 5000

Image

We should instead show the correct amount as we do for all the other use cases. In the screenshot the "Amount (BTC)" field should be filled with 2000 - `fee` = 0.00001901 BTC in this case. 
If, as in the example, the amount of the coins selected is lower than `dust limit` we should throw the same error that we show when the user manually inputs a dust amount:

Image

pythcoiner:
closes #1865
closes #1910

### DIFF
--- a/tests/test_rpc.py
+++ b/tests/test_rpc.py
@@ -1017,7 +1017,7 @@ def test_start_rescan(lianad, bitcoind):
     # First, get some coins
     for _ in range(10):
         addr = lianad.rpc.getnewaddress()["address"]
-        amount = random.randint(1, COIN * 10) / COIN
+        amount = random.randint(500, COIN * 10) / COIN
         txid = bitcoind.rpc.sendtoaddress(addr, amount)
         bitcoind.generate_block(random.randint(1, 10), wait_for_mempool=txid)
     wait_for(lambda: len(list_coins()) == 10)
@@ -1027,7 +1027,7 @@ def test_start_rescan(lianad, bitcoind):
     # without change, single or multiple inputs, sending externally or to self).
     for _ in range(5):
         addr = lianad.rpc.getnewaddress()["address"]
-        amount = random.randint(1, COIN * 10) / COIN
+        amount = random.randint(500, COIN * 10) / COIN
         txid = bitcoind.rpc.sendtoaddress(addr, amount)
         avail = list(unspent_coins())
         to_spend = random.sample(avail, random.randint(1, len(avail)))


liana-gui/src/app/mod.rs
```diff
@@ -374,6 +374,12 @@ impl App {
     }
 
     pub fn update(&mut self, message: Message) -> Task<Message> {
+        if !matches!(
+            message,
+            Message::Tick | Message::UpdateDaemonCache(_) | Message::CacheUpdated
+        ) {
+            println!("{message:?}");
+        }
```
jp1ac4: should this be part of the PR?

liana-gui/test_assets/global_settings.json
```diff
@@ -77,4 +77,4 @@
     "width": 1248.0,
     "height": 688.0
   }
-}
+}
```
jp1ac4: I don't think this change is required.

liana/src/spend.rs
```diff
@@ -55,6 +58,8 @@ pub enum SpendCreationError {
     SanityCheckFailure(Psbt),
     FetchingTransaction(bitcoin::OutPoint),
     CoinSelection(InsufficientFunds),
+    MaxUnderDustSend,
+    MaxUnderDustSelect,
```
jp1ac4: I don't think we should add these error variants if they're not used in the spend module itself.

liana-gui/src/app/mod.rs
pythcoiner: done: https://github.com/wizardsardine/liana/compare/c2d3ff13a53912de6258792518f5ae4fa7240166..a8822cd90a68d99c17455331e2fb1640e5babd7e

liana-gui/test_assets/global_settings.json
pythcoiner: yeah, but I dont know how this came up, it's not even a change & tests are still passing

liana/src/spend.rs
pythcoiner: I thinks it's fine for now, changing this imply changing the way the error hare displayed in the spend page, and this is likely to happen in a follow up, because still under discussion w/ no clear specs.

liana-gui/src/services/connect/client/auth.rs
```diff
@@ -161,6 +163,7 @@ impl AuthClient {
             ))
             .header("apikey", &self.api_public_key)
             .header("Content-Type", "application/json")
+            .header("User-Agent", format!("liana-gui/{}", crate::VERSION))
```
edouardparis, line 166: Please rebase, because it is weird to have change in the PRs

edouardparis (approved):
ACK 1a888fd3155185f399fdd050992ce52fec4a627a

PR #1901, opened 2025-10-16
Remove label for non owned outputs

### LINKED ISSUE #961: Do not label coins we don't own
Gransaction Screen:
image

the second coin is not mine, but u can add a label, it's a bit confusing.

sdmg15:
This is an attempt to remove the labels for non-owned outputs of the wallet.
If there's a better approach I'm happy to adjust.

Closes #961 

image
image

### DIFF
--- a/liana-gui/src/app/view/label.rs
+++ b/liana-gui/src/app/view/label.rs
@@ -1,6 +1,7 @@
 use iced::{advanced::text::Shaping, widget::row, Alignment};
 
 use liana_ui::{
+    color,
     component::{button, form},
     icon,
     widget::*,
@@ -64,5 +65,27 @@ pub fn label_editing(
         .align_y(Alignment::Center),
     )
     .into();
+    e.map(move |msg| view::Message::Label(labelled.clone(), msg))
+}
+
+pub fn label_non_editable(
+    labelled: Vec<String>,
+    label: Option<&String>,
+    size: u16,
+) -> Element<view::Message> {
+    let label_text = label.map(|s| s.as_str()).unwrap_or("(External Ouput)");
+
+    let e: Element<view::LabelMessage> = Container::new(
+        row![Container::new(
+            Text::new(label_text)
+                .size(size)
+                .width(iced::Length::Fill)
+                .color(color::GREY_1)
+        ),]
+        .spacing(5)
+        .align_y(Alignment::Center),
+    )
+    .into();
+
     e.map(move |msg| view::Message::Label(labelled.clone(), msg))
 }
--- a/liana-gui/src/app/view/psbt.rs
+++ b/liana-gui/src/app/view/psbt.rs
@@ -91,10 +91,11 @@ pub fn psbt_view<'a>(
                     .push(outputs_view(
                         &tx.psbt.unsigned_tx,
                         network,
-                        Some(tx.change_indexes.clone()),
+                        &tx.change_indexes,
                         &tx.labels,
                         labels_editing,
                         tx.is_single_payment().is_some(),
+                        false,
                     )),
             )
             .push(if saved {
@@ -693,12 +694,12 @@ pub fn inputs_view<'a>(
 pub fn outputs_view<'a>(
     tx: &'a Transaction,
     network: Network,
-    change_indexes: Option<Vec<usize>>,
+    change_indexes: &'a [usize],
     labels: &'a HashMap<String, String>,
     labels_editing: &'a HashMap<String, form::Value<String>>,
     is_single_payment: bool,
+    is_external: bool,
 ) -> Element<'a, Message> {
-    let change_indexes_copy = change_indexes.clone();
     Column::new()
         .spacing(20)
         .push({
@@ -707,11 +708,10 @@ pub fn outputs_view<'a>(
                 .iter()
                 .enumerate()
                 .filter(|(i, _)| {
-                    if let Some(indexes) = change_indexes_copy.as_ref() {
-                        !indexes.contains(i)
-                    } else {
-                        true
+                    if is_external {
+                        return true;
                     }
+                    !change_indexes.contains(i)
                 })
                 .count();
             if count > 0 {
@@ -757,15 +757,20 @@ pub fn outputs_view<'a>(
                             .iter()
                             .enumerate()
                             .filter(|(i, _)| {
-                                if let Some(indexes) = change_indexes_copy.as_ref() {
-                                    !indexes.contains(i)
-                                } else {
-                                    true
+                                if is_external {
+                                    return true;
                                 }
+                                !change_indexes.contains(i)
                             })
                             .fold(
                                 Column::new().padding(20),
                                 |col: Column<'a, Message>, (i, output)| {
+                                    let mut is_editable = true;
+
+                                    if is_external && !change_indexes.contains(&i) {
+                                        is_editable = false;
+                                    }
+
                                     col.spacing(10).push(payment_view(
                                         i,
                                         tx.compute_txid(),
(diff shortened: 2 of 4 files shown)


liana-gui/src/app/view/label.rs
```diff
@@ -66,3 +67,25 @@ pub fn label_editing(
     .into();
     e.map(move |msg| view::Message::Label(labelled.clone(), msg))
 }
+
+pub fn label_non_editable(
```
edouardparis, line 71: From the issue, we actually do not want to display a label at all, as the user cannot edit it.
Maybe we can just add "External" to be explicit for the user that the payment/output is not for him.

liana-gui/src/app/view/psbt.rs
```diff
@@ -764,8 +766,9 @@ pub fn outputs_view<'a>(
                                 }
                             })
                             .fold(
-                                Column::new().padding(20),
+                                Column::new().padding(20),                                                                                                                                                                                                                                                   
                                 |col: Column<'a, Message>, (i, output)| {
+                                    let is_editable = !(change_indexes_copy.is_none() && is_single_payment && i == 0 && has_outgoing.unwrap());
```
edouardparis: we want to display "External" and remove any label edit form for payments/outputs that have these conditions: it is from an incoming transaction AND the change_indexes does not contain its index.

edouardparis (commented):
Please do not forget to use `cargo clippy` to catch linter errors, thank you.

liana-gui/src/app/view/label.rs
sdmg15: Alright, any idea on how it should look like in the design? 
Could still use label style written "External" on it but not clickable what do you think?

liana-gui/src/app/view/psbt.rs
sdmg15: Yes but inside `output_view` the `change_indexes` is been cleared when it's incoming aka (`tx.is_external()`) so that's why I was puzzled here.

liana-gui/src/app/view/label.rs
pythcoiner: lets wait feedback from @nondiremanuel

nondiremanuel: I personally don't see the point of showing the details of external outputs in this screen at all. We already have the Txid in case someone is curious about it, but we are not a block explorer. I would personally remove not-owned outputs entirely, which would make UI easier too, unless either there is some reason I'm missing or it's very complicated in terms of logic.

pythcoiner: I think it makes sense to show the details, if ones is not interested to see the details he can just stay on the `Home` page

nondiremanuel: I don't disagree with having some details, but what is the usefulness of this specific detail (e.g. knowing which other UTXOs where part of an incoming transaction)? In the context of a software wallet, it seems just some potentially confusing data to me, but since it's there it is likely I'm missing something. The useful details in this page to me seem to be: Date, TxId, Receiving address of owned UTXOs, their labels and amounts.

pythcoiner: I'd say mainly for technical purpose

nondiremanuel: what technical purpose? Sorry I need to understand well what's the purpose of this data.

liana-gui/src/app/view/psbt.rs
sdmg15: @edouardparis waiting for you inputs here please.

edouardparis: It is not cleared, please look at the HistoryTransaction in the model.rs. You will see that the incoming_amount is calculated with the change_indexes https://github.com/wizardsardine/liana/blob/master/liana-gui/src/daemon/model.rs#L311
I know that the name change_indexes field is confusing, but in our case it is the array of output indexes that belong to the wallet, it is the key for you to say which incoming outputs belong to user and which one are external outputs.

sdmg15: Yes I already understood that, it's the way it's implemented :wink:

By cleared I meant this here https://github.com/wizardsardine/liana/blob/master/liana-gui/src/app/view/transactions.rs#L437. My work around is to add another parameter which will contain the original `change_indexes` so that the check could be done inside the `output_view` 


```rust
        .push(super::psbt::outputs_view(
            &tx.tx,
            cache.network,
            if tx.is_external() {
                None
            } else {
                Some(tx.change_indexes.clone())
            },
            &tx.labels,
            labels_editing,
            tx.is_single_payment().is_some(),

            Some(tx.change_indexes.clone()) // adding this parameter
        ))
```

edouardparis: Ok I understand, I think you can modify the whole psbt::outputs_view signature then for 
having the is_external passed as bool (always false for psbts) and the change_indexes without any Option.

pythcoiner:
image

I dont think it's the expected behavior:
- 0.00000499 coin is a coin that I received from an external wallet (I expect to label it)
- 0.07323419 coin is a change from external to external (I do not expect to label it)

liana-gui/src/app/view/psbt.rs
```diff
@@ -766,9 +766,14 @@ pub fn outputs_view<'a>(
                                 }
                             })
                             .fold(
-                                Column::new().padding(20),                                                                                                                                                                                                                                                   
+                                Column::new().padding(20),
                                 |col: Column<'a, Message>, (i, output)| {
-                                    let is_editable = !(change_indexes_copy.is_none() && is_single_payment && i == 0 && has_outgoing.unwrap());
```
pythcoiner: Here you overwrite what have been added in the previous commit, maybe there is no need for 2 commits and you can squash them?

liana-gui/src/app/view/psbt.rs
```diff
@@ -766,6 +768,12 @@ pub fn outputs_view<'a>(
                             .fold(
                                 Column::new().padding(20),
                                 |col: Column<'a, Message>, (i, output)| {
+                                    let mut is_editable = false;
+
+                                    if let Some(indexes) = &initial_change_indexes {
+                                        is_editable = indexes.contains(&i);
```
pythcoiner: hum, seems to me we filter-out owned indexes in the previous .filter() call? so you can never land here?

sdmg15: Yes you're right it's filtered that's why I created a different parameter, but suggestion from Edouard needs to be applied

liana-gui/src/app/view/psbt.rs
sdmg15: Yes indeed will squash them

liana-gui/src/app/view/psbt.rs
```diff
@@ -757,15 +757,16 @@ pub fn outputs_view<'a>(
...
-                                    true
+                                if is_external {
+                                    return true;
                                 }
+                                !change_indexes.contains(i)
                             })
                             .fold(
                                 Column::new().padding(20),
                                 |col: Column<'a, Message>, (i, output)| {
+                                    let is_editable = change_indexes.contains(&i);
```
edouardparis: I tested for a sent transaction, and the payment is marked as not editable. Something is wrong.
if transaction is external, is_editable should be true, if transaction is not external, the editable should be true in all cases.

sdmg15: Please can you provide me with screenshot of which part you're talking about?
image

edouardparis: Your example show what is wrong:
it is a outgoing payment where I cannot label the payment.
We want the `(external output)` only for output that are not ours in incoming payment (external tx).

sdmg15: Yes indeed you're right. I just fixed it. Thanks for taking your time to take a look a this.

image

PR #1908, opened 2025-10-23
installer: update hws aliases from key editing modal

### LINKED ISSUE #1905: Hardware aliases not shown when registering descriptor
The hardware alias indication should be shown somewhere in the descriptor registration modal. Probably something like "`HW_Alias` (#`fingerprint`)" instead of just the fingerprint should be enough

Image

pythcoiner:
closes #1905


edouardparis (approved):
utACK b292a9aac359f7e93d1110c91a57a3eaf6ac5c0d

PR #1913, opened 2025-10-30
Add login redirection on auth errrors for liana-connect wallets

edouardparis:
- Authorization error may be caused by session deletion. The error should simply be logged and user redirected in this case.


edouardparis (commented):
Self-ACK 6f6fe04d5ce1cab85bfde41d91a74fabf4276cfa

PR #1914, opened 2025-10-31
gui: bump async-hwi

pythcoiner:
This in order to fix a design flow w/ coldcard devices: previously it was not notified to the coldcard devices with which descriptor it needs sign, thus the device was picking the first descriptor that has a matching key, leading to signature fails if several descriptor using same key where registered on the device.

With Coldcard EDGE firmware from 6.4.0, we can notify the device of which descriptor use with this update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,9 +83,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


liana-ui/src/component/toast.rs
```diff
@@ -232,47 +232,6 @@ where
         .translate(Vector::new(self.position.x, self.position.y))
     }
 
-    fn on_event(
```
edouardparis, line 235: this commit does not bump anything, it only moves the method below. This commit should be maybe dropped

pythcoiner: done

edouardparis (approved):
ACK 2875cd0ee9876862a046227a105702ba3091fe93

PR #1918, opened 2025-11-06
docs: minor improvement for docs

huyejia:
minor improvement for docs

### DIFF
--- a/liana-gui/src/backup.rs
+++ b/liana-gui/src/backup.rs
@@ -434,7 +434,7 @@ pub enum KeyRole {
     Main,
     /// Key that will be used for recover in case loss of main key(s)
     Recovery,
-    /// Key that wil inherit coins if main user disapear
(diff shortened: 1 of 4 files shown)


liana-gui/src/backup.rs
```diff
@@ -434,7 +434,7 @@ pub enum KeyRole {
     Main,
     /// Key that will be used for recover in case loss of main key(s)
     Recovery,
-    /// Key that wil inherit coins if main user disapear
+    /// Key that will inherit coins if main user disappear
```
jp1ac4: Could you also please add "s", i.e. "disappears".

jp1ac4 (commented):
Thanks very much for this PR.

There's one small additional change that would be good to include.

liana-gui/src/backup.rs
huyejia: Of course. Added.

PR #1806, opened 2025-08-07
WIP - Payjoin POC - DO NOT MERGE

benalleng:
#### THIS PR IS NOT INTENDED TO BE MERGED

This commit acts as the initial approach set out by Artur, Armin, and myself to add payjoin support into liana.

Any changes from this PR will be split up into their individual issues / PRs for more precise review.

Some initial starting points may be to add bip21 support as an example which is a small lift and is ultimately not payjoin specific

Issue #1040, opened 2024-03-26
[GUI] Hint user thae he can verify address by QRCode for Specter


pythcoiner:
![image](https://github.com/wizardsardine/liana/assets/124568858/9624f349-4bfd-4810-b799-b60457a08086)

PR #1922, opened 2025-11-07
Add bip21 receive display support

benalleng:
This adds the scaffolding for the receiver to display bip21 QR codes when a payjoin or other bip21 receive format is required.

This is the first commit to begin support for #1806 by splitting it up into many smaller commits.  This particular commit does not require any payjoin support.

### DIFF
--- a/liana-gui/src/app/state/receive.rs
+++ b/liana-gui/src/app/state/receive.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::str::FromStr;
 use std::sync::Arc;
 
 use iced::{widget::qr_code, Subscription, Task};
@@ -286,9 +287,21 @@ impl State for ReceivePanel {
                     Task::none()
                 }
             }
-            Message::View(view::Message::ShowQrCode(i)) => {
-                if let (Some(address), Some(index)) = (self.address(i), self.derivation_index(i)) {
-                    if let Some(modal) = ShowQrCodeModal::new(address, *index) {
+            Message::View(view::Message::ShowQrCode(i, bip21)) => {
+                if let Some(address) = self.address(i) {
+                    if bip21.is_some() {
+                        if let Some(modal) =
+                            ShowQrCodeModal::new(&bip21.clone().unwrap_or(address.to_string()))
+                        {
+                            self.modal = Modal::ShowQrCode(modal);
+                        } else {
+                            tracing::error!(
+                                "Failed to create QR modal for BIP21 '{:?}' (address {})",
+                                bip21,
+                                address
+                            );
+                        }
+                    } else if let Some(modal) = ShowQrCodeModal::new(&address.to_string()) {
                         self.modal = Modal::ShowQrCode(modal);
                     }
                 }
@@ -420,13 +433,21 @@ pub struct ShowQrCodeModal {
 }
 
 impl ShowQrCodeModal {
-    pub fn new(address: &Address, index: ChildNumber) -> Option<Self> {
-        qr_code::Data::new(format!("bitcoin:{}?index={}", address, index))
-            .ok()
-            .map(|qr_code| Self {
+    pub fn new(address: &str) -> Option<Self> {
+        if Address::from_str(address).is_ok() {
+            qr_code::Data::new(format!("bitcoin:{}", address))
+                .ok()
+                .map(|qr_code| Self {
+                    qr_code,
+                    address: address.to_string(),
+                })
+        } else {
+            // Already in bip21 format
+            qr_code::Data::new(address).ok().map(|qr_code| Self {
                 qr_code,
                 address: address.to_string(),
             })
+        }
     }
 
     fn view(&self) -> Element<view::Message> {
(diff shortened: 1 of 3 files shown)


liana-gui/src/app/state/receive.rs
```diff
@@ -39,6 +40,7 @@ pub enum Modal {
 #[derive(Debug, Default)]
 pub struct Addresses {
     list: Vec<Address>,
+    bip21s: HashMap<Address, String>,
```
pythcoiner: I dont think we should store uri strings, they can be generated in the view, it's not ressources intensive

liana-gui/src/app/state/receive.rs
```diff
@@ -420,13 +433,20 @@ pub struct ShowQrCodeModal {
 }
 
 impl ShowQrCodeModal {
-    pub fn new(address: &Address, index: ChildNumber) -> Option<Self> {
-        qr_code::Data::new(format!("bitcoin:{}?index={}", address, index))
-            .ok()
-            .map(|qr_code| Self {
+    pub fn new(address: &str, index: ChildNumber) -> Option<Self> {
+        if Address::from_str(address).is_ok() {
+            qr_code::Data::new(format!("bitcoin:{}?index={}", address, index))
```
pythcoiner: when an address is displayed as QR, its expected to be passed to a peer for payment, I dont see a good reason to let him know our derivation index.

The index can be usefull only for a signing device to verify the address, like https://github.com/wizardsardine/liana/issues/1040 or with Coldcard-Q/Krux devices, but in this case we should also pass the descriptor name. (Not related to this PR, but maybe we should open an issue more generic tan #1040 since we support more QR devices @nondiremanuel)

liana-gui/src/app/state/receive.rs
benalleng: If we don't store the uri string how do you think we should flag that the address being displayed should have some extra bip21 information, just a simple bool flag to indicate some additional method should be called?

pythcoiner: > If we don't store the uri string how do you think we should flag that the address being displayed should have some extra, just a simple bool flag to indicate some additional method should be called?

I think so, but I'm not sure to have the whole context in mind, what additional infos can the uri potentially contains in our usage? I imagine:
 - the amount user expect to receive (we do not yet support this feature, do we want it in the future @nondiremanuel?)
 - information for receive payjoin (Directory url + auth metadata?), iiuc this is per-session based, do we store it in the DB? can several addresses share sames set of metadata? Can we have to deal with different set of metadata while the receive screen is still open? can an address have to been appended with different set of metatdat in some edge cases?

benalleng: You're right for all that. I was mostly asking in reference to just the physical display to keep the Qr code methods simple to just add some `is_bip21` flag that just appends the relevant data on at the end since `Message::ShowQrCode` would need to either be passed or generate the bip21 data as it is built for a simple address index right now

pythcoiner: what about this:

> do we store it in the DB? can several addresses share sames set of metadata? Can we have to deal with different set of metadata while the receive screen is still open? can an address have to been appended with different set of metatdat in some edge cases?

nondiremanuel: Is the expected behavior for this kind of transactions that the receiver inserts the amount? If not, I would stick with the current UX without amount on the receive side.

pythcoiner: Bullbitcoin wallet allow user to add amount & note for a PJ (receive), but I mean it can be another feature (not related to PJ) also

benalleng: > do we store it in the DB? can several addresses share sames set of metadata? Can we have to deal with different set of metadata while the receive screen is still open? can an address have to been appended with different set of metatdat in some edge cases?

So no, each address should have a new pj session, reused addresses will cause problems as far as I know. There aren't really any practical edge cases where I could think of wanting to do that practically

> Can we have to deal with different set of metadata while the receive screen is still open?

As for this I think the only 3 things that are really notable to happen during this stage
1. the directory rejects / never receives our payjoin initialization
2. the sender responds before we even exit this screen so we jump to the second signing stage immediately
3. All other states are simply waiting for the sender

benalleng: The amount is not explicitly required in a pj, so for now I don't think we need to add it unless we think that an amount would dramatically change or improve the UX, especially with looking at the receive screen as it stands now it would likely require some design work to think about how to add a fixed amount seamlessly when it is not necessary for payjoins

pythcoiner: > each address should have a new pj session,

where we store metadata about the PJ session? in RAM? in DB?
just to get back to the topic of this thread, I dont think it makes sense to store it here, as `ReceivePanel` is reset on reload: https://github.com/wizardsardine/liana/blob/ca768d628132e3a516a19f298de090fe5aae4f92/liana-gui/src/app/state/receive.rs#L313

benalleng: A payjoin session data would ideally be stored in the DB but the bip21 associated with it would ideally not be recalled as a user really shouldn't be reusing it outside of the initial receive screen as it would be essentially restarting a sessions and could encourage some bad user patterns like giving a bip21 to 2 separate senders.

I think it might be helpful to perhaps open a more complete receiver PR to give some more context on some design decisions.

benalleng:
This is superseded by #2011 as it implements this code while also expanding the scope to get a better understanding of what a good UX here is.

PR #1924, opened 2025-11-17
test: fix flaky test_rpc.py/test_start_rescan

pythcoiner:
This PR fixes the flaky test in `test_rpc.py`.
Coins & transactions are created with random amount in this test and some times it ended up a single input where spend with and amount > DUST but with not enough room for fees. Coins w/ value <= 1_000 sats are now sorted out from spending in this test.

Note: this PR also revert 1a888fd3155185f399fdd050992ce52fec4a627a changes

### DIFF
--- a/tests/test_rpc.py
+++ b/tests/test_rpc.py
@@ -1017,7 +1017,7 @@ def test_start_rescan(lianad, bitcoind):
     # First, get some coins
     for _ in range(10):
         addr = lianad.rpc.getnewaddress()["address"]
-        amount = random.randint(500, COIN * 10) / COIN
+        amount = random.randint(1, COIN * 10) / COIN
         txid = bitcoind.rpc.sendtoaddress(addr, amount)
         bitcoind.generate_block(random.randint(1, 10), wait_for_mempool=txid)
     wait_for(lambda: len(list_coins()) == 10)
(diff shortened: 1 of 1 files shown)


tests/test_rpc.py
```diff
@@ -1027,9 +1027,9 @@ def all_spent(coins):
     # without change, single or multiple inputs, sending externally or to self).
     for _ in range(5):
         addr = lianad.rpc.getnewaddress()["address"]
-        amount = random.randint(500, COIN * 10) / COIN
+        amount = random.randint(1, COIN * 10) / COIN
         txid = bitcoind.rpc.sendtoaddress(addr, amount)
-        avail = list(unspent_coins())
+        avail = list(c for c in unspent_coins() if c["amount"] > 1_000)
         to_spend = random.sample(avail, random.randint(1, len(avail)))
```
jp1ac4, line 1033: Although it's unlikely, it might be worth checking that `avail` is not empty before sampling coins as I think this line could throw an error in that case.

edouardparis (approved):
utACK f20a18e66f982659fb5d8cc340e445b2d4940277

tests/test_rpc.py
```diff
@@ -1027,9 +1027,9 @@ def all_spent(coins):
     # without change, single or multiple inputs, sending externally or to self).
     for _ in range(5):
         addr = lianad.rpc.getnewaddress()["address"]
-        amount = random.randint(500, COIN * 10) / COIN
+        amount = random.randint(1, COIN * 10) / COIN
```
edouardparis, line 1030: So, the maximum possible value of amount is 10.

pythcoiner: 10 BTC, it is the original value

tests/test_rpc.py
pythcoiner: (my original response from early this week seems have never reach GH, likely cloudflare related)

if figure ou during debugging that breaking early in the for loop let the test fail later, so I'm not in favor of doing so, but rather to make it "extremelly unlikelly" to happen, like by dropping the max spend amount from 10 BTC to 1

jp1ac4: OK sure, we can come back to it if need be.

edouardparis (approved):
utACK f20a18e66f982659fb5d8cc340e445b2d4940277
Merging so people can rebase to fix their PR ci

PR #1928, opened 2025-11-26
gui: make it more understandable when maxxed output is under dust limit

### LINKED ISSUE #1910: Throw "Invalid amount" error when "MAX" of selected coins is dust

pythcoiner:
This PR add a warning message if maxxed output is uder dust limit:

image

image

closes #1910

### DIFF
--- a/liana-gui/src/app/state/spend/step.rs
+++ b/liana-gui/src/app/state/spend/step.rs
@@ -237,6 +237,7 @@ impl DefineSpend {
 
     pub fn self_send(mut self) -> Self {
         self.recipients = Vec::new();
+        self.is_user_coin_selection = true;
         self
     }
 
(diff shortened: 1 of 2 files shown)


jp1ac4 (commented):
This looks to be working fine for a single recipient that has MAX selected.

If there are two recipients, there's a slight inconsistency which I'm not sure is intended:

1. When there are sufficient funds to pay the non-MAX recipient, then the other recipient shows 0 BTC as per the current behaviour:

image

2. When there are also insufficient funds to pay the non-MAX recipient, then there's a warning message shown for the MAX recipient:

image

For this second case, I was expecting to see something like this:

image

jp1ac4:
> see the comment when the issue has been reopened, we only want to handle for single output, multiples outputs are out of scope: [#1910 (comment)](https://github.com/wizardsardine/liana/issues/1910#issuecomment-3527868416)

This is why I was expecting the behaviour in the second case to match the current behaviour and didn't expect the new inline error message to appear.

jp1ac4 (approved):
ACK ea0001697ce49f30d98fc97d8c42f170c257f13a.

PR #1944, opened 2026-01-06
making the license file relative to reference the existing file in th…

ulrichard:
"cargo deb" failed with an error indicating the license file could not be found.
Each of the packages seems to specify the license differently. It might be good to unify them.


edouardparis (approved):
ACK 76d99caa80bc466f29d93391102dce5c10e5947a

PR #1948, opened 2026-01-12
liana-gui: refactor main

pythcoiner:
this PR refactor `liana-gui::main` & makes `parse_args()` composable & move it to `lib` in order to be exportable, it's preparatory work towards liana-business integration

### DIFF
--- a/liana-gui/src/main.rs
+++ b/liana-gui/src/main.rs
@@ -1,6 +1,6 @@
 #![windows_subsystem = "windows"]
 
-use std::{error::Error, io::Write, path::PathBuf, process, str::FromStr};
+use std::{error::Error, fmt::Display, io::Write, path::PathBuf, process, str::FromStr};
 
 #[cfg(target_os = "linux")]
 use iced::window::settings::PlatformSpecific;
@@ -10,7 +10,7 @@ use tracing_subscriber::filter::LevelFilter;
 extern crate serde;
 extern crate serde_json;
 
-use liana::miniscript::bitcoin;
+use liana::miniscript::bitcoin::Network;
 use liana_ui::{component::text, font, image, theme};
 
 use liana_gui::{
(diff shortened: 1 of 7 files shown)


liana-gui/src/main.rs
```diff
@@ -1,141 +1,46 @@
...
-        Some(LevelFilter::from_str(&l)?)
-    } else {
-        None
-    };
-
-    setup_panic_hook(&config.liana_directory);
+    let args = parse_args(
+        std::env::args().collect(),
+        VERSION,
+        &[Bitcoin, Testnet, Signet, Regtest],
```
jp1ac4: Should this include both `Testnet` and `Testnet4`?

pythcoiner: fixed, we only allow Testnet4 now

jp1ac4: We still allow `Testnet` for those users who have an existing wallet on this network, but I suppose they'll still be able to access it via the Launcher.

jp1ac4 (approved):
ACK 2b4d022bf513096f13c23e8f381670853f279f7a.

jp1ac4 (approved):
Tested ACK 6105f7a4d8090d0b7073c275199e9258b4504d01.

edouardparis (commented):
I would prefer as final arch to have:
- src/args.rs instead of src/utils/args.rs
- from utils/app.rs:
    - move `args_to_config` to src/args.rs or src/gui/mod.rs (where the Config is)
    - move `parse_log_level` in src/logger.rs
    - move `create_app_settings`, `load_initial_size`and `create_window_settings` in a `src/window.rs` file.
   
My opinion is that some of the methods are better in other modules (like the parse_log_level) and for the window.rs methods they are too specific to the iced window manager to be utils that will be reused across other modules.
Thanks

liana-gui/src/main.rs
pythcoiner: I've just looked at this, in v13.0 we still allow user to create both Testnet + Tesnet4 in the GUI, this behavior is not changed by this PR, the list we provide here only restrict which network can be passed as cli arg, I've added a note

edouardparis (approved):
ACK 28f598a877bd51d1060a22a65b20d74d8fef79c3

PR #1950, opened 2026-01-12
Some changes related to fiat prices

### LINKED ISSUE #1949: CoinGecko API requires user-agent
The CoinGecko API is returning an error due to mising "user-agent" request header:

Image

jp1ac4:
This PR makes the following changes to the use of fiat prices:
1. implements #1961 by adding disclaimer on fiat settings page
2. changes default price source from CoinGecko to mempool.space (give preference to source that doesn't require user-agent header)
3. fixes #1949 by adding a user-agent header to CoinGecko requests

Following internal discussions, it was decided to use a Liana-related value for (3) together with the disclaimer from (1).

### DIFF
--- a/liana-gui/src/services/fiat/source.rs
+++ b/liana-gui/src/services/fiat/source.rs
@@ -116,4 +116,12 @@ impl PriceSource {
         };
         Ok(ListCurrenciesResult { currencies })
     }
+
+    /// Returns the User-Agent header to use in requests, if any.
+    pub fn user_agent(&self) -> Option<String> {
+        match self {
+            Self::CoinGecko => Some(format!("liana-gui/{}", crate::VERSION)),
+            Self::MempoolSpace => None,
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


liana-gui/src/services/fiat/source.rs
```diff
@@ -116,4 +116,12 @@ impl PriceSource {
         };
         Ok(ListCurrenciesResult { currencies })
     }
+
+    /// Returns the User-Agent header to use in requests, if any.
+    pub fn user_agent(&self) -> Option<String> {
+        match self {
+            Self::CoinGecko => Some(format!("liana-gui/{}", crate::VERSION)),
```
pythcoiner, line 123: Maybe we should fake a firefox/chrome std user agent in order to not dox the user is using Liana?

jp1ac4: I've left this value unchanged as per our internal discussions.

liana-gui/src/app/view/settings/general.rs
```diff
@@ -1,10 +1,12 @@
-use iced::widget::{pick_list, Column, Row, Space, Toggler};
+use iced::widget::{pick_list, tooltip, Column, Row, Space, Toggler};
```
pythcoiner, line 1: shouldn't we use the tooltip from liana_ui (I think it has a background)

jp1ac4: I've added a `tooltip_custom` function to liana_ui that allows to change icon and position.

jp1ac4:
> I would personally have put the warning icon close to the "Fiat Price:" text (on its right) instead of close to the toggle. But that's not a big deal.

I've now moved the icon to the left-hand side:

image

I also use a new function from liana_ui for the tooltip instead of using iced directly.

Issue #1962, opened 2026-01-13
Remove address index when scanning QR code (for privacy)


nondiremanuel:
From a related user report:

> When QR codes are scanned for addresses on liana, on blue wallet and on binance for example, the bitcoin address is not formatted correctly and shows "bitcoin:bc1...?index=55"

We should not display the index to an external for privacy in the Send process. The index should only be available in the verify address modal.

nondiremanuel:
> my bad, there is no way to display the QR Code in the verify address modal:
> <img alt="Image" width="991" height="885" src="https://private-user-images.githubusercontent.com/124568858/535234066-71296248-8067-4503-bbb9-a9f037ecf166.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjgzNzg3NjIsIm5iZiI6MTc2ODM3ODQ2MiwicGF0aCI6Ii8xMjQ1Njg4NTgvNTM1MjM0MDY2LTcxMjk2MjQ4LTgwNjctNDUwMy1iYmI5LWE5ZjAzN2VjZjE2Ni5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwMTE0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDExNFQwODE0MjJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT02OTI3NGIyYzNlZmIyNjllNWQ4NTM4ZjgwNWZjMTJjMzU2NDAyNzFjNzJmNzI0MGZlNTE0Y2RiYzM0NzY1MDQxJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.XPuqripG3-QCunJ_SVEFzSczwwDFcAfX1aPO4bY4_yE">
> 
> we should add a button for open the QR Code modal from here (and the index must be present in order specter devices being able to verify address), or directly display the QR

[...]

PR #1965, opened 2026-01-13
remove address index from qr code

jp1ac4:
This implements #1962.


nondiremanuel:
> > any change to the verify modal can be done separately
> 
> Note: we should not merge this PR until the change on the verify modal is done.

Could you elaborate on why is that?

Issue #1963, opened 2026-01-13
When pasting an amount with more than 8 decimals, we should truncate&round


nondiremanuel:
if I try to paste an amount with more than 8 decimals in the Send page, the paste doesn't work at all. This could be common if I do a conversion from fiat outside Liana (dividing per the bitcoin price). Nothing blocking, but I would expect that the amount would be pasted anyway, truncated and rounded (if needed)

PR #1966, opened 2026-01-13
show pasted value in btc amount form even if invalid

jp1ac4:
This implements #1963 for BTC input values.

The validation on BTC input values performed by https://github.com/wizardsardine/liana/issues/798 will be skipped:
- for pasted values, so the whole pasted value will be inserted
- if the current value is invalid (i.e. an invalid value has been pasted), which will then allow the user to manually edit the pasted value until it is valid

This PR doesn't change the paste behaviour for the fiat value input field.

### DIFF
--- a/liana-gui/src/app/state/spend/step.rs
+++ b/liana-gui/src/app/state/spend/step.rs
@@ -969,6 +969,7 @@ impl Recipient {
                 }
             }
             view::CreateSpendMessage::RecipientFiatAmountEdited(_, fiat_amt_str, converter) => {
+                self.amount.warning = None; // Clear any warning on the BTC amount as it is no longer the last edited field.
                 self.fiat_converter = Some(converter);
                 if fiat_amt_str.is_empty() {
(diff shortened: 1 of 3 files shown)


jp1ac4:
I've modified the paste logic so that float values are rounded to be no more than 8 decimal places (rounding is also applied if the pasted value is negative).

In order to allow the user to edit and fix an invalid value after pasting, I've modified the editing logic to allow deletion of characters even if the new value is still invalid. In order to keep this editing logic as simple as possible, the user must first delete all invalid characters before being able to add any new ones.

I also rebased on master.

pythcoiner:
weird behavior, I pasted the same value in both fields:
image

n the amount field the `8` is dropped

nondiremanuel:
> weird behavior, I pasted the same value in both fields: image
> 
> n the amount field the `8` is dropped

what's weird about it? It looks like the expected behavior as per the issue it's solving.

kloaec:
Ok that works. But yes always tell the user of something was done
differently than what they input.

PR #1985, opened 2026-02-05
business-installer: make card first row more visible

### LINKED ISSUE #1978: [liana-business] path policy should be more visible in the template
I think we should make the first row of each path bold in order to make it more visible (waiting for a possible different layout in the future).

Image

pythcoiner:
closes #1978

Image

Image

### DIFF
--- a/liana-business/business-installer/src/views/keys/mod.rs
+++ b/liana-business/business-installer/src/views/keys/mod.rs
@@ -92,9 +92,9 @@ fn key_card(
     // Header row: |<icon>|<Key_name>|<identity>|<spacer>|<key_type_badge>
     let header_row = Row::new()
         .spacing(10)
-        .align_y(Alignment::Center)
+        .align_y(Alignment::End)
         .push(icon::key_icon())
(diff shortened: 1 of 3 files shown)


nondiremanuel:
> image

Sorry, I just realized that by changing the color the alias is no more in bold (at least in the screenshot). I would personally keep the bold for the alias.

I also think we should align the same behavior to the manage keys page:
image
Alias should be bold and email address colored.

Let me know if it's better to open a separate issue for this

nondiremanuel:
> updates:
> image image image

Thanks.  In the second and third image the email address seems oddly aligned with the alias. Maybe it's not vertically centered.

pythcoiner:
> Thanks. In the second and third image the email address seems oddly aligned with the alias. Maybe it's not vertically centered.

It should be bottom aligned, I think it's not pixel perfect, but I've tryed centered and it's not look great either

edouardparis (commented):
utACK 3b2e2d7698ce31a07877f38fbfaba10b12c6691b

PR #1987, opened 2026-02-05
liana-ui: use ORANGE for business pills::warning

### LINKED ISSUE #1983: "Set keys" state should be amber/yellow
The "Set Keys" state for a wallet doesn't mean there is an error, but an action required. So it should not have a red state but a yellow/amber (i.e. pending) state.

Image

pythcoiner:
closes #1983

image


edouardparis (approved):
utACK 50faebb12087fc29ae62dc65d61d61bb2760257c

PR #1989, opened 2026-02-05
business-installer: differentiate owned vs non-owned keys in xpub view

### LINKED ISSUE #1981: [liana-business] In the set keys step, we should tell the manager we invited the members to add their keys
In the Set Keys step, we should tell the manager (who sees all keys) that the other members were notified to add their own keys so that he knows he doesn't necessarily need to set all keys.

Image

I would change the text to:
"Select a key to complete its setup. Keys can be set up by each key manager individually, or by the wallet manager on their behalf.
You can connect a hardware device (recommended) or (... as is)"

pythcoiner:
closes #1981 

image


edouardparis (approved):
utACK ed9dd2338dfa3f9b4107fec6be8a6a2348cb4ecb

PR #1995, opened 2026-02-06
business-installer: trigger an Update when wallet registered 

pythcoiner:
previously when registering a wallet with a ledger, auto transition to the waller was not functionning, we now trigger an update


edouardparis (approved):
utACK 6656b1f094b6b91f6c39c06fe4d383dc3604e77a

PR #1996, opened 2026-02-06
business-settings: Policy settings

### LINKED ISSUE #1992: [liana-business] Add policy description to Settings>Wallet page
In this page, we should have the policy written in "human" form as we have in Liana :

Image

pythcoiner:
This PR add a human readable policy section in Settings > Wallet & let user copy/backup the descriptor

image

closes #1992

### DIFF
--- a/liana-business/business-settings/src/message.rs
+++ b/liana-business/business-settings/src/message.rs
@@ -1,5 +1,7 @@
 //! Message types for business settings UI.
 
+use liana_gui::{app::view::Close, export::ImportExportMessage};
+
 /// Settings section for navigation.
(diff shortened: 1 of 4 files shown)


pythcoiner:
I'll kept this in draft for now, it's not pressing and I'd like to not enable backup of encrypted descriptor before the BIP finalized

nondiremanuel:
I think that, even if not super urgent, having the possibility of reading the policy is still a good thing for users that want to be reassured about their setup after landing to the wallet. If it's not a lot of effort, I would personally consider splitting the PR and start merging that part (and possibly the plain text copy of the descriptor). Let's discuss this based on what's the effort required if needed.

PR #2005, opened 2026-02-09
business-installer: uses theme::accent for email (blue)

### LINKED ISSUE #1997: [liana-business] Email addresses should be blue instead of green
We now have several places in which we indicate the email addresses of the wallet members in green. I think they should all be blue to align with the liana-business theme. I remember seeing the email address indication in the following steps:
- At login, when requesting the token
- In the manage keys page, when seeing the list of keys (see https://github.com/wizardsardine/liana/pull/1985#issuecomment-3860623135)
- In the Set Keys step, when seeing the list of keys (see #1985)

This color may be hard-coded, if I remember correctly from my conversations with Claude.
The color should be the same of the one used on the top right when logged in with an account:

Image

pythcoiner:
closes #1997

image

image

image

### DIFF
--- a/liana-business/business-installer/src/views/keys/mod.rs
+++ b/liana-business/business-installer/src/views/keys/mod.rs
@@ -76,7 +76,7 @@ fn key_card(
     // Identity (optional - display email or other identity)
     let identity_str = key.identity.to_string();
     let identity_display = (!identity_str.is_empty())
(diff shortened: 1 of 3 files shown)


edouardparis (approved):
ACK 2a1592f1b86f8819c007c4000aef75c6cd362acb

PR #2010, opened 2026-02-10
business-installer: rework registration card

### LINKED ISSUE #1979: [liana-business] fingerprint and device name should be switched in Set Keys modal
As in the new Liana set keys modal, the device model should be the primary info and the fingerprint the secondary 
```

Jade
#fingerprint
```

Currently it's the other way around:
Image

### LINKED ISSUE #2001: [liana-business] Descriptor registration step is not very clear
I think we should redesign the descriptor registration step since right now it's not very clear what the user should do:

Image

I think we should have:

```
**Alias**
Device type - Fingerprint
```

on the right, and

"Connect device to register" on the right that becomes a "Register" button when the specific device is detected.

Below we should have a checkbox saying "I have registered the descriptor on my device(s)" that enables a "Next" button when checked. We can evaluate whether to still keep the "Skip" or not.

I've created a high-level mockup with Claude which we could follow to have an idea of the info and disposition of elements:

[register-devices.html](https://github.com/user-attachments/files/25132505/register-devices.html)

### LINKED ISSUE #2007: [liana-business] We should not allow to set a key with a device (Jade) / firmware version that doesn't support taproot
If the user uses a device (currently Jade) or a firmware version that is not supporting taproot, we should not allow him to set a key with it. We currently allow it:

Image

For now, I think we should have the same errors we throw in Liana (e.g. "This device doesn't support taproot") and disable the selection.

pythcoiner:
closes #1979 
closes #2001
closes #2007

image

image

image

### DIFF
--- a/liana-business/business-installer/src/state/mod.rs
+++ b/liana-business/business-installer/src/state/mod.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use async_hwi::{bitbox::NoiseConfig, service::HwiService};
 use crossbeam::channel;
+use liana_connect::ws_business::Wallet;
 use liana_gui::{app::settings::global::PersistedBitboxNoiseConfig, dir::LianaDirectory};
 use liana_ui::widget::{modal::Modal, Element};
 pub use message::{Message, Msg};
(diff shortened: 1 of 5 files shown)


pythcoiner:
> Should we change the last image's text to "Connect the associated device to register" 

done

> I think that "Register" is not very clear it's an action and what the user needs to do, but maybe that's part of the re-design of the UI/UX.

tell me if you want change

edouardparis (approved):
utACK aaa9b3bf6ed889e1064ce79235adfd2977592ffc

PR #2011, opened 2026-02-10
[WIP] Payjoin receiver

benalleng:
This adds support for payjoin receive in liana.

This is currently WIP and has the following TODOs

- [x] Move bip21 to be created at runtime instead of at the time of receiver creation in lianad allowing us to not need to store bip21
- [x] Ensure that the expiry is properly handled as it currently panics when an incomplete session expires


lianad/src/database/sqlite/schema.rs
```diff
@@ -122,6 +133,34 @@ CREATE TABLE labels (
...
+    id INTEGER PRIMARY KEY NOT NULL,
+    relay_url TEXT UNIQUE NOT NULL,
+    timestamp INTEGER NOT NULL,
+    key BLOB NOT NULL
+);
+
+/* Payjoin receivers */
+CREATE TABLE payjoin_receivers (
+    id INTEGER PRIMARY KEY NOT NULL,
+    derivation_index INTEGER,
```
arminsabouri: Would it make sense if this was a FK relationship with addresses?

lianad/src/database/sqlite/schema.rs
```diff
@@ -122,6 +133,34 @@ CREATE TABLE labels (
...
+    bip21 TEXT,
+    original_txid BLOB,
+    proposed_txid BLOB,
```
arminsabouri: This information should be accessible via the session manager after the events are replayed.

lianad/src/database/sqlite/schema.rs
```diff
@@ -87,6 +88,16 @@ CREATE TABLE coins (
...
 );
 
+/* Seen Payjoin outpoints
+ *
+ * The 'added_at' field is simply the time that this outpoint is added to the table for 
+ * tracking.
+ */
+CREATE TABLE payjoin_outpoints (
+    outpoint BLOB NOT NULL PRIMARY KEY,
+    added_at INTEGER NOT NULL
```
arminsabouri: For consistency

```suggestion
    created_at INTEGER NOT NULL
```

lianad/src/database/sqlite/utils.rs
```diff
@@ -482,6 +482,29 @@ fn migrate_v7_to_v8(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError
...
+
+fn migrate_v9_to_v10(conn: &mut rusqlite::Connection) -> Result<(), SqliteDbError> {
+    db_exec(conn, |db_tx| {
+        db_tx.execute("ALTER TABLE payjoin_receivers ADD COLUMN bip21 TEXT", [])?;
+        db_tx.execute("UPDATE version SET version = 10", [])?;
+        Ok(())
+    })?;
+
+    Ok(())
+}
```
arminsabouri: These migrations can be removed and the db version can stay the same

lianad/src/commands/mod.rs
```diff
@@ -359,7 +381,87 @@ impl DaemonControl {
...
+        let session = ReceiverBuilder::new(address.clone(), payjoin_dir_url, ohttp_keys)
+            .map_err(|e| CommandError::IntoUrlError(e.to_string()))?
+            .build()
+            .save(&persister)
+            .unwrap();
+
+        let bip21 = session.pj_uri().to_string();
+
+        let mut db_conn = self.db.connection();
+        db_conn.update_payjoin_receiver_bip21(new_index.into(), &bip21);
```
xstoicunicornx: Since this is now no-op this call is probably no longer needed?

lianad/src/commands/mod.rs
```diff
@@ -359,7 +381,87 @@ impl DaemonControl {
...
+
+        Ok(GetAddressResult::new(address, new_index, Some(bip21)))
+    }
+
+    /// Get Payjoin URI (BIP21) and its sender/receiver status by txid
```
xstoicunicornx:
```suggestion
    /// Get receiver session ID and its sender/receiver status by txid
```

lianad/src/commands/mod.rs
```diff
@@ -359,7 +381,87 @@ impl DaemonControl {
...
+            let (state, _) = replay_receiver_event_log(&persister)
+                .map_err(|e| CommandError::ReplayError(format!("Receiver replay failed: {e:?}")))?;
+            return Ok(state.into());
+        }
+
+        Ok(PayjoinStatus::Unknown)
+    }
+
+    /// Get all active payjoin receiver sessions with their derivation indexes
+    pub fn get_active_payjoin_sessions(&self) -> Result<Vec<u32>, CommandError> {
```
xstoicunicornx: I believe that the plan is to also support sending payjoin in liana correct? Should this fn name maybe be `get_active_receiver_payjoin_sessions` instead to remove ambiguity in future when there may also be active payjoin sender sessions?

lianad/src/database/sqlite/schema.rs
```diff
@@ -122,6 +133,24 @@ CREATE TABLE labels (
     item TEXT UNIQUE NOT NULL,
     value TEXT NOT NULL
 );
+
+/* Payjoin OHttpKeys */
+CREATE TABLE payjoin_ohttp_keys (
+    id INTEGER PRIMARY KEY NOT NULL,
+    relay_url TEXT UNIQUE NOT NULL,
```
xstoicunicornx, line 140: Should this be the payjoin directory url instead? Aren't OHttpKeys specific to a payjoin directory, not relay? 

If so`payjoin_save_ohttp_keys` and `payjoin_get_ohttp_keys` also would need to be updated.

lianad/src/database/sqlite/mod.rs
```diff
@@ -963,6 +994,198 @@ impl SqliteConn {
...
+
+            id = db_tx.last_insert_rowid();
+            Ok(())
+        })
+        .expect("Db must not fail");
+        id
+    }
+
+    /// Get bip21 for a receiver session by derivation index
+    pub fn get_payjoin_receiver_bip21(&mut self, derivation_index: u32) -> Option<String> {
```
xstoicunicornx: This seems to just return a normal address and not a bip21?

lianad/src/payjoin/helpers.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl std::error::Error for FetchOhttpKeysError {}
+impl std::fmt::Display for FetchOhttpKeysError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+pub(crate) fn fetch_ohttp_keys(
```
xstoicunicornx, line 33: Is it possible to reuse PDK's implementation of `fetch_ohttp_keys`?

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,392 @@
...
+    Ok(())
+}
+
+fn apply_fee_range(
+    proposal: Receiver<WantsFeeRange>,
+    persister: &ReceiverPersister,
+    db_conn: &mut Box<dyn DatabaseConnection>,
+    secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+) -> Result<(), Box<dyn Error>> {
+    let proposal = proposal.apply_fee_range(None, None).save(persister)?;
```
xstoicunicornx, line 243: No fee range actually being applied.

xstoicunicornx (commented):
I know this is still a WIP and my comments may be a bit premature but just wanted to brain dump some of the notes I took as part of the BOSS Payjoin Showcase session.

Some additional general comments:
- should amount be added to URI?
- should adjusting expiration will be added to settings?
- when I ran this PR Liana panicked with this error when I tried to send a payjoin to the receiver:`
2026-04-09T16:19:40.194941Z ERROR liana_gui:82: panic occurred at line 508 of file lianad/src/database/sqlite/mod.rs: Some("database must be available: SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some(\"table payjoin_outpoints has no column named added_at\"))")
   0: backtrace::backtrace::libunwind::trace
`
- in the GUI when payjoin is listed under previously generated addresses it shows the address rather than URI (I'm guessing this is related to this [comment](https://github.com/wizardsardine/liana/pull/2011/changes#r3059409105))

benalleng:
Thanks for the feedback,

Regarding the UI notes.
> should amount be added to URI?

Since it is not a mandatory addition in the payjoin spec I thought that for a desktop on-chain only wallet the UX of adding an amount on the receive side is sort of unfamiliar, especially given the current liana receiver UI

> should adjusting expiration will be added to settings?

I think again perhaps this is worth some user research but with the ability to do a fallback tx at any point I am not sure that extending the expiration as an option is great to put in the user's direct control

> in the GUI when payjoin is listed under previously generated addresses it shows the address rather than URI (I'm guessing this is related to this comment)

This was something we landed on due wanting to discourage reuse of payjoin sessions. I could share this bip21 with as many people as I want similar to address reuse but instead of just bad privacy you would get a bunch of payjoin failures

The method again is a holdout as the last time I gave this some serious focus I transitioned from storing the bip21 to the session ID directly and I previously wanted to have access to the bip21 in the db where I just generate it in the gui now and don't store it directly

benalleng:
> when I ran this PR Liana panicked with this error when I tried to send a payjoin to the receiver:2026-04-09T16:19:40.194941Z ERROR liana_gui:82: panic occurred at line 508 of file lianad/src/database/sqlite/mod.rs: Some("database must be available: SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some(\"table payjoin_outpoints has no column named added_at\"))") 0: backtrace::backtrace::libunwind::trace

This is definitely an artifact of me playing in the database between naming the timestamp `added_at` vs `created_at` https://github.com/wizardsardine/liana/pull/2011/changes#diff-4f33457fe83ad075401f91984457b7503d757068fcb26def0e090c584fd8439dR1146

yashrajd:
> > should amount be added to URI?
> 
> Since it is not a mandatory addition in the payjoin spec I thought that for a desktop on-chain only wallet the UX of adding an amount on the receive side is sort of unfamiliar, especially given the current liana receiver UI

Familiarity might not be an issue IMO, since Liana users presumably also use other (esp mobile) wallets that have this. IMO it's more about the use case or user type: how often do they want to receive an specific exact amount? That said, this seems like a reasonable starting point.

> > should adjusting expiration will be added to settings?
> 
> I think again perhaps this is worth some user research but with the ability to do a fallback tx at any point I am not sure that extending the expiration as an option is great to put in the user's direct control

My hunch is, Liana users are likely to be sophisticated enough to use such a feature properly. Perhaps there could be appropriate messaging/guidance about using the feature.
From user research side, I'd say we have zero/little data to go on with, but you can always survey users to know their preferences (albeit with the "faster horse-carriages" caveat) (again without existing user base it is hard to rely on stated user preferences).

lianad/src/payjoin/helpers.rs
benalleng: Yes, but the current PDK impl is async so we would need to add some blocking capability and without it the async stuff would cascade. So i think its best to keep it manually re-implemented. though It might be nice to add sync versions for situations like this

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,392 @@
...
+            let proposal = proposal
+                .finalize_proposal(|_| {
+                    let mut psbt = psbt.clone();
+                    finalize_psbt(&mut psbt, secp);
+                    Ok(psbt)
+                })
+                .save(persister)?;
+
+            send_payjoin_proposal(proposal, persister)?;
+        }
```
xstoicunicornx, line 280: ~~I couldn't quite figure out what the intention was here with `is_signed`, however during my testing `is_signed` never gets set to true and the proposal never gets finalized. Can you explain what validation is being done here?~~

~~Also, when I replaced `|_|` with `|psbt|` in the `finalize_psbt` closure it was still possible to clone as mut, whats the rationale for calling `psbt_to_sign` outside of this closure?~~

~~Something else that may be related is that currently `psbt_to_sign` doesn't return a psbt with sender sigs removed, this is something I am fixing in https://github.com/payjoin/rust-payjoin/pull/1446. Let me know if it would good for me to create a separate smaller PR for this just to get this fixed sooner.~~

~~I did also try hardcoding `is_signed` to true so the receiver does send out a psbt proposal, but its then erroring on the sender side during validation. So the `finalize_psbt` isn't quite working right, but that may be because the psbt thats being finalized has the leftover sender sigs still.~~

Apologies this was due to my incomplete understanding of the PSBT signing workflow

xstoicunicornx (commented):
I didn't re-review the code yet, just tried to complete receiving a payjoin and was still not completing successfully. See details below.

xstoicunicornx:
Just started looking at this again. Good job with the broadcast fallback tx button, I like it. Couple quick notes about it:
- broadcast fallback button should be available prior to signing so that its not necessary to sign to in order to broadcast
- when I try to use the broadcast fallback button I am getting an error (unfortunately not copy/pasteable so just attaching screenshot)
image

benalleng:
I think I will want to try and use the latest commit for rust-payjoiin that includes https://github.com/payjoin/rust-payjoin/pull/1470 so that I can just have `broadcast_fallback` send a `cancel()` and then send the fallback_tx it returns, It seems cleaner to do it this way as a native API call plus it leaves us with a nice cleanup of the payjoin session.

DanGould:
sure we'll cut another release that includes the new function asap

lianad/src/commands/mod.rs
```diff
@@ -356,7 +402,117 @@ impl DaemonControl {
...
+        db_conn.set_receive_index(new_index, &self.secp);
+        let address = self
+            .config
+            .main_descriptor
+            .receive_descriptor()
+            .derive(new_index, &self.secp)
+            .address(self.config.bitcoin_config.network);
+
+        let persister = ReceiverPersister::new(Arc::new(self.db.clone()), new_index.into(), "");
+        let session = ReceiverBuilder::new(address.clone(), payjoin_dir_url, ohttp_keys)
```
xstoicunicornx, line 455: Is there possibility that ohttp_keys on payjoin directory have been rotated more recently than the 7 days for which the ohttp keys have been stored here? If so there probably needs to be handling for this.

Some options:
- Maybe don't store ohttp keys at all? Fetching shouldn't be happening that frequently and would simplify code
- Try refetching keys if `RecieverBuilder::new()` returns error

Also mapping the error to a `CommandError::IntoUrlError` by default feels weird since the pj directory url ideally should have already been validated upstream.

lianad/src/payjoin/db.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl Display for PersisterError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            PersisterError::Serialize(e) => write!(f, "Serialization failed: {e}"),
+        }
+    }
+}
+
+impl std::error::Error for PersisterError {}
```
xstoicunicornx: Is empty impl block needed?

lianad/src/payjoin/helpers.rs
```diff
@@ -0,0 +1,130 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FetchOhttpKeysError {
+    Reqwest(String),
+    InvalidOhttpKeys(String),
+    InvalidUrl(String),
+    UrlParseError,
+    UnexpectedStatusCode(reqwest::StatusCode),
+}
+
+impl std::error::Error for FetchOhttpKeysError {}
```
xstoicunicornx, line 79: Is empty impl block needed?

lianad/src/commands/mod.rs
```diff
@@ -898,14 +1054,25 @@ impl DaemonControl {
...
+                        return Err(CommandError::SpendFinalization(e.to_string()));
+                    }
+                    log::debug!(
+                        "Input at index {} already finalized, skipping: {}",
+                        index,
+                        e
+                    );
+                }
+            }
+        }
```
xstoicunicornx, line 1098: Why are we touching `broadcast_spend`? I don't think there is a payjoin flow that requires this?

Broadcasting the fallback transaction shouldn't require finalizing the inputs, and it would probably be cleaner to implement separate `broadcast_payjoin_fallback` if necessary.

lianad/src/database/mod.rs
```diff
@@ -194,6 +196,43 @@ pub trait DatabaseConnection {
 
     /// Dump all labels
     fn get_labels_bip329(&mut self, offset: u32, limit: u32) -> Labels;
+
+    /// Get the next Session Id
```
xstoicunicornx:
```suggestion
    /// Get OHttpKeys
```

lianad/src/database/sqlite/mod.rs
```diff
@@ -963,6 +994,183 @@ impl SqliteConn {
...
+        _txid: &bitcoin::Txid,
+    ) -> Option<SessionId> {
+        None
+    }
+
+    /// Create new Receiver Session
+    pub fn save_new_payjoin_receiver_session(
+        &mut self,
+        derivation_index: u32,
+        _bip21: &str,
```
xstoicunicornx: Why not remove entirely?

lianad/src/database/mod.rs
```diff
@@ -194,6 +196,43 @@ pub trait DatabaseConnection {
...
+    fn payjoin_get_ohttp_keys(&mut self, ohttp_relay: &str) -> Option<(u32, OhttpKeys)>;
+
+    /// Save OHttpKeys
+    fn payjoin_save_ohttp_keys(&mut self, ohttp_relay: &str, ohttp_keys: OhttpKeys);
+
+    /// Save Receiver Session
+    fn save_new_payjoin_receiver_session(&mut self, derivation_index: u32, bip21: &str) -> i64;
```
xstoicunicornx: The `bip21` param doesn't seem to be needed anymore.

lianad/src/database/sqlite/schema.rs
xstoicunicornx: Bumping this, aren't the ohttp keys specific to the payjoin directory?

lianad/src/database/sqlite/mod.rs
```diff
@@ -479,6 +482,34 @@ impl SqliteConn {
         .expect("Database must be available")
     }
 
+    pub fn insert_outpoint_seen_before<'a>(
+        &mut self,
+        outpoints: impl IntoIterator<Item = &'a bitcoin::OutPoint>,
```
xstoicunicornx: Maybe change this to be called on single outpoint at a time? Validation is being done on single outpoint at a time anyway.

Otherwise the `is_duplicate` returned value should be iterator to be able to match outpoint granularity.

lianad/src/database/sqlite/mod.rs
```diff
@@ -963,6 +994,183 @@ impl SqliteConn {
...
+                Ok(SessionId::new(id))
+            },
+        )
+        .expect("Db must not fail")
+    }
+
+    /// Deprecated: txid-in-events byte scan never worked since events serialize
+    /// txids as JSON hex while the caller passes raw bytes. Kept as a stub
+    /// returning None; callers should use the replay-based lookup in `payjoin::receiver`.
+    pub fn get_payjoin_receiver_session_id_from_txid(
```
xstoicunicornx: Why not remove entirely?

lianad/src/testutils.rs
```diff
@@ -550,6 +566,98 @@ impl DatabaseConnection for DummyDatabase {
...
+            .ohttp_keys
+            .insert(ohttp_relay.to_string(), (0, ohttp_keys));
+    }
+
+    fn insert_input_seen_before(&mut self, _outpoints: &[bitcoin::OutPoint]) -> bool {
+        false
+    }
+
+    fn get_active_payjoin_receiver_sessions(&mut self) -> Vec<(SessionId, u32)> {
+        Vec::new()
```
xstoicunicornx, line 604: Do we not need to actually get active payjoin receiver sessions in test utils?

lianad/src/testutils.rs
```diff
@@ -550,6 +566,98 @@ impl DatabaseConnection for DummyDatabase {
...
+    fn payjoin_save_ohttp_keys(&mut self, ohttp_relay: &str, ohttp_keys: OhttpKeys) {
+        self.db
+            .write()
+            .unwrap()
+            .ohttp_keys
+            .insert(ohttp_relay.to_string(), (0, ohttp_keys));
+    }
+
+    fn insert_input_seen_before(&mut self, _outpoints: &[bitcoin::OutPoint]) -> bool {
+        false
```
xstoicunicornx, line 599: Do we not need to actually insert inputs seen before in test utils?

liana/src/spend.rs
```diff
@@ -76,19 +75,21 @@ impl fmt::Display for SpendCreationError {
...
             }
-            Self::CoinSelection(e) => write!(f, "Coin selection error: '{e}'"),
+            Self::CoinSelection(e) => write!(f, "Coin selection error: '{}'", e),
             Self::SanityCheckFailure(psbt) => write!(
                 f,
-                "BUG! Please report this. Failed sanity checks for PSBT '{psbt}'.",
+                "BUG! Please report this. Failed sanity checks for PSBT '{}'.",
+                psbt
             ),
+            Self::InvalidBip21 => write!(f, "Invalid BIP21"),
```
xstoicunicornx: Does this need to be here since this is only for receiving payjoin and will be creating a bip21 not processing one?

liana-gui/src/app/state/spend/step.rs
```diff
@@ -1112,13 +1092,31 @@ impl SaveSpend {
...
+        };
+
         let mut tx = SpendTx::new(
             None,
             psbt,
             draft.inputs.clone(),
             &self.wallet.main_descriptor,
             &self.curve,
             draft.network,
+            payjoin_status,
```
xstoicunicornx: Does this need to part of receive payjoin pr?

liana-gui/src/app/state/settings/payjoin.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl PayjoinSettingsState {
+    pub fn new(config: Option<DaemonConfig>) -> Self {
+        let payjoin_config =
+            config
+                .and_then(|c| c.payjoin_config)
+                .unwrap_or_else(|| PayjoinConfig {
+                    ohttp_relay: "https://pj.bobspacebkk.com".to_string(),
+                    payjoin_directory: "https://payjo.in".to_string(),
+                });
```
xstoicunicornx:
```suggestion
                .unwrap_or_else(|| default_payjoin_config());
```

lianad/src/payjoin/helpers.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use miniscript::{
+    bitcoin::{secp256k1, Psbt, ScriptBuf, TxOut},
+    psbt::PsbtExt,
+};
+
+use payjoin::{bitcoin::Amount, IntoUrl, OhttpKeys};
+use reqwest::{header::ACCEPT, Proxy};
+
+pub(crate) const OHTTP_RELAY: &str = "https://pj.bobspacebkk.com";
```
xstoicunicornx: Why keep this instead of forcing grabbing this from the payjoin config?

liana-gui/src/app/state/psbt.rs
```diff
@@ -201,6 +203,40 @@ impl PsbtState {
...
+                        daemon
+                            .list_coins(&[CoinStatus::Spending], &outpoints)
+                            .await
+                            .map(|res| {
+                                res.coins
+                                    .iter()
+                                    .filter_map(|c| c.spend_info.map(|info| info.txid))
+                                    .collect()
+                            })
+                            .map_err(|e| e.into())
```
xstoicunicornx: Why do we need to list coins for broadcasting the fallback tx? Maybe related to [this comment](https://github.com/wizardsardine/liana/pull/2011/changes#r3142323883)?

xstoicunicornx (commented):
Looking much better :)

General feedback:
- ~~The `Broadcast fallback` button should be available prior to signing PSBT~~
- ~~Still getting error when broadcasting fallback transaction (maybe related to [this comment](https://github.com/wizardsardine/liana/pull/2011/changes#r3142323883)?)~~
- Should user be allowed to specify multiple relays?
- Transactions display for payjoin receive tx should be more aligned with existing Liana receive txs (example screenshots below):
  - it is classified as outgoing transaction
  - only our coins should show as spent (right now includes senders coins)
  - our received payment output should be classified as payment output (right now it shows as change)
  - senders change output should show as external output
- If we only show the bip21 initially when clicking `Receive Payjoin` does it need to be stored as part of gui app receive state?
- Somewhat related and contradictory to last point, should bip21 be displayed until expiry?
- Lots of `unimplemented!`s in liana-gui/src/app/state/receive.rs
- Should we pro-actively validate that key's can be fetched from payjoin directory when it is updated in settings? Not sure how feasible that is with this UI framework

Questions:
- I couldn't follow the purpose of changing the `descs: &[descriptors::SinglePathLianaDesc]` parameters to `desc: &descriptors::LianaDescriptor` in the poller, can you help me follow why this was needed?

Edit: Broadcast fallback is working appropriately - I neglected to build on latest commit when testing sorry.

Regular receive tx:
Image

Payjoin receive tx:
Image

lianad/src/commands/mod.rs
benalleng: Good point, I was hoping that I could tie into some existing logic. but you're probably right these are best kept separate.

lianad/Cargo.toml
```diff
@@ -52,3 +52,7 @@ jsonrpc = { workspace = true, features = ["minreq_http"], default-features = fal
 
 # import/export labels
 bip329 = { workspace = true, default-features = false }
+
+# payjoin
+payjoin = { version = "1.0.0-rc.1", features = ["v2", "io"] }
+reqwest = { version = "0.11", default-features=false, features = ["json", "rustls-tls", "stream", "blocking"] }
```
pythcoiner: could you move dependencies declaration to the workspace?

lianad/src/bitcoin/mod.rs
```diff
@@ -589,6 +598,10 @@ impl BitcoinInterface for electrum::Electrum {
     fn tip_time(&self) -> Option<u32> {
         self.client().tip_time().ok()
     }
+
+    fn test_mempool_accept(&self, _rawtxs: Vec<String>) -> Vec<bool> {
+        todo!()
```
pythcoiner: iirc there is no way to do this check w/ an electrum backend, then the only choice seems to document it and return a `Vec<bool>` I guess?

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+    secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+    ohttp_relays: &[String],
+) {
+    let mut db_conn = db.connection();
+
+    for session_id in db_conn.get_all_active_receiver_session_ids() {
+        let persister = ReceiverPersister::from_id(Arc::new(db.clone()), session_id.clone());
+
+        match replay_event_log(&persister) {
+            Ok(_) => match process_receiver_session(
```
pythcoiner: `process_receiver_session()` already call `replay_event_log()` internally, so I guess the outer match is pretty useless?

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+                &mut db_conn,
+                bit,
+                desc,
+                secp,
+                persister,
+                ohttp_relays,
+            ) {
+                Ok(_) => (),
+                Err(e) => {
+                    log::warn!("process_receiver_session(): {}", e);
```
pythcoiner:
```suggestion
                    log::error!("process_receiver_session(): {}", e);
```

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+            },
+            Err(e) => {
+                let error_str = e.to_string();
+                if error_str.contains("expired") {
+                    log::info!(
+                        "Payjoin session {:?} expired, marking as closed",
+                        session_id
+                    );
+                    if let Err(close_err) = persister.close() {
+                        log::warn!("Failed to close expired payjoin session: {}", close_err);
```
pythcoiner:
```suggestion
                        log::error!("Failed to close expired payjoin session: {}", close_err);
```

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+                    }
+                    continue;
+                }
+                log::warn!("Failed to replay payjoin session {:?}: {}", session_id, e);
```
pythcoiner:
```suggestion
                log::error!("Failed to replay payjoin session {:?}: {}", session_id, e);
```

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+}
+
+fn process_receiver_session(
+    db_conn: &mut Box<dyn DatabaseConnection>,
+    bit: &mut sync::Arc<sync::Mutex<dyn BitcoinInterface>>,
+    desc: &descriptors::LianaDescriptor,
+    secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+    persister: ReceiverPersister,
+    ohttp_relays: &[String],
+) -> Result<(), Box<dyn Error>> {
```
pythcoiner, line 514: any reason to NOT return a typed error? I think it could let us avoid to match over a string in `payjoin_receiver_check()`

pythcoiner:
when opening an already existing wallet I hit this:

```
 2026-06-09T07:22:14.878392Z ERROR liana_gui:93: panic occurred at line 1125 of file lianad/src/database/sqlite/mod.rs: Some("Db must not fail: SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some(\"no such table: payjoin_receivers\"))")
   0: liana_gui::setup_panic_hook::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: <lianad::database::sqlite::SqliteConn as lianad::database::DatabaseConnection>::get_all_receiver_sessions
   8: lianad::payjoin::receiver::find_receiver_session_by_txid
   9: <alloc::vec::into_iter::IntoIter<T,A> as core::iter::traits::iterator::Iterator>::fold
  10: alloc::vec::in_place_collect::<impl alloc::vec::spec_from_iter::SpecFromIter<T,I> for alloc::vec::Vec<T>>::from_iter
  11: lianad::commands::<impl lianad::DaemonControl>::list_transactions
  12: lianad::commands::<impl lianad::DaemonControl>::list_confirmed_transactions
  13: <liana_gui::daemon::embedded::EmbeddedDaemon as liana_gui::daemon::Daemon>::list_confirmed_txs::{{closure}}
  14: liana_gui::daemon::Daemon::list_history_txs::{{closure}}
  15: liana_gui::daemon::Daemon::list_confirmed_payments::{{closure}}
  16: <futures_util::future::future::map::Map<Fut,F> as core::future::future::Future>::poll
  17: <futures_util::stream::stream::map::Map<St,F> as futures_core::stream::Stream>::poll_next
  18: <futures_util::stream::stream::chain::Chain<St1,St2> as futures_core::stream::Stream>::poll_next
  19: <futures_util::stream::select_all::SelectAll<St> as futures_core::stream::Stream>::poll_next
  20: <futures_util::stream::stream::flatten::Flatten<St,<St as futures_core::stream::Stream>::Item> as futures_core::stream::Stream>::poll_next
  21: <futures_util::stream::stream::FlatMap<St,U,F> as futures_core::stream::Stream>::poll_next
  22: <futures_util::stream::stream::flatten::Flatten<St,<St as futures_core::stream::Stream>::Item> as futures_core::stream::Stream>::poll_next
  23: <futures_util::stream::stream::FlatMap<St,U,F> as futures_core::stream::Stream>::poll_next
  24: <futures_util::stream::stream::flatten::Flatten<St,<St as futures_core::stream::Stream>::Item> as futures_core::stream::Stream>::poll_next
  25: <futures_util::stream::stream::FlatMap<St,U,F> as futures_core::stream::Stream>::poll_next
  26: <futures_util::stream::stream::forward::Forward<St,Si,Item> as core::future::future::Future>::poll
  27: <futures_util::future::future::map::Map<Fut,F> as core::future::future::Future>::poll
  28: tokio::runtime::task::core::Core<T,S>::poll
  29: tokio::runtime::task::harness::Harness<T,S>::poll
  30: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  31: tokio::runtime::scheduler::multi_thread::worker::Context::run
  32: tokio::runtime::context::scoped::Scoped<T>::set
  33: tokio::runtime::context::runtime::enter_runtime
  34: tokio::runtime::scheduler::multi_thread::worker::run
  35: tokio::runtime::task::core::Core<T,S>::poll
  36: tokio::runtime::task::harness::Harness<T,S>::poll
  37: tokio::runtime::blocking::pool::Inner::run
  38: std::sys::backtrace::__rust_begin_short_backtrace
  39: core::ops::function::FnOnce::call_once{{vtable.shim}}
  40: std::sys::pal::unix::thread::Thread::new::thread_start
  41: <unknown>
  42: <unknown>
```

Maybe to avoid migration on the wallet DB, the PJ service could have it's own DB?

pythcoiner:
in GUI settings, it should have validation on url, at least a ping:

image

lianad/src/payjoin/helpers.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/// Run `f` against a shuffled copy of `relays`, returning the first `Ok`.
+/// `Retry` advances to the next relay; `Fatal` aborts immediately.
+/// The list is randomized per call so different sessions/ticks pick different orders.
+pub(crate) fn with_relay_fallback<T, E, F>(relays: &[String], mut f: F) -> Result<T, E>
+where
+    F: FnMut(&str) -> RelayAttempt<T, E>,
+{
+    assert!(!relays.is_empty(), "ohttp_relays must be non-empty");
```
pythcoiner: can we error rather than panic here?

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+    proposal: Receiver<UncheckedOriginalPayload>,
+    persister: &ReceiverPersister,
+    db_conn: &mut Box<dyn DatabaseConnection>,
+    bit: &mut sync::Arc<sync::Mutex<dyn BitcoinInterface>>,
+    desc: &descriptors::LianaDescriptor,
+    secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+) -> Result<(), Box<dyn Error>> {
+    // Receive Check 1: Can Broadcast
+    let proposal = proposal
+        .check_broadcast_suitability(None, |tx| {
```
pythcoiner: shouldn't enforce the same min fee rate than the wallet?

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,571 @@
...
+
+fn check_inputs_not_owned(
+    proposal: Receiver<MaybeInputsOwned>,
+    persister: &ReceiverPersister,
+    db_conn: &mut Box<dyn DatabaseConnection>,
+    desc: &descriptors::LianaDescriptor,
+    secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+) -> Result<(), Box<dyn Error>> {
+    let proposal = proposal
+        .check_inputs_not_owned(&mut |script| {
```
pythcoiner: My understanding is, that this is quite critical from a security POV, and I'm not sure we should rely on the DB to infer if the address is owned, maybe we should populate a map with addresses generated with the descriptor up to some value? 1000 (bitcoind look-ahead iirc)? 10_000?

benalleng:
I need to address the payjoin tx identification being locked behind the feature gate, this is not really payjoin related logic so should not really be behind a cfg flag
image

lianad/src/payjoin/receiver.rs
benalleng: refactored so that the state from the event log is completely passed from `payjoin_receiver_check` down to `process_receiver_session` so `replay_event_log` is only called once

lianad/src/payjoin/receiver.rs
benalleng: Added this logic separately in 767c234d6970a136bcb0437d577fe6cec23966ef

liana/src/spend.rs
bc1cindy, line 752: this TODO ties output shuffle to Taproot, but the two are independent: shuffling output position is a pure reorder, no Taproot needed, shippable now. Taproot only helps change script-type indistinguishability, a separate fingerprint.

suggest shuffling

ref: [payjoin comment](https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4618489879)

pythcoiner: ping @darosior , I guess it was a real reason even it it's not obvious now

pythcoiner: I guess if we add shuffling we should have a way to opt out, as I think our integrations tests rely on ordering for many cases, and imo it's good to keep this logic deterministicaly testable.

lianad/src/payjoin/receiver.rs
```diff
@@ -0,0 +1,610 @@
...
+fn contribute_inputs(
+    proposal: Receiver<WantsInputs>,
+    persister: &ReceiverPersister,
+    db_conn: &mut Box<dyn DatabaseConnection>,
+    desc: &descriptors::LianaDescriptor,
+    secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+) -> Result<(), Box<dyn Error>> {
+    let coins = db_conn.coins(&[CoinStatus::Confirmed], &[]);
+
+    let mut candidate_inputs_map = HashMap::<OutPoint, (Coin, TxIn, Input, Weight)>::new();
```
bc1cindy, line 189: makes sense to use `BTreeMap` here so the contributed input is deterministic, aligning with https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4661862568 ?

pythcoiner: can you open a separate issue for this?

bc1cindy: done! #2191

pythcoiner: thanks!

PR #2020, opened 2026-02-12
business-installer: do not update state before server update

### LINKED ISSUE #1999: [liana-business] we should check for same device/xpub set for more than one key
I don't think we currently check for duplicates, i.e. the same device/xpub used for multiple keys in the setup. I think we should disallow duplicates as we do in Liana.

pythcoiner:
closes #1999


edouardparis (approved):
utACK d4d92307b38bdc78c7e5d0fa74bfd66a9f908c3b

PR #2023, opened 2026-02-13
business-installer: switch to loading page on skip registration

### LINKED ISSUE #2002: [liana-business] Unclear text when loading the wallet
When skipping the descriptor registration step, I have this unclear message while the wallet is loading:

Image

I think we should instead have a loading wallet page.

pythcoiner:
closes #2002


edouardparis (approved):
ACK cf0e28bfcb229495fa17557cd7c048d8654ef47c

PR #2025, opened 2026-02-13
business-installer: use new key cards for select key source

### LINKED ISSUE #1979: [liana-business] fingerprint and device name should be switched in Set Keys modal

### LINKED ISSUE #2018: [liana-business] "Paste xpub" option should show a textbox
Currently, the "Paste xpub" option is a button:

https://github.com/user-attachments/assets/e9082e61-d417-4327-a9b5-5964dc7dc845

I think it should show a text box so that the user can see a preview of what he needs to paste and that he is pasting something, similarly to what is done in Liana:

https://github.com/user-attachments/assets/fa3419bf-19c8-4ee2-b6f0-b1aea1107d17

pythcoiner:
this PR do 2 things:
- uses same key card than liana-gui installer rather than legacy ones
- restrict paste xpub to WalletManager roles

image

closes #1979
closes #2018

### DIFF
--- a/liana-business/business-installer/src/state/message.rs
+++ b/liana-business/business-installer/src/state/message.rs
@@ -87,6 +87,7 @@ pub enum Msg {
     XpubRetry,                                                  // Retry fetch after error
     XpubLoadFromFile,                                           // Trigger file picker
     XpubFileLoaded(Result<(String, String), String>),           // (content, filename) or error
+    XpubSelectPaste,                                            // Toggle paste input expanded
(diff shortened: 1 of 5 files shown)


edouardparis (approved):
utACK 8c290b1ab6521bea93fbd8228d47e3987bd9a495

PR #2027, opened 2026-02-16
business-installer: hint that jade do not support taproot miniscript

### LINKED ISSUE #2007: [liana-business] We should not allow to set a key with a device (Jade) / firmware version that doesn't support taproot

pythcoiner:
closes #2007 

image

### DIFF
--- a/liana-business/business-installer/src/views/xpub/modal.rs
+++ b/liana-business/business-installer/src/views/xpub/modal.rs
@@ -11,7 +11,7 @@ use iced::{
 use liana_gui::hw::{is_compatible_with_tapminiscript, min_taproot_version, UnsupportedReason};
 use liana_ui::{
     component::{
-        button, card, form, modal,
+        button, card, form, hw, modal,
         text::{self, p1_bold},
         tooltip,
     },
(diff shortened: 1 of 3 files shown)


liana-business/business-installer/src/views/xpub/modal.rs
```diff
@@ -339,11 +337,20 @@ fn device_card(data: DeviceRenderData) -> Element<'static, Msg> {
                 UnsupportedReason::WrongNetwork => "Wrong network in device settings".to_string(),
                 UnsupportedReason::Version {
                     minimal_supported_version,
-                } => {
-                    format!("Unsupported firmware, upgrade to > {minimal_supported_version}")
-                }
-                UnsupportedReason::Method(m) => format!("Unsupported method: {m}"),
-                UnsupportedReason::AppIsNotOpen => "Please open the app on device".to_string(),
```
edouardparis, line 346: what happened to these two variants ?

pythcoiner: they were accidentaly dropped, they are back now

edouardparis (approved):
ACK f5255ec4d3ac9094dd814818a7de8f1a7c0ecbab

PR #2028, opened 2026-02-16
liana-business: make mainnet default network, signet via ENV var

pythcoiner:
this PR make mainnet the default network
signet can still be used by setting env var `LIANA_BUSINESS_SIGNET` to `1`

Note: `LIANA_BUSINESS_SIGNET_API_URL` & `LIANA_BUSINESS_SIGNET_WS_URL` can be used to route to a custom backend

image

Issue #1960, opened 2026-01-13
Fiat price should be activated by default for Liana Business


nondiremanuel:
In Liana Business binaries, fiat price should be on by default. The price should be fed by our backend as defined in #1959.

This is a nice to have for this cycle, but must have in general (we could do immediately after the cycle).

PR #2030, opened 2026-02-23
Liana Business fiat price

### LINKED ISSUE #1959: Fiat price should be fed by our backend in Liana Business
In Liana Business binaries, we should have the fiat price fed by our own backend in order to avoid breaking changes which would need version upgrade.

This is a nice to have for this cycle, but must have in general (we could do immediately after the cycle).

pythcoiner:
This PR add fiat price integration from liana-backend

fiat price settings are stored on the backend per wallet per user, meaning several users of a same wallet can have differents settings

depends on https://github.com/wizardsardine/liana-backend/pull/361

#2028 must be merged first

#closes #1959 & #1960

### DIFF
--- a/liana-business/business-settings/src/views/mod.rs
+++ b/liana-business/business-settings/src/views/mod.rs
@@ -1,6 +1,6 @@
 //! View functions for business settings UI.
 
-use iced::widget::{scrollable, Column, Container, Row, Space};
+use iced::widget::{pick_list, scrollable, Column, Container, Row, Space, Toggler};
 use iced::{Alignment, Length};
 use liana_ui::{
     component::{badge, button, card, separation, text::*},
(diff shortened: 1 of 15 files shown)


nondiremanuel:
I've tested the PR on our playground env and it looks good. For consistency with Liana, I would keep the general settings on top and with the same icon:

image

image

pythcoiner:
> I would keep the general settings on top and with the same icon:

done

liana-business/business-settings/src/views/mod.rs
```diff
@@ -31,6 +37,7 @@ pub fn list_view() -> Element<'static, Msg> {
         .spacing(20)
         .width(Length::Fill)
         .push(header)
+        .push(general)
```
pythcoiner, line 40: @nondiremanuel here you can see general is first in order

nondiremanuel: I have seen it in my testing. My issue was with the icon.

pythcoiner: ah my bad, icon changed in https://github.com/wizardsardine/liana/compare/b3c14380a9c0dedaeea8b1f4f103c57e1aa21c3f..4d37114b27c2c297c76d5891f173aec76723dc9a

PR #2033, opened 2026-02-25
liana-installer: display pairing code for bitbox when taproot enabled

### LINKED ISSUE #2032: [Liana-Gui] Bug introduced in v13: un-paired Bitbox triggers an error message about Taproot support.
When plugging in a Bitbox02 or Bitbox Nova, that were never paired on Liana before (and only in that case), while selecting the Taproot option in Advanced settings, we get an error message:
"this device do not support taproot"

The device is presenting the pairing code at this step.

Prior versions of Liana, including v12, correctly display the pairing code on the Liana GUI and do not show such an error message. This bug was introduced in v13.

To reproduce, delete the globalsettings.json file in .liana, which contains the pairing data.

Extra point for fixing the error message grammar mistake: should be **does** not support taproot instead of **do** not.

pythcoiner:
closes #2032

### DIFF
--- a/liana-gui/src/installer/step/descriptor/editor/key.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/key.rs
@@ -1090,11 +1090,11 @@ impl SelectKeySource {
         let support_taproot = device.3;
         let mut enabled = true;
         let message = match (state, support_taproot, self.taproot) {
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
utACK 0a1562a4ce9d6120b921c199187d10727a0dc028

PR #2034, opened 2026-02-25
liana-business: handle when wallet backend connection fails on startup

pythcoiner:
image (2)


edouardparis (approved):
utACK bd77c3697db304295a0bf976f4ccee5229087acb

PR #2039, opened 2026-03-04
fix UX: force user to save psbt before exporting it

edouardparis:
To be sure that the user does not export the psbt without saving it which may be the cause of change address reuse. As the derivation index is not incremented in database unless the psbt is saved, it is better to make the user commit to the tx if he is satisfied.


edouardparis:
20260304_10h10m35s_grim
20260304_10h10m47s_grim

PR #2071, opened 2026-04-02
last fixes before release

### LINKED ISSUE #2064: Co-signing and Safety Net key info is not helpful
Right now, for Co-signing and Safety Net keys, in place of the email address - which is not applicable - we place the token:

Image

I don't think this is useful. We should have something else instead. I propose the company name if we have this information.

### LINKED ISSUE #2072: Hide "Key description" field in Liana Business
We don't use the "Key description" field anywhere. 

Image

I think we should hide it from the GUI until (if ever) we use it to avoid useless data entry to the user.

pythcoiner:
this PR:
- reinstate the KeyIdentity variant replaced in previous commit while renaming the newly introduced
- display the token provider name in business-installer when available (close #2064)
- fix a parsing bug in ws_business
- make the ws_business protocol future proof
- use a different pill style for each key type:
image
- remove the description input in the key modal: (close #2072 )
image

- fix setting breadcrumbs colors:
image

### DIFF
--- a/liana-business/business-installer/src/client.rs
+++ b/liana-business/business-installer/src/client.rs
@@ -924,7 +924,7 @@ fn wss_thread(
     tracing::debug!("wss_thread: waiting for Connected response");
     if let Ok(msg) = ws_stream.read() {
         match Response::from_ws_message(msg) {
-            Ok((Response::Connected { user, .. }, _)) => {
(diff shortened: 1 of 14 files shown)


edouardparis (approved):
utACK 21888a06c960634407e71501533cccd0a114c8aa

PR #2085, opened 2026-04-09
Fix reproducible Windows build by stripping COFF symbol table

edouardparis:
The GNU linker appends a COFF symbol table to the PE binary containing auxiliary symbol data (sizes, section offsets) that depends on compilation ordering. Since Rust/LLVM processes codegen units in parallel, this metadata varies between builds even with identical inputs.

The COFF symbol table is not used at runtime (Windows ignores it) and modern debugging uses separate PDB files instead. Stripping it with --strip-all removes the non-deterministic data and also reduces binary size significantly

this is a backport of #2084

### DIFF
--- a/nix/liana-business.nix
+++ b/nix/liana-business.nix
@@ -28,7 +28,7 @@ let
 
     SOURCE_DATE_EPOCH = 1;
     CARGO_BUILD_TARGET = "x86_64-pc-windows-gnu";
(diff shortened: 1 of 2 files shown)


edouardparis (commented):
Self-ACK cf7a17ea44b25b678df3993e8d1cd181c1a85ab1
backport of 2080

PR #2096, opened 2026-04-18
Add CI job to test each commit of a PR

pythcoiner:
This PR adds a new CI job that will run fmt + clippy + build + units tests for each commits of the PR, each in a different task (so the total CI time is unchanged).
integration tests are still only run at the HEAD.

tested with a broken commit at HEAD - 1:
image


edouardparis (approved):
ACK de0e422db108379a3cc5ddb24f6c52246c03a29f

PR #2097, opened 2026-04-21
fix release: correct binary naming

edouardparis:
Find latest matching tag by version sorting (works even if tag is not an ancestor of HEAD).
Use versionsort.suffix to ensure release candidates sort before their final release
(e.g., v14.0 > v14.0rc1 > v13.1).


pythcoiner:
@edouardparis is this still relevant?

PR #2086, opened 2026-04-10
Bump iced v14.0

edouardparis:

PR #2098, opened 2026-04-22
Rework sidebar responsive mechanism

pythcoiner:
This is preparatory work to try fixing an issue that arise while bumping iced version, see [this comment](https://github.com/wizardsardine/liana/pull/2086#issuecomment-4289945351).

Before we where draing the view from the closure passed to `responsive()`, now we use only `responsive()` as a "probe" to populate `Cache` with the pane size, and draw the view independently.


edouardparis (approved):
utACK 91859672c4c09dce03aaafa6fc8e504cfe4a6197

PR #2111, opened 2026-04-30
Remove address index from QR Code

### LINKED ISSUE #1962: Remove address index when scanning QR code (for privacy)

pythcoiner:
This PR remove the address index in the QR Code share to a sender, and adds a button to show the in dex in a QR Code in the verify address modal, it's still needed for Specter DIY devices as there is no implemented way to trigger an address verification via USB:

image

it's an alternative to #1965 

closes #1962


pythcoiner:
@nondiremanuel I added "or display a QR Code (Specter DIY)" to the label, feel free if you prefer something else and/or if you prefer the button to at a different position

nondiremanuel:
Do you have a screencast example of how it would work with "Specter DIY"? Just to be sure I'm getting it correctly.

pythcoiner:
> Do you have a screencast example of how it would work with "Specter DIY"? Just to be sure I'm getting it correctly.

my specter is broken, but the flow is:
- User click on Show QR
- Scan with specter
- Specter show the address at this index if it match a loaded descriptor

nondiremanuel:
I would personally do the following to make it clearer:

- Avoid adding the "or display a QR Code (Specter DIY)" part in the text, which makes it more ambiguous for users who don't have Specter DIY (the vast majority, I guess).
- Add a "Plug in a hardware device ..." icon + text when no hardware is detected (as in the set keys modal):
image
This is good for consistency and UI in general

- Move the "Show QR Code" button below that (or below the list of devices)
- Add an info tip to the "Show QR Code" button saying on hover something like. "For airgap devices such as Specter DIY", or just "For Specter DIY devices"

WDYT?

pythcoiner:
lgtm

pythcoiner:
image

on hover tooltip:
image

image

@nondiremanuel

nondiremanuel:
The "Show QR Code" option is too big IMHO. It shouldn't have the same priority as selecting a device since it's much less likely to be useful, especially if a user has a device plugged in. I would keep it as a small button on the bottom right.

pythcoiner:
well, I'd argue with that, it's just "another" type of signing device if we want to hide it I think we should use an "other options" section, but keep the same type of button than for other devices:

image

pythcoiner:
added an optional section:

image

image

PR #2123, opened 2026-05-11
UI Rework

### LINKED ISSUE #2104: UX/UI rework: buttons
Following #2099 discussion, I think we should take the opportunity of the UX/UI rework to align the buttons design system to what we have implemented for Liana Business. Working with Claude, I've come up with the following proposal:

[liana_buttons_final.html](https://github.com/user-attachments/files/27002331/liana_buttons_final.html)

here is the table of colors used:

[liana_button_color_codes.html](https://github.com/user-attachments/files/27002344/liana_button_color_codes.html)

Let's keep discussing in case there is something which could be improved

### LINKED ISSUE #2113: UI/UX Rework: Pills
# Liana Business

## Home

> Change "Unconfirmed" badge (amber filling, black text)

> Adding a "Confirmed" badge -> I think it's not strictly required. It seems clear enough that what's not "Unconfirmed" is confirmed IMO

Image

---

`Unconfirmed`:

Image

Image

---

`Confirmed`:

Image

Image

---

## Drafts & Approvals:

> Add "Unsigned" and "To broadcast" badges, both with amber border and text (I don't like the figma green "Signed" as explained in my last [comment](https://www.figma.com/design/zNR7J87fsvqrSnPJhQatRx?node-id=2612-2780#1685032167))

Image

---

`Unsigned`:

Image

Image

---

`To broadcast`:

Image

Image

---

> Add a green "Confirmed" badge (and remove the "Spent" one. As a side note, I personally like more the "Spent" wording so we could evaluate to keep it).

idem home

---

> Move more to the right and change formatting of "Batch" badge. In the figma it wasn't properly defined, but I would use the same position and format for the "Recovery" and "Deprecated" badges (I don't know if I miss any).

`Batch`:

Image

---

`Fingerprint`:

Image

Image

32 pix height
---

# Liana GUI

Image

### LINKED ISSUE #2121: Business-installer cleanup
- [x] do not populate with dummy email
- [x] Scroll bar overly onto buttons
- [x] `Disconnect misbehavior`
- [x] `Connect with another email` must be disabled during an account loading
- [x] Replace "Search organizations..." by "Filter organizations..."
- [x] Replace "Search wallets..." by "Filter wallets..."

---

Do not populate with dummy email:

Image

---

Scroll bar overlay onto buttons:

Image

---

`Disconnect misbehavior`:

Image

  ONLY AT FIRST CONNECTION, Clicking on disconnect actualy route here:

Image

  while it should route here:

Image

the issue seems to be we do not cache the newly created account, just write to disk

---

`Connect with another email` must be disabled during an account loading:

Image

--- 

Replace "Search organizations..." by "Filter organizations..."
or smth similar, and maybe display onli if > 4/5 orgs

Image

---

Replace "Search wallets..." by "Filter wallets..."
or smth similar, and maybe display onli if > 4/5 wallets

Image
---

### LINKED ISSUE #2125: UI/UX Rework: Button with tooltip component
from https://github.com/wizardsardine/liana/pull/2123#issuecomment-4421821217:

> Some buttons have different sizes (possibly due to font size)

it's because the button is constructed inline in the view, we must have a proper component

Image

### LINKED ISSUE #2126: The text doesn't wrap/truncate
On master, there are multiple occasions in which the text doesn't wrap or truncate. Some examples from the Liana Business installer:

Image

Image

This is possibly due to the recent iced version bump

### LINKED ISSUE #2128: User should be able to see entire path in the Approve step of Liana Business wallet setup
In order to be able to approve, the Wallet Manager should be able to see the policy. Currently, it sometimes just shows the first key(s) and the timelock indication is missing too:

Image

We should:
- Improve visibility at first glance. Relatively simple (<= 3 keys) paths should be understandable without needing to expand (by reducing the font size?) and timelocks should be shown below (as they were IIRC) for recovery paths.
- Give the option to the user to expand each path to see the full info (through a modal?).

Alternatively, we could re-implement the step to be more similar to what we have in the admin page:

Image

Note that if we go this route, we should have the timelock based on time instead of blocks (that could still be in parenthesis)

### LINKED ISSUE #2132: Bad error messages when the device locks itself
We've experienced this with a ledger device, for which it's pretty common that it locks itself, but I'm not sure it applies also to other devices as well.

When trying to perform an action on the device while in the meantime it locked itself (for some idle time has passed), we get error "messages" of this kind:

Image

We should have a proper message such as "Device is unreachable. Please make sure it's correctly plugged and unlocked" or something like this.

### LINKED ISSUE #2136: UI/UX Rework: Make hw clickable card consistants with a new component
Hws clickable card are contructed inline with a combo card + button, we must implem a new component to enforce consistency

Image

Image

Image

Image

Image

Image

Image

Image

Image

Image

Image

### LINKED ISSUE #2137: We should avoid showing the "+test" in email addresses of the "Edited by" field in Liana Business
We currently show the full name+admin email address in the "Edited by" field. 

Image

I propose one of the following:
- Show a generic "Edited by Admin"
- Show the email without "+test" (e.g. manuel@wizardsardine.com in place of manuel+admin@wizardsardine.com)

### LINKED ISSUE #2138: Edited by "Unknown" in organization screen
We currently always show "Edited by Unknown" at the organization level step:

Image

We should show the correct information (following the same consideration done in #2137 for admins).

### LINKED ISSUE #2142: "Settings" moves when going to a sub-page in Liana Business
The position of the "Settings" title slightly moves going from the main page to the sub-pages (I think it was already fixed for liana gui):

https://github.com/user-attachments/assets/5868b3e2-1f4e-4b42-bd03-ed6667128259

### LINKED ISSUE #2145: Fix scrollable
Scrollable are broken in iced v14, scrollbar overlap content:

Image

we must implement our own companent that will add padding in order to "push" the scrollbar outside the content

pythcoiner:
Todo:

 - [x] Fix button color in liana-gui
 - [x] Reset timelock button primary
 - [ ] ~~Align pills in coin selection~~
 - [x] Hover text should be "Recovery option(s) available"
 - [x] Some buttons have different sizes (possibly due to font size) tracked in #2125 
 
 closes #2132 
 closes #2125
 closes #2126
 closes #2136
 closes #2104
 closes #2121
 closes #2145
 closes #2142
 closes #2128
 closes #2113
 closes #2138
 closes #2137


nondiremanuel:
Some comments from a first quick look (if some are out of scope for the PR and were going to be tackled anyway, please ignore them):

image

As we were discussing elsewhere, the green for the primary buttons and the borders and text of the secondary ones should be brighter (same as the current green and the one on the menu bar).

image

Shouldn't the "Reset timelock" button be primary as it was before? I see that in the figma of Liana Business it was changed to tertiary (and I'm sorry I didn't notice it in advance), but I personally think it should be the primary action.

image

- It would be nice to have the "Available" pills aligned
- Hover text should be "Recovery option(s) available" IMO

Same for the "Coins/UTXOs" tab

image

Shouldn't the "copy address" icon be a secondary button too?

image

Some buttons have different sizes (possibly due to font size)

pythcoiner:
> Shouldn't the "copy address" icon be a secondary button too?

always has been a transparent button afaik (border became green on hover)

related:
image

pythcoiner:
> * Hover text should be "Recovery option(s) available" IMO

done

image

pythcoiner:
> Shouldn't the "Reset timelock" button be primary as it was before? I see that in the figma of Liana Business it was changed to tertiary (and I'm sorry I didn't notice it in advance), but I personally think it should be the primary action.

done

pythcoiner:
> It would be nice to have the "Available" pills aligned

I'll not do this here,as we have to rework the spend panel anyway

edouardparis (approved):
utACK 98c72109be7eca306e0c22e2aff8959654e8c20a

Issue #672, opened 2023-08-31
Allow to start without a Bitcoin backend (ie at the moment, without `bitcoind`)


darosior:
It's been requested a couple times, and it makes sense for instance for a passive co-signatory in a multisig. The one crafting the transaction would use a regular Liana with a bitcoind, the cosigner doesn't need to be syncing bitcoind.

I expect it's non-trivial as there are assumptions that a Bitcoin backend is running all around the lianad codebase.

PR #2146, opened 2026-05-18
Make GUI updates non-blocking and avoid redraw artifacts

starius:
## Summary

- Run embedded daemon control calls on Tokio's blocking pool in `liana-gui/src/daemon/embedded.rs`, so synchronous lianad work no longer occupies async executor workers directly.
- Replace GUI update-path `Handle::block_on` calls with `Task::perform` or background runtime tasks across `liana-gui/src/app/mod.rs`, `liana-gui/src/app/state/spend/step.rs`, `liana-gui/src/installer/mod.rs`, `liana-gui/src/launcher.rs`, `liana-gui/src/gui/tab.rs`, and `liana-gui/src/loader.rs`.
- Add explicit async result messages in `liana-gui/src/app/message.rs` and `liana-gui/src/installer/message.rs` so preselection, daemon config reloads, spend redrafts, failed-install cleanup, remote alias sync, and wallet deletion complete without blocking GUI message handling.
- Remove the blocking sleep in loader sync in `liana-gui/src/loader.rs`.
- Avoid cursor-blink redraw artifacts in form cards by removing the semi-transparent shadow from `card::simple` in `liana-ui/src/theme/card.rs`.

## Rationale

`liana-gui` is async, but embedded `lianad` remains synchronous internally. Before this change, several GUI-facing async methods or update paths still crossed that boundary without an explicit blocking boundary, or synchronously waited for daemon/database/network work from GUI update handling. This could make navigation, spend editing, settings changes, installer cleanup, wallet deletion, startup, and close handling sensitive to daemon latency.

This keeps `lianad` synchronous and limits the change to the GUI boundary: synchronous embedded daemon calls are isolated on the blocking pool, and GUI state transitions schedule work as iced tasks instead of waiting inline.

### DIFF
--- a/liana-gui/src/app/message.rs
+++ b/liana-gui/src/app/message.rs
@@ -30,6 +30,7 @@ use crate::{
 pub enum Message {
     Tick,
     RedirectLianaConnectLogin,
+    RemoteBackendAliasUpdated(Result<(), String>),
     UpdateDaemonCache(Result<DaemonCache, Error>),
     CacheUpdated,
     Fiat(FiatMessage),
--- a/liana-gui/src/app/message.rs
+++ b/liana-gui/src/app/message.rs
@@ -60,6 +60,8 @@ pub enum Message {
     HardwareWallets(HardwareWalletMessage),
     HistoryTransactionsExtension(Result<Vec<HistoryTransaction>, Error>),
     HistoryTransactions(Result<Vec<HistoryTransaction>, Error>),
+    PreselectedHistoryTransaction(Txid, Result<Option<HistoryTransaction>, Error>),
+    PreselectedSpendTransaction(Txid, Result<Option<SpendTx>, Error>),
     Payments(Result<Vec<Payment>, Error>),
     PaymentsExtension(Result<Vec<Payment>, Error>),
     Payment(Result<(HistoryTransaction, usize), Error>),
--- a/liana-gui/src/app/message.rs
+++ b/liana-gui/src/app/message.rs
@@ -16,6 +16,7 @@ use crate::{
         wallet::Wallet,
     },
     daemon::model::*,
+    daemon::Daemon,
     export::ImportExportMessage,
     hw::HardwareWalletMessage,
     services::fiat::{
@@ -34,6 +35,7 @@ pub enum Message {
     UpdatePanelCache(/* is current panel */ bool),
     View(view::Message),
     LoadDaemonConfig(Box<DaemonConfig>),
+    DaemonConfigReloaded(Result<Arc<dyn Daemon + Sync + Send>, Error>),
     DaemonConfigLoaded(Result<(), Error>),
     LoadWallet(Wallet),
     Info(Result<GetInfoResult, Error>),
--- a/liana-gui/src/app/message.rs
+++ b/liana-gui/src/app/message.rs
@@ -2,9 +2,10 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 use liana::miniscript::bitcoin::{
+    address,
     bip32::{ChildNumber, Fingerprint},
     psbt::Psbt,
-    Address, Txid,
+    Address, Amount, Txid,
 };
 use lianad::config::Config as DaemonConfig;
 
@@ -51,6 +52,7 @@ pub enum Message {
     Labels(Result<HashMap<String, String>, Error>),
     SpendTxs(Result<Vec<SpendTx>, Error>),
     Psbt(Result<(Psbt, Vec<String>), Error>),
+    RedraftSpend(u64, RedraftSpendResult),
     RbfPsbt(Result<Txid, Error>),
     Recovery(Result<SpendTx, Error>),
     Signed(Fingerprint, Result<Psbt, Error>),
@@ -71,6 +73,17 @@ pub enum Message {
     BroadcastModal(Result<HashSet<Txid>, Error>),
     RbfModal(Box<HistoryTransaction>, bool, Result<HashSet<Txid>, Error>),
     Export(ImportExportMessage),
+}
+
+#[derive(Debug)]
+pub struct RedraftSpendResult {
+    pub send_max_to_recipient: Option<usize>,
+    pub total_recipients: usize,
+    pub max_address: Address<address::NetworkUnchecked>,
+    pub is_user_coin_selection: bool,
+    pub is_recovery: bool,
+    pub result: Result<CreateSpendResult, Error>,
+    pub amount_left_to_select: Option<Amount>,
 }
 
 impl From<ImportExportMessage> for Message {
--- a/liana-gui/src/app/mod.rs
+++ b/liana-gui/src/app/mod.rs
@@ -247,31 +247,37 @@ impl<S: SettingsTrait> App<S> {
 
         match &menu {
             menu::Menu::TransactionPreSelected(txid) => {
-                if let Ok(Some(tx)) = Handle::current().block_on(async {
-                    self.daemon
-                        .get_history_txs(&[*txid])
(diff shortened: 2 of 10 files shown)


pythcoiner:
concept NACK, the rationale seems AI bullshit to me, and I dont see what problem it solves

starius:
The concrete issue I see is that, when lianad is embedded in the GUI, the GUI exposes async daemon calls but `EmbeddedDaemon::command` [runs](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/daemon/embedded.rs#L33-L39) the synchronous `DaemonControl` method inline while holding the async mutex, with no blocking boundary. If any of those daemon paths block on backend IO, the GUI task can stall. I observed GUI hanging in some of my experiments and came to the conclusion that this is the root cause.

Examples:

- Node -> Blockchain rescan: the GUI [calls](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/app/state/settings/bitcoind.rs#L636-L644) `start_rescan`, embedded lianad [maps](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/daemon/embedded.rs#L224-L230) it directly to `DaemonControl::start_rescan`, and the daemon [checks](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/commands/mod.rs#L1190-L1211) backend state before calling backend `start_rescan`. With bitcoind this [reaches](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/bitcoin/d/mod.rs#L1129-L1162) `importdescriptors`, prune checks, retries, and `std::thread::sleep`.
- Broadcast: the GUI [calls](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/app/state/psbt.rs#L370-L381) `broadcast_spend_tx`, embedded lianad [maps](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/daemon/embedded.rs#L215-L221) it directly to `broadcast_spend`, and the daemon [broadcasts](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/commands/mod.rs#L910-L927) through the backend before waiting for the poller with `rx.recv()`. The backend call is bitcoind [`sendrawtransaction`](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/bitcoin/d/mod.rs#L1047-L1053) or Electrum [`transaction_broadcast`](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/bitcoin/electrum/client.rs#L108-L110).
- Node config reload: saving node settings [emits](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/app/state/settings/bitcoind.rs#L364-L374) `LoadDaemonConfig` for bitcoind or [does](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/app/state/settings/bitcoind.rs#L473-L484) the same for Electrum. The app [handles](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/liana-gui/src/app/mod.rs#L544-L579) it synchronously by `block_on`-stopping the daemon and calling `EmbeddedDaemon::start` inline. Startup [sets up](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/lib.rs#L406-L439) bitcoind/Electrum backends; Electrum startup [pings](https://github.com/wizardsardine/liana/blob/de4f0b90feb7875ab7811981a2e1a315a6863329/lianad/src/bitcoin/electrum/client.rs#L57-L79) the server and creates a retrying client.

So the PR is a boundary fix: keep lianad synchronous, but move these potentially blocking embedded-daemon operations off the GUI async/update path.

pythcoiner:
> I observed GUI hanging in some of my experiments and came to the conclusion that this is the root cause.

After hundread hours stress-testing the GUI, I never experienced this, we need a reproducible user-facing problem before a fix. Fixing potential bugs is a good way to introduce REAL bugs.

starius:
@pythcoiner You are right, changes like this should be handled carefully and only when the bug is proven. I think my PR's scope is too large, but some fixes are useful.

I reproduced the following scenario. When bitcoind stops, embedded lianad enters its synchronous bitcoind retry loop. It logs `Transient error when sending request to bitcoind` once per second while sleeping and retrying. Then the user closes the GUI. The GUI close handler calls embedded daemon shutdown synchronously from the GUI path. That shutdown sends `Shutdown` to the lianad poller thread and then waits for the poller thread to join. But the poller thread is currently stuck inside the blocking bitcoind retry loop, so it cannot process the shutdown message until the retry loop finishes. As a result, the GUI thread is waiting for lianad to stop, lianad is waiting/retrying bitcoind, and the window stops processing events. The window manager marks it as not responding, while logs keep printing because the daemon retry code is still running.

I also added a repro script for this scenario, and I'm attaching both the script and the video:

https://github.com/user-attachments/assets/8ea074f9-e006-4fa3-92fb-e5a82837edc0

The script: [reproduce-gui-close-freeze.sh](https://github.com/user-attachments/files/28003696/reproduce-gui-close-freeze.sh)


PS. The video was recorded by AI, the scenario was found by AI. I think this is an interesting way to find GUI glitches: let AI run the app in a virtual X server and look for bugs. It seems better to burn AI time than human time on this kind of issue hunting.

pythcoiner:
> I reproduced the following scenario. When bitcoind stops, embedded lianad enters its synchronous bitcoind retry loop.

Well my understanding is that, the "bug" is an edge case, that cant happend during a normal usage

> It seems better to burn AI time than human time on this kind of issue hunting.

I totaly disagree on this when it's about, easy to find, UI glitches, that do not impact users in real world. Our bottleneck is litteraly human time, and honnestly, this burn himan time.

> The script: [reproduce-gui-close-freeze.sh](https://github.com/user-attachments/files/28003696/reproduce-gui-close-freeze.sh)

PS: please if you want to share scripts, just inline them in the discussion.

starius:
I think it is still a normal situation: stop bitcoind / electrs and then try to close liana-gui and it gets stuck and becomes not responding. IMHO the expected behavior is a normal shutdown in this situation.

Maybe I'm over-engineering here. I tried to add another backend (BIP157) and got GUI stuck in many situations. I propose to close this PR, unless you find some pieces are useful.

For example, this one looks useful for me: https://github.com/wizardsardine/liana/pull/2146/changes/99e22d4276b627a88d5b0c50e37690fc6c7c47c3 (`std::thread::sleep` -> `tokio::time::sleep` in GUI code). What do you think about it?

pythcoiner:
> I think it is still a normal situation: stop bitcoind / electrs and then try to close liana-gui

No, its absolutely not a normal situation, it's an edge case:
- an electrum server is expected be be always running
- a managed bitcoind must NOT be stopped by the user if it has been started by Liana

I agree we should handle this more gracefully, but the solution I'd go is something like https://github.com/wizardsardine/liana/issues/672.

>  I tried to add another backend (BIP157)

Thanks for looking at this, but I think (in respect of your time), the approach should be discussed in an issue before spending much time on works that have large scope like this: there are many undocumented details that can make such integration hard to get merged.

PR #2148, opened 2026-05-22
hwi: fix bug w/ ledger devices always locked

### LINKED ISSUE #2131: Issue with device locking state
We've experienced this with a ledger device, but I'm not sure it applies also to other devices. When setting a key, at first the device appears correctly:

Image

If I quit the bitcoin (testnet) app on the device or if the device locks itself after some time, I get the following state:

Image

(for the visibility issue see #2130).

The issue is that if I unlock the device and/or re-open the bitcoin (testnet) app, the state on the GUI doesn't change and I'm not able to select the device.

This is specific to Liana Business since in Liana GUI the device can be selected again.

pythcoiner:
closes #2131

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -287,9 +287,9 @@ dependencies = [
 
 [[package]]
 name = "async-hwi"
-version = "0.0.31"
+version = "0.0.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


pythcoiner:
for reproduce the bug, before this in liana-business:
- BEFORE starting liana-business, unlock your ledger devce but do NOT open the bitcoin app 
- go to a signing device modal
- open the ledger bitcoin app
- your device button keep disabled

edouardparis (approved):
ACK 70567a440fbb64810366607841f33ee6149d3404

PR #2156, opened 2026-05-29
Fix button size in business installer

pythcoiner:
This PR move all definitions of buttons using `BtnWidth` in liana-ui and fix BtnWidth values to fit the new button font


edouardparis (approved):
ACK 698f3c528beb3b49b0a0371839a33935b43a47f9

PR #2164, opened 2026-06-05
gui: disable policy edits while compiling

### LINKED ISSUE #2163: Stale async descriptor compile can advance installer after reset/template change
it looks to me we can hit a race in the installer descriptor editor.

When clicking `Continue`, we take `pending_compile` and spawn an async compile:

- `liana-gui/src/installer/mod.rs:292-312`

but when the compile result comes back, the editor always stores it and sends
`Message::Next`:

```rust
self.compiled = Some(*desc);
return Task::done(Message::Next);
```

- `liana-gui/src/installer/step/descriptor/editor/mod.rs:441-447`

So if the user reset/change template while the compile is still in flight, the
stale compile can still advance the installer with a descriptor built from the
previous inputs.

The reset/customize buttons can change the descriptor state from the same view:

- `liana-gui/src/installer/view/editor/template/mod.rs:53-70`

I have not reproduced it yet, maybe it's hard to hit in practice, but we should
probably invalidate/ignore compile results when the template/editor state changed,
so only the latest compile can send `Message::Next`.

ws-plaude:
The descriptor editor can accept policy-edit actions while compilation is pending.

Fixes #2163

### DIFF
--- a/liana-gui/src/installer/view/editor/mod.rs
+++ b/liana-gui/src/installer/view/editor/mod.rs
@@ -45,22 +45,31 @@ impl std::fmt::Display for DescriptorKind {
 }
 
 #[allow(clippy::too_many_arguments)]
-pub fn define_descriptor_advanced_settings<'a>(use_taproot: bool) -> Element<'a, Message> {
+pub fn define_descriptor_advanced_settings<'a>(
+    use_taproot: bool,
+    editable: bool,
+) -> Element<'a, Message> {
+    let descriptor_kind = if use_taproot {
+        DescriptorKind::Taproot
+    } else {
+        DescriptorKind::P2WSH
+    };
+    let descriptor_type: Element<_> = if editable {
+        pick_list::pick_list(&DESCRIPTOR_KINDS[..], Some(descriptor_kind), move |kind| {
+            Message::CreateTaprootDescriptor(kind == DescriptorKind::Taproot)
+        })
+        .padding(10)
+        .into()
+    } else {
+        card::simple(text(descriptor_kind.to_string()))
+            .padding(10)
+            .into()
+    };
+
     let col_wallet = Column::new()
         .spacing(10)
         .push(text("Descriptor type").bold())
-        .push(container(
-            pick_list::pick_list(
-                &DESCRIPTOR_KINDS[..],
-                Some(if use_taproot {
-                    DescriptorKind::Taproot
-                } else {
-                    DescriptorKind::P2WSH
-                }),
-                |kind| Message::CreateTaprootDescriptor(kind == DescriptorKind::Taproot),
-            )
-            .padding(10),
-        ));
+        .push(container(descriptor_type));
 
     container(
         Column::new()
@@ -88,13 +97,14 @@ pub fn path(
     threshold: usize,
     keys: Vec<Element<message::DefinePath>>,
     fixed: bool,
+    editable: bool,
 ) -> Element<message::DefinePath> {
     let keys_len = keys.len();
     Container::new(
         Column::new()
             .spacing(10)
             .push_maybe(title.map(|t| Row::new().push(Space::with_width(10)).push(p1_bold(t))))
-            .push(defined_sequence(sequence, warning))
+            .push(defined_sequence(sequence, warning, editable))
             .push(
                 Column::new()
                     .spacing(5)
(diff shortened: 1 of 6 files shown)


pythcoiner:
@ws-plaude it's a good start but not sufficient, [Add a recovery option] and [Add a safety net] must also been greyed out

ws-plaude:
Fixed in 141971b3ba3874d638d395799eada30cf3312a56. The custom template now disables both `Add recovery option` and `Add Safety Net` while descriptor compilation is processing.\n\nVerification:\n- `cargo fmt --check`\n- `cargo check -p liana-gui --lib`\n- `cargo test -p liana-gui installer::step::descriptor::editor::tests::test_define_descriptor_ignores_policy_edits_while_processing --lib`\n\nGitHub reports no checks for the branch yet.

pythcoiner:
@ws-plaude can you squash your commits?

ws-plaude:
Done: squashed the branch down to one commit (`3c326257b1bd3f3b5132e4576b5c79f4a836eb7d`) and force-pushed it.

Verification: confirmed the squashed commit has the same tree as the previous PR head and `git diff --check origin/master..HEAD` passes.

liana-gui/src/installer/step/descriptor/editor/mod.rs
```diff
@@ -234,6 +234,22 @@ impl Step for DefineDescriptor {
...
+                    | Message::DefineDescriptor(message::DefineDescriptor::AddSafetyNetPath)
+                    | Message::DefineDescriptor(message::DefineDescriptor::KeysEdited(_, _))
+                    | Message::DefineDescriptor(message::DefineDescriptor::KeysEdit(_, _))
+                    | Message::DefineDescriptor(message::DefineDescriptor::AliasEdited(_, _))
+                    | Message::DefineDescriptor(message::DefineDescriptor::Reset)
+                    | Message::DefineDescriptor(message::DefineDescriptor::Path(_, _))
+            )
+        {
+            return Task::none();
+        }
```
pythcoiner: @ws-plaude I dont think this is necessary, while this is correct we generally consider that if a message arrives to update() it must be handled

ws-plaude: Done in 83f83c5abd73f46699af6f731d5ba19175e9e198: dropped the update-level guard and the corresponding test. The PR now only disables the UI controls while processing, so any message that reaches update() is still handled.

Verification:
- `cargo fmt --check`
- `cargo check -p liana-gui --lib`
- `cargo test -p liana-gui installer::step::descriptor::editor::tests::test_define_descriptor_use_hotkey --lib`
- `git diff --check origin/master...HEAD`

GitHub reports no checks for the branch yet.

pythcoiner:
@ws-plaude you breaks something, none of the buttons are disabled during processing

ws-plaude:
Fixed in 18d6feab71ea2551091444e8d48d9b8bf8e78c7c. I kept the branch squashed to one commit.

The view still greys out the editor buttons, and the step now also ignores policy-edit messages while descriptor compilation is processing so stale clicks cannot mutate the policy.

Verification:
- `cargo test -p liana-gui installer::step::descriptor::editor::tests::test_define_descriptor_ignores_policy_edits_while_processing --lib`
- `cargo fmt --check`
- `cargo check -p liana-gui --lib`

GitHub reports no checks for the branch yet.

liana-gui/src/installer/step/descriptor/editor/mod.rs
```diff
@@ -234,6 +234,14 @@ impl Step for DefineDescriptor {
...
     fn update(&mut self, hws: &mut HardwareWallets, message: Message) -> Task<Message> {
         self.error = None;
+        if self.processing
+            && !matches!(
+                message,
+                Message::DefineDescriptor(message::DefineDescriptor::Compiled(_))
+            )
+        {
+            return Task::none();
+        }
```
pythcoiner: @ws-plaude is this strictly necessary or we could drop it?

ws-plaude: done

nondiremanuel:
I think we should have a better loading state. The "Processing..." alone is not very visible. Could you try adding a modal with a load spinner that disables any edits while compiling? Do this in a different PR please @ws-plaude 
Please ask if anything is unclear or needs to be decided

ws-plaude:
done: #2165

pythcoiner:
I'm closing this as it's superseded by #2165

PR #2165, opened 2026-06-05
gui: show policy compile modal

ws-plaude:
Adds a modal over the policy editor while descriptor compilation is running.

The spinner and modal live in liana-ui so the component can be reused.

### DIFF
--- a/liana-business/src/main.rs
+++ b/liana-business/src/main.rs
@@ -79,6 +79,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     .scale_factor(LianaBusiness::scale_factor)
     .subscription(LianaBusiness::subscription)
     .settings(settings)
+    .antialiasing(true)
     .window(window_settings)
     .run()
     {
--- a/liana-gui/src/installer/step/descriptor/editor/mod.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/mod.rs
@@ -16,7 +16,7 @@ use liana::{
 };
 
 use liana_ui::{
-    component::form,
+    component::{form, spinner},
     widget::{modal::Modal, Element},
 };
 
@@ -642,7 +642,15 @@ impl Step for DefineDescriptor {
             ),
         };
         if self.processing {
-            Modal::new(content, view::editor::processing_modal()).into()
+            Modal::new(
+                content,
+                spinner::spinner_modal(
+                    "Compiling descriptor",
+                    "Generating and validating your wallet descriptor. This can take a few seconds.",
+                ),
+            )
+            .backdrop(liana_ui::color::BLACK_60)
+            .into()
         } else if let Some(modal) = &self.modal {
             Modal::new(content, modal.view(hws))
                 .on_blur(if modal.processing() {
--- a/liana-gui/src/installer/step/descriptor/editor/mod.rs
+++ b/liana-gui/src/installer/step/descriptor/editor/mod.rs
@@ -641,27 +641,36 @@ impl Step for DefineDescriptor {
                 self.processing,
             ),
         };
-        if self.processing {
-            Modal::new(
-                content,
+        let modal: Option<Element<'a, Message>> = if self.processing {
+            Some(
                 spinner::spinner_modal(
                     "Compiling descriptor",
                     "Generating and validating your wallet descriptor. This can take a few seconds.",
-                ),
+                )
+                .into(),
             )
-            .backdrop(liana_ui::color::BLACK_60)
-            .into()
-        } else if let Some(modal) = &self.modal {
-            Modal::new(content, modal.view(hws))
-                .on_blur(if modal.processing() {
-                    None
-                } else {
-                    Some(Message::Close)
-                })
-                .into()
         } else {
-            content
-        }
+            self.modal.as_ref().map(|modal| modal.view(hws))
+        };
+
+        let on_blur = self.modal.as_ref().and_then(|modal| {
+            if modal.processing() {
+                None
+            } else {
+                Some(Message::Close)
+            }
+        });
+
+        let backdrop = if self.processing {
+            liana_ui::color::BLACK_60
+        } else {
+            liana_ui::color::BLACK_80
+        };
+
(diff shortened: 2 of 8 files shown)


pythcoiner:
here the result build with the very basic spinner we have:

image

 I'll try to make something more nice

pythcoiner:
@ws-plaude could you make the spinner looks better, for instance something like this:

```
# "Compiling descriptor" modal — implementation spec

Implementer-ready spec for the descriptor-compile modal shown in the Liana wallet
installer. Geometry and timing are exact; framed for the iced/Rust GUI but portable.
**Replace the existing modal's spinner entirely with the two-ring construction below.**

---

## Behavior / trigger logic

- **Trigger:** shown when the user clicks **Continue** on the *Set up your keys*
  step, while the wallet descriptor is being compiled/validated.
- **Non-dismissible:** no close button, no click-outside-to-dismiss, no Esc. It is
  purely a busy/progress indicator. The only way out is the compile finishing.
- **Lifecycle:** show on compile-start → keep showing while the compile runs →
  dismiss automatically the instant the compile completes, then advance to the next
  step. Bind to the actual async compile result, **not** a timer. (The HTML mock
  uses a fixed 2.2s placeholder only because it has no real compile.)
- **Indeterminate:** the spinner shows activity only — it does **not** represent
  percentage progress.

---

## Layout

- **Backdrop:** covers the whole window, fill `rgba(0, 0, 0, 0.6)`. Content centered
  on both axes. Captures/blocks all input to the screen behind it.
- **Card:**
  - background `#141414`
  - border `1px solid #3F3F3F`
  - corner radius `16px`
  - padding `40px` vertical, `36px` horizontal
  - width `380px` (cap to ~90% of window width on small screens)
  - vertical stack, center-aligned, `22px` gap between the spinner and the text block.
- **Text block** (centered, `8px` gap between title and body):
  - **Title:** "Compiling descriptor" — IBM Plex Sans, weight 700, 24px, color `#FFFFFF`.
  - **Body:** "Generating and validating your wallet descriptor. This can take a few
    seconds." — body style, color `#717171`.

---

## The spinner (the important part)

Two concentric rings: a static dim **track** + one **arc** of fixed angular length
that **rotates continuously**. Nothing scales, fades, or changes length — only the
arc's rotation angle changes over time.

### Exact geometry
- Bounding box: **56 × 56 px**.
- Stroke width (both rings): **4 px**.
- Radius = `(56 − 4) / 2 = 26 px`, centered at `(28, 28)`.
- **Track ring:** a full 360° circle stroked in `#272727`, width 4. Always static.
- **Arc ring:** an arc of **90° (π/2 radians)** stroked in `#00FF66` (Liana green),
  width 4, drawn on top of the track. Butt caps match the CSS reference; round is
  also fine.

### Animation
- The arc's **start angle increases linearly with time**, completing a full
  revolution every **0.9 seconds** (≈ 400°/s), looping forever, **linear** easing
  (constant angular velocity — no ease-in/out).
- Formula: `start_angle = (elapsed_seconds / 0.9) * 2π  (mod 2π)`; the arc spans
  `[start_angle, start_angle + π/2]`.

### iced implementation note
Use a `Canvas` widget driven by an animation subscription (e.g. `window::frames()`
or a `time::every` tick) that stores `elapsed` and triggers a redraw:

1. Stroke the full circle (track) in `#272727`, width `4.0`.
2. Build a `Path` with
   `path.arc(Arc { center, radius: 26.0, start_angle, end_angle: start_angle + FRAC_PI_2 })`
   and stroke it in `#00FF66`, width `4.0`.
3. Each frame, recompute `start_angle` from accumulated elapsed time and request a
   redraw.

No gradients, shadows, blur, or opacity tricks are needed — deliberately, so it
renders identically and cheaply in iced.

---

## Token summary

| Element | Value |
|---|---|
| Backdrop | `rgba(0,0,0,0.6)` |
| Card bg / border | `#141414` / `1px #3F3F3F` |
| Card radius / padding / width | `16px` / `40px × 36px` / `380px` |
| Spinner size / stroke | `56px` / `4px` |
| Track color | `#272727` |
| Arc color / sweep | `#00FF66` / `90°` |
| Rotation period / easing | `0.9s` / linear, looping |
| Title color | `#FFFFFF` (IBM Plex 700, 24px) |
| Body color | `#717171` |

---

Reference implementation: `ui_kits/wallet/Installer.jsx` (`Spinner` and
`CompilingModal` components) + the `@keyframes liana-spin` rule in
`ui_kits/wallet/index.html`.
```

ws-plaude:
done

pythcoiner:
@ws-plaude it's near but the spinner is still a lot pixelated, can we have some antialiasing?

ws-plaude:
done

pythcoiner:
@ws-plaude I'd like you to:
- drop the first commit, this is not usefull as the blur is already making buttons no clickable
- move the spinner & modal in liana-ui as 2 new components in liana-ui::component::spinner.rs:
  - spinner() -> Element
  - spinner_modal(title: Display, description: Display) -> Element

ws-plaude:
done

pythcoiner:
@ws-plaude there is something really annoying i this view: everytime we show a modal, the view scroll position is lost and reset, can you try to fix that in a separate commit?

ws-plaude:
done

pythcoiner:
@ws-plaude what part of the code lock the scroll position?

ws-plaude:
The scroll state is preserved by keeping the same widget tree around the editor:

- `liana-gui/src/installer/step/descriptor/editor/mod.rs:670` always returns `Modal::with_optional(content, modal)`, even when no modal is shown.
- `liana-ui/src/widget/modal.rs:64-77` keeps the base view as child 0 in `children()` and `diff()`.
- `liana-ui/src/widget/modal.rs:90-115` and `176-185` keep layout, update, and operate delegated to `state.children[0]`.

Before that, the view switched between `content` and `Modal::new(content, ...)`, so iced rebuilt the base scrollable from a different root widget and its scroll state reset.

pythcoiner:
superseded by #2169

PR #2155, opened 2026-05-28
UI 3

pythcoiner:
This PR inplement the rework of the send panel described in #2105:
- change the flow, forcing user to add a label before display the address
- rework general appearance following figma
- after this it became mandatory to enter a label when generating an address
- display address chuncked by 4/6/no chunks (coldcard display chucnks of 4, specter 6)  

closes #2105
closes #837 
closes #1024
closes #1962
closes #1967

PR #2170, opened 2026-06-08
gui: chunk address displays

ws-plaude:
Display wallet addresses in 4-character chunks, dimming odd chunks for visual verification.

Clicking an address toggles that widget between chunked and plain display.

### DIFF
--- a/liana-gui/src/app/view/coins.rs
+++ b/liana-gui/src/app/view/coins.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use iced::{widget::Space, Alignment, Length};
 
 use liana_ui::{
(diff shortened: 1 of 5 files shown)


pythcoiner:
@ws-plaude can you split it in 2 commits:
- one that create a component in liana-ui::component::address
- one that impl usages

pythcoiner:
@ws-plaude about dimming chuncks, maybe I was not enough clear, what I want is :
- even chunks are not dimmed
- odd chunks are dimmed

can the component implement Widget and own its (toggled) state rather leaking in the view state?

ws-plaude:
done

ws-plaude:
done

pythcoiner:
superseded by #2155

PR #2180, opened 2026-06-12
gui: add copy button feedback

ws-plaude:
`btn_copy()` now returns a stateful copy button widget.

On click, it keeps publishing the caller message, swaps the copy icon for a check icon, and switches back after 3 seconds.

Fixes #2179

PR #2183, opened 2026-06-17
Use copy feedback buttons everywhere

ws-plaude:
Replace the remaining inline copy buttons with `btn_copy()` so they use the same temporary checkmark feedback added in #2180.

This covers the wallet views, PSBT views, transaction views, coin details, settings rows, and the installer descriptor copy button.

### DIFF
--- a/liana-gui/src/app/view/coins.rs
+++ b/liana-gui/src/app/view/coins.rs
@@ -200,13 +200,9 @@ fn coin_list_view<'a>(
                                             Row::new()
                                                 .align_y(Alignment::Center)
                                                 .push(address_view(address.clone()))
-                                                .push(
-                                                    Button::new(icon::clipboard_icon())
-                                                        .on_press(Message::Clipboard(
-                                                            address.clone(),
-                                                        ))
-                                                        .style(theme::button::transparent_border),
-                                                ),
+                                                .push(button::btn_copy(Some(Message::Clipboard(
+                                                    address.clone(),
+                                                )))),
                                         )
                                         .spacing(5),
                                 )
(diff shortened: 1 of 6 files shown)


pythcoiner:
@ws-plaude update_spend_view has no callers anywhere in liana-gui — it's dead code (the live import/update-PSBT screen is psbts.rs::import_psbt_view, which has no copy button). Please delete the whole update_spend_view function in its own commit rather than converting its copy button, it never renders, so the btn_copy change here is untestable. Drop this hunk along with the function.

pythcoiner:
@ws-plaude in settings > wallet the [Copy] button has been replaced by the icon-only button, I think it's fine but we should then move the button on the previous row (after the descriptor string)

pythcoiner:
> I think it's fine but we should then move the button on the previous row (after the descriptor string)

@ws-plaude we should do that also in the installer at the descriptor backup step

ws-plaude:
done: https://github.com/wizardsardine/liana/pull/2183
Verified:
- `update_spend_view` is deleted.
- the unused `separation` import is gone.
- `cargo check -p liana-gui` passes.
session_id: 20260617_070451_581e61

ws-plaude:
done

ws-plaude:
done

pythcoiner:
image

image

tACK bc7b0319

ws-plaude:
Handled.
The comment is a tACK for commit bc7b0319, with no requested follow-up. PR #2183 is still open and the referenced commit is the current latest commit on the PR branch.
No GitHub reply was posted because that would be acknowledgement-only.
session_id: 20260617_075817_937ea8

PR #2184, opened 2026-06-17
gui: fix business key type tooltip

### LINKED ISSUE #2118: nit: "key" instead of "keys" in the Key Type info tip for Liana Business
When I add a key in Liana Business I need to select a Key Type. We describe the different types in an info tip:

Image

For consistency, we should use "key" instead of "keys" for Internal and External descriptions too.

ws-plaude:
Use singular "key" for the Internal and External Key Type tooltip descriptions.

Fixes #2118.


pythcoiner:
@nondiremanuel 

image

tACK ed949749

PR #2185, opened 2026-06-17
gui: leave new business path timelock empty

### LINKED ISSUE #2140: Default timelock should not be set for Liana Business
We currently have 1 day as a default value for Liana Business recovery paths timelock:

Image

In order to avoid mistakes for the WS admin, it would be better this field is not set so that it requires an action before continuing.

ws-plaude:
The business installer create-path modal now opens the recovery path timelock input empty instead of pre-filling 1 day. The empty field stays invalid, so the save button remains disabled until a timelock is entered.

Fixes #2140


pythcoiner:
tACK ad422667

image

@nondiremanuel

PR #2190, opened 2026-07-03
ci: isolate Python test dependencies

pythcoiner:
integrations test were failing because weirds dependencies issues, this is solved by running it in venv & re-creating a new venv for each CI task.
The machine run a single runner & pip cache packages so it's near no-cost.


edouardparis (approved):
ACK 154559040305e510120c4498fe26556d4f9eb9a4

PR #2188, opened 2026-06-29
 UI 5

pythcoiner:
This PR introduce an UI rework of business installer

## Comments adressed:

- [x] (token step) text gets wrongly truncated in some instances
- [x] path/keys cards dimmed when disabled

## To address:

- [x] A wallet manager should only see the "Keys and signers" tab when the wallet is in Draft status. First time I enter in a wallet as manager I see this:
- [x] "Mark keys as ready" button is not working (nothing happens)
- [x] Trying to edit/clear a SN key, we have a "you can't edit manually"
- [x] Skip registration must be a secondary button at bottom
- [x] Path timelock pills icon must be same color than text

PR #2195, opened 2026-07-10
gui: reduce button proportions

ws-plaude:
Stacks on #2188.

Reduces shared button padding and width presets so the GUI buttons take less space.


ws-plaude:
Superseded by https://github.com/pythcoiner/liana/pull/7, based on `ui_5`.

Issue #2198, opened 2026-07-13
Fix primary vs secondary vs tertiary buttons definition in both installer and wallet itself


nondiremanuel:
After the redefinition of the secondary and tertiary button styles, we have some mismatches in both the installer and the wallet itself on which button should be of one kind or another, both in liana gui and liana business binaries. I noted the following edits to be made (or at least tried out) divided by "category":

### liana gui installer:

- “Share Xpubs” tertiary througout the whole installer
- Trash icon tertiary in the launcher
- “+ Add wallet” tertiary in the launcher
- “< Back to wallet list” tertiary in the “Welcome Back” screen
- Template options (”Simple Inheritance” …) tertiary in the “Choose wallet type” step
- detected hardware buttons in the “Select key source” modal → tertiary
- “Clear all” and “Customize” buttons in the “Set Keys” step → tertiary
- “+Add recovery option” and “+ Add Safety Net” in the “Set Keys” step of the “Build your own” template → tertiary
- buttons in the “Other options” section in the “Select key source” modal → tertiary
- “Next” button in “Back up your mnemonic” step → primary
- “Next” button in “Register descriptor” step → primary
- “Send token” in the “Liana Connect” authentication step → disabled when email text field is empty, primary when active
- “< Change Email” and “Resend token” in Liana Connect authentication step → tertiary
- “Next” button in “Start Bitcoin full node” step → primary
- “Next” button in “give your wallet an alias” step → primary

### liana business installer:

- “< Change Email” and “Resend token” in Liana Business authentication step → tertiary

### liana gui / liana business wallet:

- “See transaction details” in both Outgoing payment and Incoming payment pages → tertiary
- “Clear” in Send panel → tertiary
- “< Previous” in signing step of Send process → tertiary
- “Save” in signing step of Send process → secondary
- “Export” and “Import”  in signing step of Send process → tertiary
- device selection buttons in “Select signing device to sign with:” modal → tertiary
- “Delete” in PSBT screen → tertiary
- “Verify on hardware device” and “Show QR Code” in Receive panel (newly generated address modal and already generated address list) → tertiary
- Copy icon in eceive panel (newly generated address modal and already generated address list) → “secondary” (green border icon)
- “Back up encrypted descriptor” and “Register on hardware device” in “Settings>Wallet” page → tertiary

nondiremanuel:
@ws-plaude can you give a try to this? Open a PR on wizardsardine/liana repo based on the last commit of #2196

[...]

PR #2196, opened 2026-07-10
gui: reduce button proportions

ws-plaude:
Stacked on #2188.

Reduce the shared button padding and width presets for the normal and compact button components.

PR #2199, opened 2026-07-13
gui: adjust button roles

ws-plaude:
Stacked on #2196.

Adjust the installer, wallet, and shared component button roles listed in #2198.


nondiremanuel:
Thanks. Testing it, I have further requests to do / some changes with respect to what I defined previously:

### liana gui installer
- “+ Add wallet” in the launcher -> instead of tertiary try to use the dashed button used for "Connect with another email" in the liana business installer with the liana gui style (you can define a new component if needed)
- The two "Select" buttons in the "Welcome back" and the two in the "Choose backend" screens -> tertiary
- The trash icons for keys in the "Set keys" page (visible for the "Build your own" template) -> tertiary
- "Edit" button for already set keys in the "Set keys" page -> tertiary
- “Send token” in the “Liana Connect” authentication step should be disabled when no text is presend in the email text box (as it is when there is a wrongly formatted email)
- The following that I defined as tertiary, should instead be selectable buttons (as the wallet list in the launcher):
   - Template options (”Simple Inheritance” …) in the “Choose wallet type” step
   - detected hardware buttons in the “Select key source” modal
   - buttons in the “Other options” section in the “Select key source” modal

### liana gui / liana business wallet
- "+ Add payment" in the Send screen -> tertiary
- Copy icon in Receive panel (newly generated address modal and already generated address list) -> only the icon should be visible with green border (no button border)
- "Update" in “Settings>Wallet” page → tertiary
- device selection buttons in “Select signing device to sign with:” modal → instead of tertiary they should be selectable buttons (as the wallet list in the launcher)

@ws-plaude can you try implement this?

nondiremanuel:
It looks good. Some further comments:

### liana gui installer

- “+ Add wallet” in the launcher -> drop the component you made and keep it as tertiary. We will work on the right component in a separate issue/PR
- In the "Select key source" modal, for the "Other options" title, can you restore the button component we have currently on master (if it's still defined in the current codebase)?


### liana gui / liana business wallet
- In liana gui an, the copy icon should get green on hover as it does in liana business (getting blue there). Can you do it or should we work separately on the component?

@ws-plaude try implement those. Answer and discuss if / when needed

PR #2200, opened 2026-07-13
liana-ui: forward base overlays without modal

pythcoiner:
This fix an issue where pick_list() for descriptor type in liana-gui installer was not receiving click event, and in a general way, several components implementing widget had their event blocked.


edouardparis (approved):
utACK 0df759bc64892b4c98e8123951d4e4bda854573a

PR #2204, opened 2026-07-17
gui: limit label inputs to 80 bytes

pythcoiner:
Change the label length limit from 100 to 80 to match liana backend & enforce it in the gui from a single source of truth


edouardparis (approved):
ACK ed8de0170fdaacfe8adebd30efdea33ec9ba605a

PR #2206, opened 2026-07-20
Hide liana-business wallets to liana-gui installer

edouardparis:
We should not suggests these wallets to liana-gui to better have clear separation of products


pythcoiner:
utACK 5e9facb

PR #2215, opened 2026-07-24
launcher: make wallet entries consisatnt in height

### LINKED ISSUE #2208: Cards of wallets with different backends have different sizes in liana gui
As you can see from the screenshot, wallets with Liana Connect as a backend have a greater height due to the additional row used to show the email address:

Image

One way we could address this would be to use a slightly smaller font size for both the fingerprint and the email address and have them in the same row, similar to what we have in the following [mockup ](https://claude.ai/design/p/610495bf-a69c-41dc-a9aa-4ccb2a4518ec?file=index.html&via=share) (note that many things are not updated with what we have on master). Of course in this case we should  apply a minimum spacing between the two and truncation of the email address in case they don't fit both, but sizes should be so that the majority of the email address info is visible (the fact that fingerprint is fixed length helps here). 
Other suggestions and solutions are welcome.

pythcoiner:
image

closes #2208

Note: email are shortened if 35+ chars

### DIFF
--- a/liana-gui/src/launcher.rs
+++ b/liana-gui/src/launcher.rs
@@ -11,7 +11,7 @@ use liana_ui::{
         card, installer as installer_layout,
         list::{self, EntryAccent},
         notification,
-        text::new,
+        text::{new, short_email},
     },
     icon, image, theme,
     widget::{modal::Modal, Column, Container, Element, SpaceExt},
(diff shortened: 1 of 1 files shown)


edouardparis (approved):
ACK b8993c3c05c548e8f94d142be47c4479910ba4bc

PR #2216, opened 2026-07-24
Move edit detailed info on hover

### LINKED ISSUE #2207: Move the editor info on hover to avoid wrapping in Liana Business wallet list
We currently wrap on the "Edited by" sentence even with not-so-long emails:

Image

I think we could avoid this by just showing when it was edited and a way to show on hover who edited it (either with an icoin or underlying the time to suggest there is some info to be seen, as Github does with comments for example)

For example, we would have:
"Edited 3 weeks ago" visible
"Edited by manuel@wizardsardine.com on the [date and time]" on hover

Open to different suggestions.

pythcoiner:
image

closes #2207


edouardparis (approved):
ACK 5ed109f94c5656f80d1177427ee0295ef4bad4f9

PR #2219, opened 2026-07-24
gui: fix signing device entry

pythcoiner:
image

Drop duplicate device kind & wrap warning

### DIFF
--- a/liana-gui/src/view/hw.rs
+++ b/liana-gui/src/view/hw.rs
@@ -229,9 +229,7 @@ where
     let kind = kind.map(|kind| kind.to_string());
     let alias = alias.map(|alias| alias.to_string());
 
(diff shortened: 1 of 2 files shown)


nondiremanuel:
> image
> 
> Drop duplicate device kind & wrap warning

why do you talk about "duplicate device kind"? Wasn't the first one the alias given by the user?

nondiremanuel:
> why do you talk about "duplicate device kind"? Wasn't the first one the alias given by the user?

We've clarified elsewhere that the duplication happens when no alias is set by the user (e.g. device not belonging to the specific Liana wallet).

edouardparis (approved):
ACK 0368b250d6c78a62464e1db115250b7de641a693

PR #2222, opened 2026-07-27
business-installer: Add key binding to select signet network 

pythcoiner:
This PR:
- disable auto-switching to email login step when no accounts
- in the login account select step, if user hit `escape` key 3 times we change to signet


edouardparis (approved):
ACK 5204eb8c9d1e5c37103eebac129bd3618cdd4716

PR #2226, opened 2026-07-29
lianad: warn `createrecovery` to an address of the same wallet.

### LINKED ISSUE #1654: Warn if own address is used for recovery
The recovery transaction should be to an external address as otherwise the funds will be re-locked, assuming it is no longer possible to spend using the primary path.

We don't currently check whether the address used for recovery belongs to the same wallet. We should check this and at the very least warn the user or even return an error and prevent such a recovery transaction being created.

I think it would be easier to check this on the backend since the backend will have easier access to known wallet addresses. In the case of a warning, we could reuse the existing warning system that is currently used to warn the user about additional fees due to CPFP and change outputs being too small.

There may be some users who wish to lock funds in the primary path with an unspendable key and only intend to spend using recovery paths. In that case, it may be better for now to allow such recovery spends back to the user's own wallet and simply show a warning.

qlrd:
~~This commit adds a new error (code 1001, just after the existing 1000) bound to `createrecovery` -- as warning response when a user wants to re-lock funds from/to addresses derived from the wallet descriptor.~~

~~As stated by jp1ac4 the starting point is to check this on the backend since it have easier access to known wallet addresses (needs a check on how to deal with derivation indices beyond what the DB knows).~~

~~The proposed flow check if the provided address belongs to the user if the DB have `Some` or `None`. If `None`, just continue the flow; if `Some`, a `RecoveryToOwnAddress` error will be added to a `warnings` field as a string.~~

EDIT: As stated by jp1ac4 the starting point is to check this on the backend since the backend has easier access to known wallet addresses (needs a check on how to deal with derivation indices beyond what the DB knows -- worth of a follow-up).

If the DB knows the sweep address as a derived one, it belongs to the wallet's descriptor and recovered funds would be locked under the same descriptor again.

In that case a `CreateRecoveryWarning::ToOwnAddress` enum is used as a warning with the address to be added to a new `warnings` field of the `createrecovery` response so the rpc command could respond with `{"psbt": ..., "warnings": [{ "to_own_address": "<addr>"}] }`.

Close #1654.

## Details

~~A follow-up commit (the GUI confirmation flow) is being drafted while discussing.~~

EDIT: the `gui` commit handles the previous commit `createrecovery` rpc update to `Daemon` trait, so when user request do a self-transfer when using recovery options on GUI, it will be warned about the re-lock.

EDIT 2: Maybe it needs some GUI tests? maybe now or follow up?

### DIFF
--- a/liana-gui/src/daemon/client/mod.rs
+++ b/liana-gui/src/daemon/client/mod.rs
@@ -3,8 +3,10 @@ use std::fmt::Debug;
 use std::iter::FromIterator;
 
 use async_trait::async_trait;
-use lianad::bip329::Labels;
-use lianad::commands::{GetLabelsBip329Result, UpdateDerivIndexesResult};
+use lianad::{
+    bip329::Labels,
+    commands::{CreateRecoveryWarning, GetLabelsBip329Result, UpdateDerivIndexesResult},
+};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
@@ -209,7 +211,7 @@ impl<C: Client + Send + Sync + Debug> Daemon for Lianad<C> {
         coins_outpoints: &[OutPoint],
         feerate_vb: u64,
         sequence: Option<u16>,
-    ) -> Result<Psbt, DaemonError> {
+    ) -> Result<(Psbt, Vec<CreateRecoveryWarning>), DaemonError> {
         let mut params = serde_json::Map::new();
         params.insert("address".to_string(), json!(address));
         params.insert("outpoints".to_string(), json!(coins_outpoints));
@@ -218,7 +220,7 @@ impl<C: Client + Send + Sync + Debug> Daemon for Lianad<C> {
             params.insert("timelock".to_string(), json!(sequence));
         }
         let res: CreateRecoveryResult = self.call("createrecovery", Some(params))?;
-        Ok(res.psbt)
+        Ok((res.psbt, res.warnings))
     }
 
     async fn get_labels(
--- a/liana-gui/src/daemon/embedded.rs
+++ b/liana-gui/src/daemon/embedded.rs
@@ -1,5 +1,9 @@
-use lianad::bip329::Labels;
-use lianad::commands::UpdateDerivIndexesResult;
+use lianad::{
+    bip329::Labels,
+    commands::{CoinStatus, CreateRecoveryWarning, LabelItem, UpdateDerivIndexesResult},
+    config::Config,
+    DaemonControl, DaemonHandle,
+};
 use std::collections::{HashMap, HashSet};
 use tokio::sync::Mutex;
 
@@ -8,11 +12,6 @@ use crate::dir::LianaDirectory;
 use async_trait::async_trait;
 use liana::miniscript::bitcoin::{
     address, bip32::ChildNumber, psbt::Psbt, Address, Network, OutPoint, Txid,
-};
-use lianad::{
-    commands::{CoinStatus, LabelItem},
-    config::Config,
-    DaemonControl, DaemonHandle,
 };
 
 pub struct EmbeddedDaemon {
@@ -236,11 +235,11 @@ impl Daemon for EmbeddedDaemon {
         coins_outpoints: &[OutPoint],
         feerate_vb: u64,
         sequence: Option<u16>,
-    ) -> Result<Psbt, DaemonError> {
+    ) -> Result<(Psbt, Vec<CreateRecoveryWarning>), DaemonError> {
         self.command(|daemon| {
             daemon
                 .create_recovery(address, coins_outpoints, feerate_vb, sequence)
-                .map(|res| res.psbt)
+                .map(|res| (res.psbt, res.warnings))
                 .map_err(|e| DaemonError::Unexpected(e.to_string()))
         })
         .await
--- a/liana-gui/src/daemon/mod.rs
+++ b/liana-gui/src/daemon/mod.rs
@@ -16,10 +16,11 @@ use liana::miniscript::bitcoin::{
     psbt::Psbt,
     secp256k1, Address, Network, OutPoint, Txid,
 };
-use lianad::bip329::Labels;
-use lianad::commands::UpdateDerivIndexesResult;
 use lianad::{
-    commands::{CoinStatus, LabelItem, TransactionInfo},
+    bip329::Labels,
+    commands::{
+        CoinStatus, CreateRecoveryWarning, LabelItem, TransactionInfo, UpdateDerivIndexesResult,
+    },
     config::Config,
     StartupError,
 };
@@ -182,7 +183,7 @@ pub trait Daemon: Debug {
         coins_outpoints: &[OutPoint],
         feerate_vb: u64,
         sequence: Option<u16>,
-    ) -> Result<Psbt, DaemonError>;
+    ) -> Result<(Psbt, Vec<CreateRecoveryWarning>), DaemonError>;
     async fn list_txs(&self, txid: &[Txid]) -> Result<model::ListTransactionsResult, DaemonError>;
     async fn get_labels(
         &self,
--- a/liana-gui/src/services/connect/client/backend/mod.rs
+++ b/liana-gui/src/services/connect/client/backend/mod.rs
@@ -16,7 +16,10 @@ use liana::{
 };
 use lianad::{
     bip329::Labels,
-    commands::{CoinStatus, GetInfoDescriptors, LCSpendInfo, LabelItem, UpdateDerivIndexesResult},
+    commands::{
+        CoinStatus, CreateRecoveryWarning, GetInfoDescriptors, LCSpendInfo, LabelItem,
+        UpdateDerivIndexesResult,
+    },
     config::Config,
 };
 use reqwest::{Error, IntoUrl, Method, RequestBuilder};
@@ -936,7 +939,7 @@ impl Daemon for BackendWalletClient {
         coins_outpoints: &[OutPoint],
         feerate_vb: u64,
         sequence: Option<u16>,
-    ) -> Result<Psbt, DaemonError> {
+    ) -> Result<(Psbt, Vec<CreateRecoveryWarning>), DaemonError> {
         let timelock = sequence.ok_or(DaemonError::Unexpected("Missing sequence".to_string()))?;
         let res: api::DraftPsbt = self
             .inner
@@ -955,7 +958,13 @@ impl Daemon for BackendWalletClient {
             )
             .await?;
 
-        Ok(res.raw)
+        let warnings = res
+            .warnings
+            .into_iter()
+            .map(CreateRecoveryWarning::String)
+            .collect();
+
+        Ok((res.raw, warnings))
     }
 
     async fn get_labels(
--- a/lianad/src/commands/mod.rs
+++ b/lianad/src/commands/mod.rs
@@ -1251,6 +1251,10 @@ impl DaemonControl {
     /// otherwise not currently recoverable using the given recovery path.
     ///
     /// Note that not all coins may be spendable through a single recovery path at the same time.
+    ///
+    /// A warning will be included in the result's `warnings` field if the sweep address is
+    /// known to belong to this same wallet, since recovered funds would be locked under the
+    /// same descriptor again.
     pub fn create_recovery(
         &self,
         address: bitcoin::Address<address::NetworkUnchecked>,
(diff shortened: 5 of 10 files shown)


lianad/src/commands/mod.rs
```diff
@@ -1311,7 +1325,21 @@ impl DaemonControl {
...
+            if !allow_own_address.unwrap_or(false) {
+                return Err(CommandError::RecoveryToOwnAddress(sweep_addr.addr));
+            }
+            warnings.push(format!(
+                "Address {} belongs to the same wallet. Recovered funds would \
+                be locked under the same descriptor again.",
+                sweep_addr.addr
+            ));
+        }
+
```
pythcoiner, line 1325: better for maintenance

```suggestion
        if sweep_addr_info.is_some() {
            let err = CommandError::RecoveryToOwnAddress(sweep_addr.addr.clone());
            if !allow_own_address.unwrap_or(false) {
                return Err(err.clone());
            }
            warnings.push(err.to_string());
        }

```

lianad/src/commands/mod.rs
```diff
@@ -1251,12 +1261,16 @@ impl DaemonControl {
...
+    ///
+    /// By default an error is returned if the sweep address belongs to this same wallet. Set
+    /// `allow_own_address` to allow recovered funds to be locked under the same descriptor.
     pub fn create_recovery(
         &self,
         address: bitcoin::Address<address::NetworkUnchecked>,
         coins_outpoints: &[bitcoin::OutPoint],
         feerate_vb: u64,
         timelock: Option<u16>,
+        allow_own_address: Option<bool>,
```
pythcoiner: I can overlook something, but I dont think this flag is really useful, it seems to me we can just return warnings

qlrd: I will get my annotations made before "that" and lyn :)

qlrd: I think i got your idea and make sense. LMK if it' correct: no need to `allow_own_address` option so the directed `warnings` value could go directly to GUI when doing it, and no need to call twice (one for "doing, deny and warning plus one for doing and warning" `vs` "one for doing and warning" ? My UI/UX thinking was to deny first, so the user could have full knowledge about will happen and need to be sure about before do. Your perspective is that the user already know the risks?

lianad/src/commands/mod.rs
qlrd: done

lianad/src/commands/mod.rs
pythcoiner: My perspective is we just want to warn the user, not prevent doing this, and making 2 call is overkill/overcpmplex imo

qlrd:
I've made the required changes and verified tested here with:

```bash
cargo fmt --all --check
cargo clippy
cargo test
(venv) python3 -m pytest tests/
```

If you agree with the current code, I will proceed with the GUI plumbing (if not, let me know). Thxs for the tips, it reduced the diff a lot.

lianad/src/commands/mod.rs
```diff
@@ -1311,7 +1324,17 @@ impl DaemonControl {
             return Err(CommandError::RecoveryNotAvailable);
         }
 
+        // If DB knows (as derived address) about the provided address, it means
+        // the sweep address belongs to this wallet.
+        // limit `with_capacity` to 1 for now (maybe another better approach?)
+        // if more warning messages are desired in this command
         let sweep_addr_info = sweep_addr.info;
+        let mut warnings: Vec<String> = Vec::with_capacity(1);
```
pythcoiner: I think `with_capacity(1)` is overkill here

pythcoiner:
nit, it's not a server error:

```diff
diff --git a/lianad/src/jsonrpc/rpc.rs b/lianad/src/jsonrpc/rpc.rs
index d33aae7a5..f6536e01f 100644
--- a/lianad/src/jsonrpc/rpc.rs
+++ b/lianad/src/jsonrpc/rpc.rs
@@ -51,9 +51,6 @@ pub struct Request {
 /// A failure to broadcast a transaction to the P2P network.
 const BROADCAST_ERROR: i64 = 1_000;

-/// A failure while sweeping funds to an address from the wallet itself.
-const RECOVERY_TO_OWN_ADDRESS_ERROR: i64 = 1_001;
-
 /// JSONRPC2 error codes. See https://www.jsonrpc.org/specification#error_object.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum ErrorCode {
@@ -170,16 +167,13 @@ impl From<commands::CommandError> for Error {
             | commands::CommandError::OutpointNotRecoverable(..) => {
                 Error::new(ErrorCode::InvalidParams, e.to_string())
             }
-            commands::CommandError::RescanTrigger(..) => {
+            commands::CommandError::RescanTrigger(..)
+            | commands::CommandError::RecoveryToOwnAddress(..) => {
                 Error::new(ErrorCode::InternalError, e.to_string())
             }
             commands::CommandError::TxBroadcast(_) => {
                 Error::new(ErrorCode::ServerError(BROADCAST_ERROR), e.to_string())
             }
-            commands::CommandError::RecoveryToOwnAddress(_) => Error::new(
-                ErrorCode::ServerError(RECOVERY_TO_OWN_ADDRESS_ERROR),
-                e.to_string(),
-            ),
         }
     }
 }
```

pythcoiner:
overall lgtm, feel free to continue with the GUI

lianad/src/commands/mod.rs
```diff
@@ -1523,6 +1546,9 @@ pub struct TransactionInfo {
 pub struct CreateRecoveryResult {
     #[serde(serialize_with = "ser_to_string", deserialize_with = "deser_fromstr")]
     pub psbt: Psbt,
+
+    #[serde(default)]
+    pub warnings: Vec<String>,
```
pythcoiner: nit:

```suggestion
    pub warnings: Vec<CommandError>,
```

seems better so consumer can match on a strong type rather than a String

edit: maybe a new `CreateRecoveryWarning::ToOwnAddress` enum is more appropriate rather than `commands::CommandError::RecoveryToOwnAddress`

`

qlrd: Done and did some rebase

qlrd:
The pumbled code into GUI preview (with floresta backend btw :)
liana-gui

pythcoiner:
@nondiremanuel I guess you'd like to review the wording of the warning

nondiremanuel:
> @nondiremanuel I guess you'd like to review the wording of the warning

Yes, thanks. I would personally use something more action oriented like:

`⚠ Recovery address belongs to the same wallet. If you can no longer spend with the primary path, use an address from a different wallet instead.`

Happy to discuss if you guys prefer other options.

qlrd:
> > @nondiremanuel I guess you'd like to review the wording of the warning
> 
> Yes, thanks. I would personally use something more action oriented like:
> 
> `⚠ Recovery address belongs to the same wallet. If you can no longer spend with the primary path, use an address from a different wallet instead.`
> 
> Happy to discuss if you guys prefer other options.

I liked this one. For me sometimes is hard to put things in a clear way. And wen readed that, i think your phrase is the most synthetic and pedagogical than mine.

qlrd:
> > @nondiremanuel I guess you'd like to review the wording of the warning
> 
> Yes, thanks. I would personally use something more action oriented like:
> 
> `⚠ Recovery address belongs to the same wallet. If you can no longer spend with the primary path, use an address from a different wallet instead.`
> 
> Happy to discuss if you guys prefer other options.

Following your idea
liana

lianad/src/testutils.rs
```diff
@@ -363,9 +379,9 @@ impl DatabaseConnection for DummyDatabase {
 
     fn derivation_index_by_address(
         &mut self,
-        _: &bitcoin::Address,
+        addr: &bitcoin::Address,
     ) -> Option<(bip32::ChildNumber, bool)> {
-        None
+        self.db.read().unwrap().derived_addresses.get(addr).copied()
```
pythcoiner: I'd rather avoid storing derived address, instead I think you should try to derive addresses up to `DummyDbState.deposit_index`/`.change_index`, it feel weird to have to manually populate `derived_addresses` in tests

liana-gui/src/daemon/client/mod.rs
```diff
@@ -4,7 +4,7 @@ use std::iter::FromIterator;
 
 use async_trait::async_trait;
 use lianad::bip329::Labels;
-use lianad::commands::{GetLabelsBip329Result, UpdateDerivIndexesResult};
+use lianad::commands::{CreateRecoveryWarning, GetLabelsBip329Result, UpdateDerivIndexesResult};
```
pythcoiner: nit: you should merge imports

```suggestion
use lianad::{
    bip329::Labels,
    commands::{CreateRecoveryWarning, GetLabelsBip329Result, UpdateDerivIndexesResult},
};

```

liana-gui/src/daemon/embedded.rs
```diff
@@ -1,5 +1,7 @@
-use lianad::bip329::Labels;
-use lianad::commands::UpdateDerivIndexesResult;
+use lianad::{
+    bip329::Labels,
+    commands::{CreateRecoveryWarning, UpdateDerivIndexesResult},
+};
```
pythcoiner, line 6: nit: merge imports

```
use lianad::{
    bip329::Labels,
    commands::{CoinStatus, CreateRecoveryWarning, LabelItem, UpdateDerivIndexesResult},
    config::Config,
    DaemonControl, DaemonHandle,
};
```

liana-gui/src/daemon/mod.rs
```diff
@@ -17,7 +17,7 @@ use liana::miniscript::bitcoin::{
     secp256k1, Address, Network, OutPoint, Txid,
 };
 use lianad::bip329::Labels;
-use lianad::commands::UpdateDerivIndexesResult;
+use lianad::commands::{CreateRecoveryWarning, UpdateDerivIndexesResult};
```
pythcoiner: nit: merge imports

```
use lianad::{
    bip329::Labels,
    commands::{
        CoinStatus, CreateRecoveryWarning, LabelItem, TransactionInfo, UpdateDerivIndexesResult,
    },
    config::Config,
    StartupError,
};

```

liana-gui/src/services/connect/client/backend/mod.rs
```diff
@@ -955,7 +958,8 @@ impl Daemon for BackendWalletClient {
             )
             .await?;
 
-        Ok(res.raw)
+        // the connect API does not know `CreateRecoveryWarning` and gives warnings as raw text.
```
pythcoiner: @nondiremanuel we must implement this check also on `liana-backend` side in a follow-up

pythcoiner:
tested `51632a7ec`, overall good, see my nits comments 

image

qlrd:
> tested `51632a7ec`, overall good, see my nits comments

Sure, thxs a lot. I will get a look and apply them.

pythcoiner:
@qlrd [this](https://github.com/pythcoiner/liana/pull/8/changes/892eac88859066aa63a5ea7cde6e94ca530df3e0) will be needed for handling warnings from liana-connect

lianad/src/testutils.rs
qlrd: yup, bad old habits with test database. Thxs!

qlrd:
> @qlrd [this](https://github.com/pythcoiner/liana/pull/8/changes/892eac88859066aa63a5ea7cde6e94ca530df3e0) will be needed for handling warnings from liana-connect

I think is done (previous fails on CI solved) -- lmk if it's correct for liana-connect (also the db approach on testutils)

lianad/src/commands/mod.rs
```diff
@@ -3242,6 +3283,16 @@ mod tests {
             Err(CommandError::OutpointNotRecoverable(dummy_op, 11)),
         );
 
+        // allow to recover with own address with a warning attached.
+        db_handle.set_descriptor(control.get_info().descriptors.main);
```
pythcoiner: hum, my view on this is we just moving the problem around, i originally not liked the `.insert_derived_address()` because I feel it is not natural to do so (and one can forgot in test added in the future), but `set_descriptor()` is just a tiny better, maybe we should instead pass descriptoras as mandatory argument of `DummyDatabase::new()`?

lianad/src/testutils.rs
```diff
@@ -363,8 +370,29 @@ impl DatabaseConnection for DummyDatabase {
...
+            (desc.receive_descriptor(), db.deposit_index, false),
+            (desc.change_descriptor(), db.change_index, true),
+        ];
+
+        // From first to last index used, if gets we found
+        for (wall_desc, idx, is_change) in data {
+            for i in 0..=u32::from(idx) {
+                let index = bip32::ChildNumber::from(i);
+                let look = wall_desc.derive(index, &context).address(net);
+                if look == addr.clone() {
```
pythcoiner: I dont think you need to clone:

```suggestion
                if &look == addr {
```

lianad/src/commands/mod.rs
qlrd: Applied, made muuuuch more sense (if it was that what you meant). Also moved the default xpub and timelock to constants, so it would be easier to edit tests

pythcoiner:
tACK 4614faba8

PR #2229, opened 2026-07-30
business-installer: use state network for app handoff

edouardparis:
The Escape key binding switches State's network to signet, but BusinessInstaller kept its own copies of network and datadir frozen at construction. exit_maybe() passed the stale mainnet network to NextState::RunLianaBusiness, so the app connected to the mainnet API with mainnet-cached credentials and failed to fetch the signet wallet.

Drop the duplicated fields and read network and datadir from State.


edouardparis (commented):
Self-ACK 41e59c86cd3b20f890c36dc04a7662a72260c774

PR #2230, opened 2026-07-30
business-installer: use state network for app handoff

edouardparis:
The Escape key binding switches State's network to signet, but BusinessInstaller kept its own copies of network and datadir frozen at construction. exit_maybe() passed the stale mainnet network to NextState::RunLianaBusiness, so the app connected to the mainnet API with mainnet-cached credentials and failed to fetch the signet wallet.

Drop the duplicated fields and read network and datadir from State.

backport of #2229


edouardparis (commented):
Self-ACK b12fa05bb65846f09e3973bc0b9f70b5f65ceca4

Issue #2241, opened 2026-08-06
Warn users about ColdCard Mk3 seed-entropy flaw (Coinkite advisory, Jul 2026)


nrobi144:
## Summary

On 30–31 July 2026, Coinkite published a security advisory disclosing that a firmware build error caused ColdCard seed generation to silently fall back to a software RNG (MicroPython) instead of the STM32 hardware RNG. On the **Mk3 this reduced generated-seed entropy to roughly ~40 bits** (vs. the intended 128) — brute-forceable on commodity hardware. Approximately **594 BTC (~$38M) was drained from ~500 wallets on 31 July 2026** (Coinkite's investigation into that sweep is ongoing).

Affected, per Coinkite:
- **Mk2 / Mk3:** firmware **4.0.1 – 4.1.9** (fixed in **v4.2.0**+; device-generated seeds)
- **Mk4 / Mk5 / Q:** reduced to ~72 bits before fixed firmware (standard **v5.6.0** / Edge **v6.6.0X**; Q **v1.5.0Q** / Edge **v6.6.0QX**)
- Seeds created with **≥50 user dice rolls**, or protected by a strong BIP-39 passphrase, are not affected in the same way.
- Fix: update firmware, generate a **new** seed, and migrate funds.

Advisory: https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/

## Why this concerns Liana

Liana documents and supports ColdCard as a signing device (added in v5.0):
- `doc/SIGNING_DEVICES.md` — ColdCard listed (notes it requires Edge firmware)
- Driver lives in the `async-hwi` dependency (`src/coldcard.rs`); GUI xpub import (`.ccxp`) added in liana

Liana does **not** generate the ColdCard seed (the device does), so this is **not a flaw in Liana's code**. But users who set up ColdCard for a Liana wallet may hold funds on affected seeds and would reasonably look to Liana for guidance.

## Suggested changes

1. **Add a visible warning** in `doc/SIGNING_DEVICES.md` and in the ColdCard import UI, linking Coinkite's advisory and the fixed-firmware / reseed guidance.
2. **Deprioritize ColdCard** in the supported-device list until a user is known to be on fixed firmware with a freshly generated seed.

Happy to open a PR for the docs/UI change if that's welcome.

## Disclosure note

[...]

PR #2242, opened 2026-08-06
docs: add Coldcard Mk3 seed-entropy security warning to SIGNING_DEVICES

nrobi144:
## Summary

Adds a security warning to the Coldcard section of `doc/SIGNING_DEVICES.md`.

In July 2026 Coinkite [disclosed](https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/) a firmware build error that reduced generated-seed entropy on **Coldcard Mk3** (firmware 4.0.1–5.0.3) and pre-fix **Mk4/Q**. Users who set up a Coldcard for a Liana wallet may hold funds on an affected seed. This adds a one-line `WARNING:` (matching the existing style used for the Jade section) linking the advisory.

Docs-only change. Refs #2241.

### DIFF
--- a/doc/SIGNING_DEVICES.md
+++ b/doc/SIGNING_DEVICES.md
@@ -29,7 +29,7 @@ writing. It is only supported by the [Edge
 firmware](https://github.com/Coldcard/firmware?tab=readme-ov-file#long-lived-branches).
 For use in Taproot descriptors you should use version 6.3.3 or higher.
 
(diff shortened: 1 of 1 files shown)


nondiremanuel:
Thanks for taking the time to do it. I think it's good to have a warning about it. Maybe we should also point to our own [blog post ](https://wizardsardine.com/blog/coldcard-rng-vulnerability/)which contains more specific info to Liana users.

nrobi144:
Thanks @nondiremanuel — good call. I've added a pointer to Wizardsardine's Liana-specific write-up (https://wizardsardine.com/blog/coldcard-rng-vulnerability/) alongside the Coinkite advisory link in the same WARNING line. Also corrected the firmware range to Coinkite's advisory-accurate values (Mk2/Mk3 4.0.1-4.1.9, plus Mk4/Mk5/Q before their fixed firmware).

PR #2255, opened 2026-08-25
Select key modal

### LINKED ISSUE #2248: Issues with "Select key source" modal
Some issues were found with the "Select key source" modal:
1. The main one is that the modal doesn't scroll, which creates usability problems in policies with many keys:

Image

2. As it's also visible from the previous screenshot, the icon used for "Already used sources" when the key source is an uploaded / pasted pubkey is incorrect. We should use a generic key one. 
3. Background color of the hot key option is different from others, we should align it

Image

pythcoiner:
This PR does 3 things in the select key modal:
 - make the modal scrollable
 - fix style of the hot key tile
 - add a new tile for imported keys
 
image

closes #2248


edouardparis (approved):
utACK 13ffefe68115cf7917f772477f3a16d3d6f74c3d

nondiremanuel:
Why are two of the icons red in the screenshot?

pythcoiner:
ah my bad, the hot key one was originally red, to warn user, but the imported key one should not

pythcoiner:
@nondiremanuel fixed in #2260

PR #2257, opened 2026-08-25
installer: fix node selection view

### LINKED ISSUE #2245: Wrong format for "Address" section title in own node setting step (installer)
When setting the own node address for both "Bitcoin Core" and "Electrum", the "Address" title has the wrong font size and weight (not sure about the font type itself). It should be the same as the other section titles:

Image

pythcoiner:
this PR make the node selection (bitcoind) consistant by applying the same font/size on all labels & add a missing `:` on electrum address label

image

closes #2245


edouardparis (approved):
utACK 4d28435404721e5bd1df44663560272a6d40c8f4

PR #2258, opened 2026-08-25
installer: improve mouseless navigation in the key import flow

pythcoiner:
this PR try to improve the UX in the import key flow (alias modal):
- auto focus into the input field when opening the alias modal
- hitting enter send the "Apply" message
- add a validation on the alias

image


edouardparis (approved):
utACK c41822f00e7b8994bec32ece005b7743f691112d

PR #2267, opened 2026-08-29
[debug]drop wgpu

pythcoiner:


edouardparis:
I understand the intention and how it could prevent us to have those annoying bugs.
I would like us to try to locate why the fallback to tiny-skia is not working before removing this feature.

pythcoiner:
in the last case wgpu was working, but have a bug that make the screen freeze

btw this PR was not intended to be merged, it was to share for an user to test it

PR #2269, opened 2026-08-31
docs(installer): clarify dummy user initialization

ginavalent:
Clarifies why the business installer pre-populates the user collection with a dummy entry, while correcting the spelling of “prefill” and “nonexistent”.


pythcoiner:
utACK 053e5358

PR #2157, opened 2026-05-29
[POC]i18n draft

pythcoiner:
Thi PR is a POC implementing internationalization:
- en
- pt-PT
- it
- fr

PR #2282, opened 2026-09-01
I18n prepa

pythcoiner:
This PR introduce some preparatory changes for #2157: it changes signatures of many ui functions to accept owned string, using `impl Display` as much as possible and fall back to `Into<String>` or `String`


edouardparis (approved):
utACK 26cce71711c6c712d5ae7a55731f7cbfab780c4a

Issue #2290, opened 2026-09-03
Translate template introduction svgs


nondiremanuel:
Currently the template introduction images are svgs:

<img width="1678" height="926" alt="Image" src="https://github.com/user-attachments/assets/45226bf7-417d-4025-82ed-8513ef9ffd00" />

When we translate the app wording to other languages, we need to have these translated as well. We are currently discussing if it's worth "transforming" the svgs to code and include it in the normal translation flow or it's better generate svgs for each language. I open this issue as a discussion starting point.

PR #2293, opened 2026-09-03
Generate the installer template images from Typst sources

LoicPandul:
@pythcoiner 

The installer illustrates each wallet template with a diagram of its spending policies. These were 3 SVGs exported from Figma with English text baked in, which cannot follow the localization of the GUI #2157 

Each image is now described once and rendered per locale:
- `liana-ui/static/images/templates/`: one TOML spec per template (events, policies, keys, states), one TOML file of strings per locale (`en`, `fr`, `it`, `pt-PT`...), and a small Typst library built on the liana-ui design tokens.
- `cargo xtask gen-images` checks the specs and the key parity of the locales, renders every image and rewrites the accessors in `liana-ui/src/image/template_images.rs`. The outputs are committed, so building Liana never needs Typst, and the reproducible builds are untouched (no new dependency, the crate is not a default member).
- CI runs `cargo xtask gen-images --check` with the pinned Typst 0.15.1 and fails on stale outputs.
- The 3 template views take the fallback locale for now. Passing the running locale is a little modification once #2157 lands.

To try it: install Typst 0.15.1 and run `cargo xtask gen-images`.

inheritance

multisig

custom


nondiremanuel:
Thanks for doing it! Looking at the screenshots, it looks good to me.
Related to #2290

pythcoiner:
I'll integrate it into the multi-language flow

LoicPandul:
Closing in favor of #2296

PR #2303, opened 2026-09-15
Coin selection header and divider color

nondiremanuel:
Closes #2279 

Here's how it looks with GREY_4:

image

See [this comment](https://github.com/wizardsardine/liana/issues/2279#issuecomment-5678517193) for more details and an alternative

PR #2298, opened 2026-09-10
UI6 Psbt panel

pythcoiner:
## rework the Psbt panel:
  - follow the mew design
  - not display "0 payments" in PSBT details when send-to-self

image

image

image

image

image

closes #2106

## implement the new style for foldable sections:
  - add a new `FoldableCard` component
  - use consistantly `FoldableCard` all over the gui
  
Note: I've gone a bit out of scope by refactoring coins view as I was integrating `FoldableCard` but kept it in this PR as it is part of the global UI/UX rework in progress

image

close #2152

PR #2306, opened 2026-09-16
lianad: answer the poller sync checks in the daemon_startup test

pythcoiner:
the poller's first recv_timeout(Duration::ZERO) can run before stop() pushes Shutdown, so it sends a getblockchaininfo the dummy bitcoind never answers because the test thread is already in t.join(); the request then burns 61 retries x 180s socket timeout, hanging the job ~3h before panicking. The mock now keeps answering sync checks (chain still syncing, so no real poll starts) until the daemon thread signals it stopped and wakes the last accept().

this had stuck CIs of #2303 & #2298 yesterday


edouardparis (approved):
utACK 55d619fc62dc9f587675ad2e8602ece96e0015fe

